PR #6, opened 2016-10-05
8.0 Add support for partner bank matching on invoice update

alexis-via:
Add support for partner bank matching on invoice update. Before, this feature was only supported on invoice creation. I developed this new feature this afternoon for my demo at my Odoo experience conference.

### DIFF
--- a/base_business_document_import/models/business_document_import.py
+++ b/base_business_document_import/models/business_document_import.py
@@ -238,6 +238,36 @@ class BusinessDocumentImport(models.AbstractModel):
         return partner
 
     @api.model
+    def _match_partner_bank(
+            self, partner, iban, bic, chatter_msg, create_if_not_found=False):
+        assert iban, 'iban is a required arg'
+        assert partner, 'partner is a required arg'
+        partner = partner.commercial_partner_id
+        iban = iban.replace(' ', '')
+        rpbo = self.env['res.partner.bank']
+        self._cr.execute(
+            """SELECT id FROM res_partner_bank
+            WHERE replace(acc_number, ' ', '')=%s
+            AND state='iban'
+            AND partner_id=%s
+            """, (iban, partner.id))
+        rpb_res = self._cr.fetchall()
+        if rpb_res:
+            return rpbo.browse(rpb_res[0][0])
+        elif create_if_not_found and bic:
+            partner_bank = rpbo.create({
+                'partner_id': partner.id,
+                'state': 'iban',
+                'acc_number': iban,
+                'bank_bic': bic,
+                })
+            chatter_msg.append(_(
+                "The bank account <b>IBAN %s</b> has been automatically "
+                "added on the supplier <b>%s</b>") % (
+                iban, partner.name))
+            return partner_bank
+
+    @api.model
     def _match_product(self, product_dict, chatter_msg, seller=False):
         """Example:
         product_dict = {
--- a/account_invoice_import/README.rst
+++ b/account_invoice_import/README.rst
@@ -45,7 +45,7 @@ To use this module, go to the menu *Accounting > Suppliers > Import Invoices* an
 
 .. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
    :alt: Try me on Runbot
-   :target: https://runbot.odoo-community.org/runbot/95/8.0
+   :target: https://runbot.odoo-community.org/runbot/226/8.0
 
 Known issues / Roadmap
 ======================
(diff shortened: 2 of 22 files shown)


base_business_document_import/models/business_document_import.py
```diff
@@ -238,6 +238,36 @@ def _match_shipping_partner(self, shipping_dict, partner, chatter_msg):
...
     @api.model
+    def _match_partner_bank(
+            self, partner, iban, bic, chatter_msg, create_if_not_found=False):
+        assert iban, 'iban is a required arg'
+        assert partner, 'partner is a required arg'
+        partner = partner.commercial_partner_id
+        iban = iban.replace(' ', '')
+        rpbo = self.env['res.partner.bank']
+        self._cr.execute(
+            """SELECT id FROM res_partner_bank
```
pedrobaeza, line 249: Why make this with SQL?

base_business_document_import/models/business_document_import.py
```diff
@@ -238,6 +238,36 @@ def _match_shipping_partner(self, shipping_dict, partner, chatter_msg):
...
+        if rpb_res:
+            return rpbo.browse(rpb_res[0][0])
+        elif create_if_not_found and bic:
+            partner_bank = rpbo.create({
+                'partner_id': partner.id,
+                'state': 'iban',
+                'acc_number': iban,
+                'bank_bic': bic,
+                })
+            chatter_msg.append(_(
```
pedrobaeza, line 264: You have to modify the translations due to this file change I think

base_business_document_import/models/business_document_import.py
```diff
@@ -238,6 +238,36 @@ def _match_shipping_partner(self, shipping_dict, partner, chatter_msg):
         return partner
 
     @api.model
+    def _match_partner_bank(
+            self, partner, iban, bic, chatter_msg, create_if_not_found=False):
```
pedrobaeza, line 242: To make it more tolerant, put `chatter_msg` as optional.

base_business_document_import/models/business_document_import.py
alexis-via: I need the replace() method in SQL because the IBAN can have spaces in res_partner_bank. In v9, this problem is solved by the acc_number_sanitized field, so I will remove this SQL request in v9 and use the ORM instead.

base_business_document_import/models/business_document_import.py
alexis-via: Translations are handled automatically via transifex, no ?

base_business_document_import/models/business_document_import.py
alexis-via: Mmm, it is that way in all the other methods of this module.

pedrobaeza: But you can start with this one being more tolerant :wink: for reusing it in models without chatter.

base_business_document_import/models/business_document_import.py
pedrobaeza: OK

base_business_document_import/models/business_document_import.py
pedrobaeza: Yes, but you will lose any of the already present translations. It's just to check and modify translation files to see the comment line about the code that contains this sentence.

alexis-via: There are no translation for this module so far.

base_business_document_import/models/business_document_import.py
alexis-via: When I'll change that, I'll change it everywhere ; I don't want to do it partially.

pedrobaeza: This is a good moment for that :wink:, but I know that you're not going to do it, hehe

base_business_document_import/models/business_document_import.py
pedrobaeza: OK

pedrobaeza (approved):
Squashing on merging

pedrobaeza:
@alexis-via, what are the plans for bringing this to v9?

alexis-via:
I'll port it to v10 in December 2016 or January 2017 (when my customers will go live on v10) ; I don't want to start the port too early because I still have things that I want to change on v8 and I would like to avoid too much backporting work. The 2 points I want to address are:
- the support of sale order import when your products have a price-include=True taxes
- add support for importing an invoice from a foreign country with VAT that you can't get back

I don't have any plans for v9, but my port to v10 should be compatible with v9 I guess.

pedrobaeza:
OK, thanks for sharing your plans. I'm not sure how to act because we need it on v9, but only for commodity (not mandatory requirement), but we want it now. I'll tell you later.

sebastienbeau (approved):

Issue #7, opened 2016-10-18
8.0 Travis red


pedrobaeza:
@alexis-via, last merge has discovered several problems on Travis. Can you please fix them?

alexis-via:
OK, but the pylint errors are not related to the latest merge, but to some new pylint rules that were not enabled before the merge.

pedrobaeza:
I have also repaired Transifex with the commit https://github.com/OCA/edi/commit/1ced127801eb46581830377ce7254ee1d9b91426. Closing for now, although it would be good to have the requirements.txt file as said by @moylop260.

PR #8, opened 2016-10-18
8.0 Use try/except when importing external libs

alexis-via:
Remove self.ensure_one() that has nothing to do in an api.model method

Fix bug #7

### DIFF
--- a/account_invoice_import_invoice2data/wizard/account_invoice_import.py
+++ b/account_invoice_import_invoice2data/wizard/account_invoice_import.py
@@ -15,7 +15,7 @@ try:
     from invoice2data.template import read_templates
     from invoice2data.main import logger as loggeri2data
 except ImportError:
-    logger.warning('Cannot import invoice2data')
+    logger.debug('Cannot import invoice2data')
 
 
 class AccountInvoiceImport(models.TransientModel):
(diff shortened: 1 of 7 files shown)


account_invoice_import_invoice2data/wizard/account_invoice_import.py
```diff
@@ -6,14 +6,17 @@
...
-
+import logging
 logger = logging.getLogger(__name__)
 
+try:
+    from invoice2data.main import extract_data
+    from invoice2data.template import read_templates
+    from invoice2data.main import logger as loggeri2data
+except ImportError:
+    logger.warning('Cannot import invoice2data')
```
moylop260: s/warning/debug/g
Like as our example from guidelines: https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#importerror

account_invoice_zugferd/models/account_invoice.py
```diff
@@ -8,15 +8,19 @@
...
-# from pprint import pprint
-
 logger = logging.getLogger(__name__)
+
+try:
+    from PyPDF2 import PdfFileWriter, PdfFileReader
+    from PyPDF2.generic import DictionaryObject, DecodedStreamObject,\
+        NameObject, createStringObject, ArrayObject
+except ImportError:
+    logger.warning('Cannot import PyPDF2')
```
moylop260: Same case

moylop260 (changes requested):
Thanks for your PR

Could you create a [requirements.txt](https://github.com/OCA/maintainer-quality-tools/blob/master/requirements.txt) in the root of the project (MQT automatically use it to install dependencies) and remove `pip install` from .travis.yml file?

pedrobaeza (approved):
Thanks for the followup. I fast-track to fix the branch.

PR #11, opened 2016-12-29
[9.0][MIG] account_invoice_import_invoice2data and dependencies: Migration to 9.0

hbrunn:


alexis-via (approved):

pedrobaeza:
Can you please check Travis?

hbrunn:
this will stay red until https://github.com/OCA/partner-contact/pull/364 is merged

hbrunn:
and then https://github.com/OCA/community-data-files/pull/20

pedrobaeza:
Ouch, I have launched rebuilds...

pedrobaeza:
I have merged all the dependencies, but Travis is still red. Please check.

alexis-via:
The field ean13 has been renamed ; code needs to be updated accordingly

hbrunn:
now it's only the lacking chart of accounts. Do we have a general solution for this or should every test create accounts as necessary?

tarteo (changes requested):
Hi @hbrunn,

I've some small requests. Other than that LGTM.

**In account_invoice_import:**

> To use this module, go to the menu Accounting > Suppliers > Import Invoices and upload a PDF or XML invoice of your supplier.


This menu item does not exists. I think it should be "Invoicing -> Purchases -> Import Invoice"

Can you move the models to the models directory (e.g. partner.py to models/partner.py), please?

**In account_invoice_import_invoice2data:**

> French users should also install the module l10n_fr_invoice_import available in the French localization, cf this PR.


The hyperlink for "This PR" can be removed; it refers to a PR for version 8.0.

**In base_business_document_import:**
additionnal -> additional

hbrunn:
I'll look into that next time I work on this. Will take a while I guess. Next time, please use github's review feature, this makes it a lot simpler to follow up on a review because then the comments are attached to the line of code the comment talks about

tarteo:
@hbrunn I agree, but I don't know how to review lines that are not changed. :(

alexis-via:
@tarteo I took into account your remarks on my PR for v10

thomaspaulb:
@tarteo Press the little button that looks like

image

and the file will expand.

tarteo:
@thomaspaulb I know that button. But when I expand the file contents I still cannot review the lines that aren't changed (the + button does not appear at the line).

hbrunn:
@tarteo done

tarteo (approved):
@hbrunn Thanks 👍

.travis.yml
```diff
@@ -24,6 +24,7 @@ addons:
 #     Search your packages here:
 #       https://github.com/travis-ci/apt-package-whitelist/blob/master/ubuntu-precise
       - wkhtmltopdf  # only add if needed and check the before_install section below
+      - poppler-utils # for invoice2data
```
hbrunn, line 27: @alexis-via I was doubting my sanity because for the life of me, I couldn't get the tests work. Then I got the suspicion that pdftotext parses the test invoice incorrectly, and that;s the case, see https://travis-ci.org/OCA/edi/jobs/230782037 - check the install phase, there I first print the output of standard pdftotext on travis, then the version you install. As you see, the total line is botched in the second version. I guess you needed the private version because back then travis had an ancient version only? Any objections to do it the way I propose? And I guess that's also the reason why this still runs on the sudo infrastructure, any objections to change this to the current OCA standard?

hbrunn:
@alexis-via @pedrobaeza it's green finally :-)

hbrunn:
@alexis-via can we merge this?

hbrunn:
@alexis-via can we merge in order to avoid duplicated efforts as in #23 ?

PR #17, opened 2017-03-22
[8.0] base_business_document_import: Add match for accounts and journals

alexis-via:
base_business_document_import: Add match for accounts, analytic accounts and journals

Add support for insertion of title in error messages


sebastienbeau (approved):
Module have been tested

PR #18, opened 2017-05-03
[ADD] 9.0 saleorder_import_invoice2data and [MIG] sale_commercial_partner / sale_order_import

thomaspaulb:
@hbrunn Tests are running but code needs review


sale_order_import/__openerp__.py
```diff
@@ -4,15 +4,16 @@
 
 {
     'name': 'Sale Order Import',
-    'version': '8.0.1.0.0',
+    'version': '9.0.1.0.0',
     'category': 'Sales Management',
     'license': 'AGPL-3',
     'summary': 'Import RFQ or sale orders from files',
-    'author': 'Akretion,Odoo Community Association (OCA)',
+    'author': 'Akretion,Odoo Community Association (OCA), Sunflower IT',
```
thomaspaulb: If done properly I think we should be second in the list

sale_order_import/wizard/sale_order_import.py
```diff
@@ -181,14 +182,20 @@ def _prepare_order(self, parsed_order, price_source):
...
                         existing_orders[0].state))
-        partner_change_res = soo.onchange_partner_id(partner.id)
-        assert 'value' in partner_change_res, 'Error in partner change'
+        values = {'partner_id': partner.id}
+        fake_so = soo.new(values)
+        fake_so.onchange_partner_id()
+        partner_change_res = dict(
+            (key, value.id if not isinstance(value, int) else value)
+            for key, value in dict(fake_so._cache).iteritems()
+        )
```
thomaspaulb: There must be a prettier way to do this, the issue was that `dict(fake_so._cache)` returns the objects in a dict, whereas we need the id's.

sale_order_import/wizard/sale_order_import_view.xml
```diff
@@ -5,17 +5,16 @@
...
 <openerp>
-<data>
 
 <record id="sale_order_import_form" model="ir.ui.view">
     <field name="name">sale.order.import.form</field>
     <field name="model">sale.order.import</field>
     <field name="arch"  type="xml">
         <form string="Import Sale Orders">
-            <group name="technical" invisible="1">
+            <group string="technical" invisible="1">
```
thomaspaulb: @Kiplangatdan Why was it necessary to change all the `name` occurrences to `string` or `group` ? IMO `name` is the preferred way to make the object accessible to inheriting views, and it's commonly used in Odoo 9.

If the aim was to add a caption, you can leave the `name` and add `string` separately.

sale_order_import_invoice2data/README.rst
```diff
@@ -0,0 +1,100 @@
+===================================
+Sale Order Import Saleorder2data
```
thomaspaulb: Invoice2data

sale_order_import_invoice2data/README.rst
```diff
@@ -0,0 +1,100 @@
...
+------------
+
+* Tom Blauwendraat <info@sunflowerweb.nl>
+* Terence Nzaywa <terrence@sunflowerweb.nl>
+* Alexis de Lattre <alexis.delattre@akretion.com>
+
+Maintainer
+----------
+
+This module is maintained by Sunflower IT and Therp BV.
```
thomaspaulb: Maybe this should be `the OCA` instead.

account_invoice_import/wizard/account_invoice_import.py
```diff
@@ -518,13 +536,16 @@ def update_invoice_lines(self, parsed_inv, invoice, seller):
...
-            currency_id=invoice.currency_id.id,
-            company_id=invoice.company_id.id)['value']
+        new_line = self.env['account.invoice.line'].new({
+            'invoice_id': invoice,
+            'qty': import_line['qty'],
+            'product_id': product,
+        })
+        new_line._onchange_product_id()
+        vals = {
+            f: new_line._fields[f].convert_to_write(new_line)
```
thomaspaulb: I think we should use this as an example to clean up [this](https://github.com/OCA/edi/pull/18/files#diff-b5d59e23d0ded258b90d7f2ca8fba4b2R188)

thomaspaulb (commented):
@Kiplangatdan requesting changes

sale_order_import/wizard/sale_order_import.py
thomaspaulb: Actually have found it in @hbrunn's commits, see comment below

thomaspaulb:
We should also fix [these style issues](https://travis-ci.org/OCA/edi/jobs/228296292#L981)

sale_commercial_partner/report/sale_report.py
```diff
@@ -11,12 +11,19 @@ class SaleReport(models.Model):
     commercial_partner_id = fields.Many2one(
         'res.partner', string='Commercial Customer', readonly=True)
 
-    def _select(self):
-        sql_str = super(SaleReport, self)._select()
-        sql_str += ', s.commercial_partner_id as commercial_partner_id'
-        return sql_str
+    # def _select(self):
```
thomaspaulb: Have seen that Odoo 9.0 already puts `partner.commercial_id` in this view, so this part can indeed be removed.

hbrunn:
@thomaspaulb I just changed my proposal for the invoice line parsing, this now is a lot more robust and gives better results. You will have to change your templates though, but the new format is a lot simpler to write

thomaspaulb:
Alright, we will change the template for the test accordingly. Do you have
an example or is it clear from the code?

Op 8 mei 2017 18:52 schreef "Holger Brunn" <notifications@github.com>:

hbrunn:
https://github.com/m3nu/invoice2data/pull/22/files#diff-f4398b8b9e01f2cb73701655ba612b47R11

hbrunn:
I changed #11 quite a bit, you should rebase your work on this branch. And hopefully we get this merged soon, then reviewing this is a lot simpler because if won't contain #11's cruft diffs any more. Github really needs the prerequisite branch feature launchpad had...

sale_order_import_invoice2data/wizard/sale_order_import_pdf.py
```diff
@@ -0,0 +1,178 @@
...
+            templates += read_templates(local_templates_dir)
+        exclude_built_in_templates = tools.config.get(
+            'invoice2data_exclude_built_in_templates', False)
+        if not exclude_built_in_templates:
+            templates += read_templates(
+                pkg_resources.resource_filename('invoice2data', 'templates'))
+        logger.debug(
+            'Calling invoice2data.extract_data with templates=%s',
+            templates)
+        testspath = os.path.dirname(os.path.realpath(__file__))
```
hbrunn: there's an API for this: https://github.com/OCA/OCB/blob/9.0/openerp/modules/module.py#L142
But in the end, I think we need a base module that provides a model
```python
class Invoice2dataTemplate(models.Model):
    type = fields.Selection([]) # other modules like yours add a type of template, as in invoice_import, sale_order_import, etc
    template = fields.Text('The template')
```
This makes it much more convenient to deliver templates in your module.

account_invoice_import/wizard/account_invoice_import.py
hbrunn: this is slightly more complicated, here the fixed version: https://github.com/OCA/edi/pull/11/files#diff-57b8c6799922b0e67a34d15dc8b010a8R142

sale_order_import_invoice2data/wizard/sale_order_import_pdf.py
```diff
@@ -0,0 +1,178 @@
...
+                    },
+                    # sale_order_import expects this way,
+                    'desc': line.get('desc'),
+                    'code': line.get('code'),
+                    'ean13': line.get('ean13'),
+                    # --------------
+                    'taxes': line.get('taxes'),
+                    'unit_price': line.get('unit_price'),
+                    'price': line.get('price'),
+                }
```
hbrunn, line 142: I think it would be nice to add the intersection of `line.keys()` and `self.env['sale.order.line']._fields.keys()` to the line values too. This way, you simply have to use odoo names in your templates, and you get whatever exotic field you might want to use without having to write custom code

sale_order_import_invoice2data/wizard/sale_order_import_pdf.py
```diff
@@ -0,0 +1,178 @@
...
+    @api.model
+    def create_order(self, parsed_order, price_source):
+        bdio = self.env['business.document.import']
+        so_vals = self._prepare_order(parsed_order, price_source)
+        order = self.env['sale.order'].create(so_vals)
+        bdio.post_create_or_update(parsed_order, order)
+        logger.info('Sale Order ID %d created', order.id)
+        return order
+
+class BusinessDocumentImport(models.AbstractModel):
```
hbrunn: one class per file

sale_order_import_invoice2data/wizard/sale_order_import_pdf.py
```diff
@@ -0,0 +1,178 @@
...
+                        "Error on PDF order line %d: the quantity should "
+                        "be strictly positive") % i)
+                vals['qty'] = qty
+                parsed_order['lines'].append(vals)
+        if 'amount_tax' in invoice2data_res:
+            parsed_order['amount_tax'] = invoice2data_res['amount_tax']
+        return parsed_order
+
+    @api.model
+    def create_order(self, parsed_order, price_source):
```
hbrunn: what's this override good for? Looks suspiciously like you just copied https://github.com/OCA/edi/blob/9.0/sale_order_import/wizard/sale_order_import.py#L212

sale_order_import_invoice2data/wizard/sale_order_import_pdf.py
```diff
@@ -0,0 +1,178 @@
...
+
+    # @api.model
+    # def fallback_parse_pdf_saleorder(self, file_data):
+    #     '''This method must be inherited by additionnal modules with
+    #     the same kind of logic as the account_bank_statement_import_*
+    #     modules'''
+    #     return self.invoice2data_parse_saleorder(file_data)
+    #
+    @api.model
+    def parse_pdf_order(self, file_data, detect_doc_type=False):
```
hbrunn: you destroy modularity here. Do something similar to what [account_invoice_import](https://github.com/OCA/edi/blob/9.0/account_invoice_import/wizard/account_invoice_import.py#L52) [does](https://github.com/OCA/edi/blob/9.0/account_invoice_import/wizard/account_invoice_import.py#L65) in [sale_order_import](https://github.com/OCA/edi/blob/9.0/sale_order_import/wizard/sale_order_import.py#L108)

sale_order_import/wizard/sale_order_import.py
```diff
@@ -152,6 +152,7 @@ def parse_pdf_order(self, order_file, detect_doc_type=False):
     #    }]
 
     @api.model
+    @api.multi
```
hbrunn: it's either `model` or `multi`, not both

hbrunn (changes requested):

account_invoice_import/wizard/account_invoice_import.py
thomaspaulb: Have made it into a helper function:

```
def get_values_for_create(instance):
    """ Helper function to get values from object as dict """
    return {
        f: instance._fields[f].convert_to_write(v)
        if not isinstance(instance._fields[f],
            fields._RelationalMulti)
        else v.ids
        for f, v in instance._cache.iteritems()
    }
```

sale_order_import_invoice2data/wizard/sale_order_import_pdf.py
hbrunn: in my own override of this, I use
```python
common_fields = set(self.env['sale.order.line']._fields.keys()) &\
    set(parsed_line.keys()) - set(order_line.keys())
for field in common_fields:
    order_line[field] = parsed_line[field]
```
this way, you can support arbitrary fields without having to write custom code

sale_order_import_invoice2data/wizard/sale_order_import_pdf.py
thomaspaulb: Is this something you have done in the invoice version of this module already?

sale_order_import_invoice2data/wizard/sale_order_import_pdf.py
thomaspaulb: I am calling super now in case I am not happy with the PDF, I hope that solves the issue

sale_order_import_invoice2data/wizard/sale_order_import_pdf.py
thomaspaulb: Suspicions confirmed.

sale_order_import_invoice2data/wizard/sale_order_import_pdf.py
hbrunn: no, not yet. For now, just remove this part of the code and set your template path in the test. Then it's very simple to change this later to use the yet-to-be-written module, and the test template will not interfere with normal operations

hbrunn:
closing this in favor of #21

PR #19, opened 2017-05-04
[8.0] Improve + modularize generation of UBL and CII XML

alexis-via:
This PR is required for my implementation of Chorus: https://github.com/OCA/l10n-france/pull/103

### DIFF
--- a/base_ubl/models/ubl.py
+++ b/base_ubl/models/ubl.py
@@ -70,8 +70,18 @@ class BaseUbl(models.AbstractModel):
             logger.warning('UBL: missing country on partner %s', partner.name)
 
     @api.model
-    def _ubl_add_contact(self, partner, parent_node, ns, version='2.1'):
-        contact = etree.SubElement(parent_node, ns['cac'] + 'Contact')
+    def _ubl_get_contact_id(self, partner):
+        return False
+
+    @api.model
+    def _ubl_add_contact(
+            self, partner, parent_node, ns, node_name='Contact',
+            version='2.1'):
+        contact = etree.SubElement(parent_node, ns['cac'] + node_name)
+        contact_id_text = self._ubl_get_contact_id(partner)
+        if contact_id_text:
+            contact_id = etree.SubElement(contact, ns['cbc'] + 'ID')
+            contact_id.text = contact_id_text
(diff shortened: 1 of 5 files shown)


hbrunn (changes requested):
looks fine technically, but I think you should add tests too

alexis-via:
I added tests in the upper-level module l10n_fr_chorus_ubl which has inherits of the new method I implemented in this PR, which allows me to test that the "full" XML is properly generated.

hbrunn:
and there's no chance to test something in the general modules? I'm very unfamiliar with UBL, that's why I can't judge this

florentboucherieconsulting (approved):
Work fine : it imported more than 300 pdf invoice for me !

base_ubl/models/ubl.py
```diff
@@ -124,21 +186,11 @@ def _ubl_add_party(
...
-            company_id.text = commercial_partner.sanitized_vat
-            tax_scheme = etree.SubElement(
-                party_tax_scheme, ns['cac'] + 'TaxScheme')
-            tax_scheme_id = etree.SubElement(
-                tax_scheme, ns['cbc'] + 'ID', schemeID='UN/ECE 5153',
-                schemeAgencyID='6')
-            tax_scheme_id.text = 'VAT'
+        self._ubl_add_party_tax_scheme(
+            commercial_partner, party, ns, version=version)
+        if company:
```
thomaspaulb, line 191: @alexis-via Do you remember if there was any particular reason why the PartyLegalEntity is only added for the partner who has the `res.company` in Odoo, eg. the sender of a sales invoice? In Netherlands the PartyLegalEntity is also required for the customer side.

alexis-via: I don't think there is a good reason. I think it was linked to the fact that i decided to use the commercial_partner of the invoice as invoice address, which is probably a bad reason, cf https://github.com/OCA/edi/issues/104 As I said in this issue, I'm OK to modify that.

thomaspaulb: @alexis-via Thanks for pointing me to the issue! I saw the fix for it was even already merged to the main 10.0 branch, so I backported it to the version I needed it for.

PR #23, opened 2017-07-17
[9.0] Migration

JayVora-SerpentCS:


hbrunn:
this migration is partially done in https://github.com/OCA/edi/pull/23 already, please check that first

PR #25, opened 2017-08-10
Added support for importing lines with invoice2data

D4N:
I have added a very rudimentary line import which processes the data from invoice2data into a format that odoo can understand. I have tried to mimic the behaviour of the UBL importer, but as invoice2pdf will usually not extract any taxes, I have left them out.

Also, I guess some sort of user documentation should be added, so that the user knows how to name their line fields in invoice2data, otherwise this thing will not work at all. Where would be the best place to do that?


bosd:
@D4N Did you make a new PR for this?

PR #38, opened 2018-01-04
[ADD] new addon to automatically import invoice from files in server directories

Cedric-Pigeon:
This PR add an automatic process to upload file from directories and feed the wizard in account_invoice_import addon.

### DIFF
--- a/account_invoice_import_directory/README.rst
+++ b/account_invoice_import_directory/README.rst
@@ -1,0 +1,87 @@
+.. image:: https://img.shields.io/badge/license-AGPL--3-blue.png
+   :target: https://www.gnu.org/licenses/agpl
+   :alt: License: AGPL-3
+
+================================
+Account Invoice Import Directory
+================================
+
+This module extends the functionality of account_invoice_import and allows to
+run the import of invoices through an automatic process scanning some
+directories on the Odoo server.
+
+Installation
+============
+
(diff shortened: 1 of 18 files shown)


Cedric-Pigeon:
@alexis-via What is your opinion on this PR ?

alexis-via:
Seems funny :) I'm afraid you caught the same virus as me on electronic invoicing and accounting automation... 

I guess that this module tries to be a solution to avoid manually dropping the PDF in Odoo ; copying to a directory of Odoo (shared by Samba to the accounting team of the company) is certainly easier. Is that your goal when developping this module ? Another approach would be to have a daemon that listen to new files in that directory (via inotify) and, when it finds a new PDF or XML file, immediately sends it to Odoo's account invoice import wizard. That would avoid the need to wait for the scheduler to run ; you would get the result in Odoo immediately.

Cedric-Pigeon:
@alexis-via The goal of this is to receive ubl files from another platform which do the OCR job (https://www.arco.be/en/).

alexis-via:
FYI, I proposed another approach than this module: a script that use inotify to auto-push invoices to Odoo. Advantage: no need to wait for the cron to run, the invoices is sent to Odoo immediately after save. https://github.com/OCA/edi/pull/84

PR #43, opened 2018-02-01
[11.0][MIG] base_ubl: port to V11

astirpe:
Depends on

- [x] base_vat_sanitized - OCA/partner-contact#552
- [x] product_uom_unece - OCA/community-data-files#35
- [x] account_tax_unece - OCA/community-data-files#34

PR #47, opened 2018-02-12
[11.0][MIG] sale_order_ubl

astirpe:
Depends on:

- [x] base_ubl: #43


astirpe:
Tests fail because of those [two lines](https://github.com/odoo/odoo/blob/11.0/odoo/addons/base/ir/ir_actions_report.py#L562-L563) that force Odoo to generate a HTML report instead of a PDF one.

PatrickMallee (approved):
Works fine

pedrobaeza:
A bit of squashing? :wink:

astirpe:
Squashed

PR #53, opened 2018-04-11
[FIX] can not change currency of existing company anymore

njeudy:
Fix failing test, because we can not change currency with company having existing data

### DIFF
--- a/base_business_document_import/tests/test_business_document_import.py
+++ b/base_business_document_import/tests/test_business_document_import.py
@@ -102,7 +102,9 @@ class TestBaseBusinessDocumentImport(TransactionCase):
         currency_dict = {'iso_or_symbol': '€'}
         res = bdio._match_currency(currency_dict, [])
         self.assertEqual(res, self.env.ref('base.EUR'))
-        self.env.user.company_id.currency_id = self.env.ref('base.KRW')
+        currency_id = self.env.ref('base.KRW').id
+        self.cr.execute("UPDATE res_company SET currency_id = %s WHERE id = 1",
+                        (currency_id,))
         currency_dict = {}
         res = bdio._match_currency(currency_dict, [])
         self.assertEqual(res, self.env.ref('base.KRW'))


njeudy:
@astirpe here is failing test.

astirpe (approved):

pedrobaeza (changes requested):
Instead of removing that part, please change currency by SQL

njeudy:
@pedrobaeza why to try to change and test something that you could not do in Odoo ?

This part of code check that if you change company currency, the function still work .. but now on V11 you could not change company currency anymore .. 

I know I can use SQL but having only usefull tests is better ?

pedrobaeza:
This is because we are testing a use case where the currency of the company is not the base one, and the same company is being used for not creating a company and all the related stuff (some kind of test shortcut). That's why I ask to keep it that way.

njeudy:
@pedrobaeza travis hangs on : base_business_document_import/tests/test_business_document_import.py:107: [E8103(sql-injection), TestBaseBusinessDocumentImport.test_match_currency] SQL injection risk. Use parameters if you can. - More info https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#no-sql-injection

And I try with sql but I think I miss something .. can you help me ?

njeudy:
@astirpe did you have some idea for this failing  test ?

astirpe:
@njeudy can you try this?

```
cr.execute("UPDATE res_company SET currency_id = %s WHERE id = 1", (currency_id,))
```

njeudy:
@astirpe nope .. that not correct .. think we are not in same env / cursor / transaction ... @pedrobaeza did you have an example for sql modification in tests ?

astirpe:
@njeudy sorry, that line was meant for the sql-injection warning (that is now fixed).

About the failing test, isn't cursor `self.cr` working fine?
https://github.com/odoo/odoo/blob/11.0/addons/account/tests/test_reconciliation.py#L29

njeudy:
@pedrobaeza @astirpe done :) and green / thanks for the help

pedrobaeza (approved):

pedrobaeza:
Fast-tracking this for having green branch.

PR #62, opened 2018-05-18
[11.0][FIX] base_business_document_import: selecting partner in a multicompany

### LINKED ISSUE #54: base_business_document_import: selecting partner in a multicompany environment
Tested with 11.0.

Here is the scenario:

- In a multicompany environment, user A is allowed to access multiple companies, eg. Company1 and Company2.
- For Company1, a vendor (partner) with vat number = XYX is present.
- In Company2 there is also a vendor with same vat number = XYX; so that vendor is known in both Company1 and Company2.
- The module `base_business_document_import` is capable to select the partner based on the result of method `def _match_partner()`. In our scenario, the match is based on the vat number: see [lines](https://github.com/OCA/edi/blob/11.0/base_business_document_import/models/business_document_import.py#L108-L111).

User A logs in with Company1 and tries to import a document (eg. an invoice).

I would expect that the retrieved vendor (partner) is the one with vat = XYX belonging to Company1.
But it could happen that the `def _match_partner()` retrieves the wrong partner: the one of Company2.

Adding `('company_id', '=', self.env.user.company_id.id)` in the domain of [these lines](https://github.com/OCA/edi/blob/11.0/base_business_document_import/models/business_document_import.py#L108-L111) seems fixing the problem.
However there could be other cases where a patch is needed in order to make the module multicompany compliant.
I don't have time to investigate further and I will not have time to propose a PR during the next weeks, so I open this issue as a reminder to open a PR. In the meanwhile any help will be very appreciated!

astirpe:
Fixes #54

### DIFF
--- a/base_business_document_import/models/business_document_import.py
+++ b/base_business_document_import/models/business_document_import.py
@@ -55,9 +55,12 @@ class BusinessDocumentImport(models.AbstractModel):
             'ref': 'C1242',
             'phone': '01.41.98.12.42',
             }
-        The keys 'phone' are used by the module
+        The key 'phone' is used by the module
         base_phone_business_document_import
         """
+        company = self.env.user.company_id
+        if hasattr(self, 'company_id') and self.company_id:
+            company = self.company_id
         rpo = self.env['res.partner']
         self._strip_cleanup_dict(partner_dict)
         if partner_dict.get('recordset'):
@@ -108,7 +111,10 @@ class BusinessDocumentImport(models.AbstractModel):
             partners = rpo.search(
                 domain + [
                     ('parent_id', '=', False),
-                    ('sanitized_vat', '=', vat)])
+                    ('sanitized_vat', '=', vat),
+                    '|',
+                    ('company_id', '=', False),
+                    ('company_id', '=', company.id)])
             if partners:
                 return partners[0]
             else:
--- a/base_business_document_import/__manifest__.py
+++ b/base_business_document_import/__manifest__.py
@@ -3,7 +3,7 @@
 
 {
     'name': 'Base Business Document Import',
-    'version': '11.0.1.0.0',
+    'version': '11.0.1.0.1',
     'category': 'Tools',
     'license': 'AGPL-3',
     'summary': 'Provides technical tools to import sale orders or supplier '


base_business_document_import/models/business_document_import.py
```diff
@@ -107,6 +107,7 @@ def _match_partner(
             # use base_vat_sanitized
             partners = rpo.search(
                 domain + [
+                    ('company_id', '=', self.env.user.company_id.id),
```
pedrobaeza: The problem is if this comes from sudo or in the ideal company-per-document approach, so don't we have a company pointer for the current document?

astirpe: Model `business.document.import` is abstract, so I don't see where can I get the company. Any suggestions?

pedrobaeza: We can test with `hasattr` if the model where is implemented has it, and if not, fallback to the one you propose.

astirpe: @pedrobaeza can you check if the latest commit is what you meant?

pedrobaeza: Yes!

pedrobaeza (approved):

base_business_document_import/models/business_document_import.py
```diff
@@ -107,6 +110,7 @@ def _match_partner(
             # use base_vat_sanitized
             partners = rpo.search(
                 domain + [
+                    ('company_id', '=', company.id),
```
pedrobaeza: Hey, wait! You have to also include with OR if the company_id is False.

base_business_document_import/models/business_document_import.py
```diff
@@ -110,7 +110,7 @@ def _match_partner(
             # use base_vat_sanitized
             partners = rpo.search(
                 domain + [
-                    ('company_id', '=', company.id),
+                    ('company_id', 'in', [company.id, False]),
```
pedrobaeza: Are you sure this works? I have always made (and seen) this as:
```python
'|',
('company_id', '=', False),
('company_id', '=', company.id),
```

etobella (approved):
LGTM :+1:

astirpe:
Is it ok to merge this one?

PR #41, opened 2018-01-28
[11.0][MIG] : mig account_invoice_import

njeudy:
PR needed:

- [X] base_vat_sanitized https://github.com/OCA/partner-contact/pull/552
- [X] account_tax_unece https://github.com/OCA/community-data-files/pull/34
- [X] product_uom_unece https://github.com/OCA/community-data-files/pull/35
- [X] base_business_document_import https://github.com/OCA/edi/pull/42
- [X] onchange_helper https://github.com/OCA/server-tools.git

PR #63, opened 2018-05-31
[ADD] account_invoice_import_facturae (11.0)

etobella:
Account Invoice Import Facturae
===============================

This module allows to import facturae invoices into odoo.

[Facturae](http://www.facturae.gob.es/) is the standard invoice syntaxis in Spain and
it is required to use this format for B2G invoices. Although the creation and
sending is managed in [oca/l10n-spain](https://github.com/oca/l10n-spain), this module allows to import supplier
invoices into your system.


pedrobaeza:
This maybe should be in OCA/l10n-spain, not here. What do you think, @alexis-via ?

etobella:
I had the same though, but I putted the module here because ZUGFeRD invoices are from germany and all the modules are managed here (in v10.0).

pedrobaeza:
Well, ZUGFeRD is a more open standard, that is also used for France. That's why is hosted here.

account_invoice_import_facturae/wizard/account_invoice_import.py
```diff
@@ -0,0 +1,144 @@
...
+            etree.parse(tools.file_open(
+                "Facturaev3_2.xsd",
+                subdir="addons/account_invoice_import_facturae/data"
+            )))
+        facturae_schema.assertValid(xml_root)
+        sign = xml_root.find(
+            'ds:Signature', namespaces={'ds': xmlsig.constants.DSigNs}
+        )
+        if sign is None:
+            raise ValidationError(_('Signature node cannot be found'))
```
AaronHForgeFlow: Would be possible to import unsigned invoices?

etobella: It could be possible, but the signature is important, because it certifies that the invoice is not modified.
We could do if sign is not None: xmlsig.SignatureContext().verify(sign)
It is ok?

AaronHForgeFlow: Yes, that is ok. I tried and it worked for both, signed and unsigned

AaronHForgeFlow:
It is working fine when I select the following invoice.import.config:

image

However it's failing if I select a different configuration because when no account is set in the invoice.import.config then the account_id field is None and that triggers an error. Also the journal id is null. Maybe that should be fixed here: https://github.com/OCA/edi/pull/41 Anyway a default account should be set in any case, shouldn't it?

etobella:
The problem could be related to the invoice you are using. Today, I made a PR oca/l10n-spain#865 that allows to export the invoices in facturae format with the ArticleCode informed.

AaronHForgeFlow:
@etobella exporting the file is fine. The output file seems well formatted. However still having issue with the  `account_invoice_import`  module not with this one. This one seems ok.

AaronHForgeFlow (approved):
:+1: Imported complex invoices successfully! Thanks.

moylop260 (approved):

pedrobaeza:
I still insists in my comment before merging https://github.com/OCA/edi/pull/63#issuecomment-393451655

etobella:
If you think that it should go to l10n-spain I can move it. I was waiting for the comment of @alexis-via in order to decide.

alexis-via:
Thanks for this great work! I'm very glad that other people are building modules above account_invoice_import and other OCA/edi modules.
About the exact location of this module, my opinion is similar to Pedro's point of view: as this facturae format seems to be only for spain and is not based on international standards (such as UBL and CII), OCA/l10n-spain seems the best choice.

etobella:
Great, Module has been moved. You can find the PR on OCA/l10n-spain#941

Issue #29, opened 2017-10-03
Migration to version 11.0


pedrobaeza:
# Todo

https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-11.0

# Modules to migrate

- [x] account_e-invoice_generate - By @astirpe - #44 
- [ ] account_invoice_download - By @rousseldenis - #91 
- [ ] account_invoice_download_weboob - By @rousseldenis - #92 
- [ ] account_invoice_factur-x
- [ ] account_invoice_factur-x_py3o
- [x] account_invoice_import - By @njeudy - #41 
- [ ] account_invoice_import_factur-x
- [ ] account_invoice_import_invoice2data - By @tarteo - #76 
- [x] account_invoice_import_ubl - By @astirpe - #49 
- [x] account_invoice_ubl - By @astirpe - #46 
- [ ] account_invoice_ubl_py3o
- [x] base_business_document_import - By @njeudy - #42 
- [ ] base_business_document_import_phone
- [ ] base_business_document_import_stock
- [x] base_ubl - By @astirpe - #43 
- [x] base_ubl_payment - By @astirpe - #45 
- [ ] base_zugferd - By @njeudy - #70
- [ ] purchase_order_import
- [ ] purchase_order_import_ubl
- [x] purchase_order_ubl - By @astirpe - #48 
- [ ] purchase_order_ubl_py3o
- [x] sale_order_import - By @tarteo - #60 
- [ ] sale_order_import_csv
- [x] sale_order_import_ubl - By @tarteo - #61 
- [x] sale_order_ubl - By @astirpe - #47 
- [ ] sale_order_ubl_py3o
- [ ] sale_stock_order_import

astirpe:
base_ubl https://github.com/OCA/edi/pull/43

astirpe:
base_ubl_payment #45

astirpe:
account_invoice_ubl https://github.com/OCA/edi/pull/46

astirpe:
@pedrobaeza you were faster than me! I was about to write:
sale_order_ubl #47
😄

pedrobaeza:
hehe, I automate in my flow this when a mail incomes.

astirpe:
purchase_order_ubl #48

astirpe:
account_invoice_import_ubl #49

emagdalenaC2i:
@pedrobaeza 
base_business_document_import PR is nº 42 and it is completed

emagdalenaC2i:
[...]

PR #65, opened 2018-08-08
[ADD] adds modules from v10, migrates factur-x modules to v11

wpichler:


pedrobaeza:
Please only one module per PR, and there are some of them already migrated. Closing meanwhile.

wpichler:
One thought about the "one module per PR" thing - i don't think that this is really a good idea - because many modules do depend on other modules - so you are not able to easy test such a PR - also runbot will fail on it. Wouldn't it be better to say "one PR for one module with its depends ?"

pedrobaeza:
No, it's one module per PR. You start from the one without any inter-dependencies, and then you can create consecutive PRs with the following modules. It's very easy to fulfill dependencies between PRs adding a temporary commit in the dependent PRs with the branch from the other PR put in oca_repositories.txt, and thus having green runbot/Travis meanwhile the base one is merged. Other contributors just wait to push the other ones until the base one is merged.

Why doing this way? Simply for easing and speeding up the reviewing process, although it seems to be the contrary at the beginning. It's far more difficult to review a PR with a bunch of modules than only one. And also a blocking comment on the final dependent module will block the entire PR (while base modules can be correct) doing on a batch mode.

wpichler:
Ok - i got the point - thanks for explaining this...

pedrobaeza:
Let me know if you need assistance in the process. Check also https://github.com/OCA/edi/issues/29 for tracking the progress of the migration of the modules of the repository.

wpichler:
Ok - maybe we will need some assistence - we do currently port all edi ubl modules to version 11. And we will add some more modules - to be able to handle the transfer of generated pdf / xml / whatever files with a transfer base module - and then some provider depended specific transfer modules (holodeck AS4 for example)

PR #66, opened 2018-08-14
[10.0] Fix Factur-X XML generation to comply with official CII schematron

alexis-via:
Add improve sale_order_import: fix import of sale order with visible discounts when price_source = 'pricelist'.
Add unit tests in sale_order_import


alexis-via:
This PR has the code that was used to generate the v8 version of the sample Factur-X invoices available on http://fnfe-mpe.org/factur-x/

MOSAMAD (approved):
Tested the import of a real sale order with discount and I confirm the bug is fixed. No code review. All right !

PR #68, opened 2018-09-12
[10.0][IMP] - check that customer vat match with user current company

sbejaoui:
This check prevent users with multi-company access to import invoices other than their current company


sbidoul:
@sbejaoui's contributions are now covered by the ACSONE ECLA

sbidoul:
@sbejaoui it looks like your change makes travis red. Can you check?

Cedric-Pigeon (approved):

alexis-via:
The check is done in account_invoice_import on that line of code : https://github.com/OCA/edi/blob/10.0/account_invoice_import/wizard/account_invoice_import.py#L465
And the 'company' key in the pivot dict is implemented in account_invoice_import_ubl.
So it already checks that the invoice is imported in the right company.

PR #70, opened 2018-10-02
 [11.0][MIG] base_zugferd

njeudy:
Migrate base_zugferd to V11


njeudy:
@astirpe @pedrobaeza can you check this ? will squash into one commit if it's ok for you

pedrobaeza:
Why there are files of `base_facture-x` here?

pedrobaeza:
OK, I see the renaming, but shouldn't it be `base_factur-x` (without `e`)?

njeudy:
@pedrobaeza yes your right. will change this ..

yucer:
In the PR for [12.0 port](https://github.com/OCA/edi/pull/82) there are some commits that are not being included here.

Do you think they are needed also in this port ? They are:

- [Factur-X invoice generation: level is now configurable](https://github.com/OCA/edi/pull/82/commits/070cde4f55e07a633f4294a6763cd5a505d281f2)

- [FIX official Factur-X XSD....](https://github.com/OCA/edi/pull/82/commits/ea278740dcecbd1bbda6d0960953c413a998c81b)

yucer:
I did try to re-base this PR over the current *11.0* here #148 

To check if that is the reason for the lowered coverage.

yucer:
> I did try to re-base this PR over the current _11.0_ here #148
> 
> To check if that is the reason for the lowered coverage.

It seems that it was just the amount of code added.

This is a base module ... @pedrobaeza  what you use to do in those cases ? 

Submit the code with the migration of other addons that use this base code ?  

- account_invoice_factur-x
- account_invoice_import_factur-x

Or just let the PR open as base for the migration other PRs ?

pedrobaeza:
No worry about the coverage, although it's always convenient that module are self-tested, but in this case can be barely impossible. This PR can be merged and then the rest of the dependent PRs can be rebased for having proper CIs.

yucer:
> No worry about the coverage, although it's always convenient that module are self-tested, but in this case can be barely impossible. This PR can be merged and then the rest of the dependent PRs can be rebased for having proper CIs.

Sorry, the following might be obvious.  Just to be sure that I understood well.

I see that there is a tendency to migrate just one module in a PR. I guess that it is some kind of community convention or good practice. 

Do you mean that in these case is good to include the module in the PRs of dependent modules ? Or it is always that way?

And what about if the module being migrated depends on a non-yet migrated module of another repo ?

 Is there anyway to select the branches to be tested together ? (I mean test two PRs with the corresponding migrations at once?)

yucer:
I am just asking because I see that this is happening very frequent.

Fore example in #79 the port of `account_invoice_factur-x` was failing because of a missing port of `account_payment_partner` from *OCA\bank-payment*.

Now that problem is solved (I guess because it was integrated), but is failing because another dependency (this time from the same repo, is missing to be migrated)

I guess if we can test many PR together, there are less interruptions one one could be working with the temporary result of the migration of the other.

But, Ok... it is not a big think. It is just to migrate the dependencies first. 

I am just learning from these cases.

pedrobaeza:
It's not a tendency, but a guideline. Reviewing a pull request with several involved modules is very hard and it's not useful. You must start from the base dependency, get merged, and then continue with the rest. If you want to publish all of them for at least having a code review, that can be done. If it's from other repo, you can temporarily adjust oca_dependencies.txt to your PR branch for getting green CIs + runbot, and if it's the same repo, you can include in the dependent PR the commits from the dependent module, but both things must be undone when the base PR is merged, and in no way the dependent PR must be merged before the base one.

yucer:
Trying to fix the module rename in #149

yucer:
@njeudy Do you want to continue this ?

I was just trying to make your port work. Because the new rename it is supposed to be taken as basis for 12.0 

If you want I can continue with the port.

PR #75, opened 2018-10-30
[10.0] Update for Factur-X specs 1.3

alexis-via:


alexis-via:
Factur-X V1.3 has been released, so I'll merge it soon.

PR #84, opened 2019-02-18
[10.0] account_invoice_import: Add script to auto-push invoices to Odoo via inotify

alexis-via:
Add method invoice_import_direct() in account.invoice.import, used by
the mass import script and by the mail gateway

### DIFF
--- a/account_invoice_import/scripts/inotify-sample.sh
+++ b/account_invoice_import/scripts/inotify-sample.sh
@@ -1,0 +1,24 @@
+#! /bin/sh
+# -*- coding: utf-8 -*-
+# Copyright 2019 Akretion France (http://www.akretion.com/)
+# @author: Alexis de Lattre <alexis.delattre@akretion.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+#
+# This script is designed to run under Linux
+# To install the required tools, run:
+# sudo apt install inotify-tools
+
+# CUSTOMIZE the 2 variables below
+INVOICE_DIR="/home/alexis/invoices2odoo"
+MASS_INVOICE_IMPORT_SCRIPT="/home/alexis/edi/account_invoice_import/scripts/mass_invoice_import.py"
+
+echo "Start to monitor $INVOICE_DIR for new files"
+
+inotifywait -m -e create -e moved_to --format "%e %w%f" "$INVOICE_DIR" |
+    while read event filepath; do
+        if [ -f "$filepath" ]; then
+            echo "File '$filepath' appeared via $event"
+            # do something with the file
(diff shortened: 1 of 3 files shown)


account_invoice_import/scripts/inotify-sample.sh
```diff
@@ -0,0 +1,24 @@
...
+# @author: Alexis de Lattre <alexis.delattre@akretion.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+#
+# This script is designed to run under Linux
+# To install the required tools, run:
+# sudo apt install inotify-tools
+
+# CUSTOMIZE the 2 variables below
+INVOICE_DIR="/home/alexis/invoices2odoo"
+MASS_INVOICE_IMPORT_SCRIPT="/home/alexis/edi/account_invoice_import/scripts/mass_invoice_import.py"
```
rvalyi, line 13: Having to alter the codebase because of such hardcoded variable is not a good practice... What about taking it from an environment variable? Eventually at least set the variable only if not defined like https://unix.stackexchange.com/a/122848

rvalyi (changes requested):
I commented here https://github.com/OCA/edi/pull/84/files#r257889039

account_invoice_import/scripts/inotify-sample.sh
alexis-via: Thanks for your remark, I'll try to improve that.
I also plan some improvements for easier multi-company support with the same Odoo user.

rvalyi: yes, I see a few lines later that you also have the user and its password hardcoded. I think the same can be done with environment variables again. Eventually there is simply a shared secret (environment variable) used as an API key instead of a user password, I'm not sure which one is better, it depends a bit on what the ERP user does exactly but eventually this is safer than throwing the admin password elsewhere like that which increases the risk of leak and make us look poor contenders when we then troll Odoo SA about security on Twitter :-)

rvalyi: @alexis-via it's worth looking at this OCA API key module as it may be a clean way to authenticate your import service: https://github.com/OCA/server-auth/tree/12.0/auth_api_key (also available in v10)

rvalyi: You may think [auth_api_key](https://github.com/OCA/server-auth/tree/12.0/auth_api_key) would make your code a lot more complex. But in fact this is a very small module... It's also what we use in ShopInvader BTW.

alexis-via: @rvalyi I think you made a misunderstanding: this script is not Odoo code, it's a small script to run on your laptop ; the script connects to Odoo to send the supplier invoices to import. That's why few variables are hard-coded in the script. Of course, I wouldn't do that if it was Odoo code !

PR #87, opened 2019-02-21
[10.0][FIX] Add PyPDF2 to travis config file

astirpe:


astirpe:
@pedrobaeza can this one be fast tracked? All the pending 10.0 PRs are red at this moment...

pedrobaeza (approved):
Although the proper way to do it is to add a requirements.txt file in the root.

PR #94, opened 2019-03-13
[12.0][MIG] base_ubl

MiquelRForgeFlow:
Standard migration.


base_ubl/__manifest__.py
```diff
@@ -0,0 +1,19 @@
...
+{
+    'name': 'Base UBL',
+    'version': '12.0.1.0.0',
+    'category': 'Hidden',
+    'license': 'AGPL-3',
+    'summary': 'Base module for Universal Business Language (UBL)',
+    'author': 'Akretion,Odoo Community Association (OCA)',
+    'website': 'https://github.com/oca/edi/',
+    'depends': [
+        'product_uom_unece',
```
astirpe:
```suggestion
        'uom_unece',
```

base_ubl/models/ubl.py
```diff
@@ -0,0 +1,695 @@
...
+
+    def ubl_parse_product(self, line_node, ns):
+        barcode_xpath = line_node.xpath(
+            "cac:Item/cac:StandardItemIdentification/cbc:ID[@schemeID='GTIN']",
+            namespaces=ns)
+        code_xpath = line_node.xpath(
+            "cac:Item/cac:SellersItemIdentification/cbc:ID", namespaces=ns)
+        product_dict = {
+            'barcode': barcode_xpath and barcode_xpath[0].text or False,
+            'code': code_xpath and code_xpath[0].text or False,
```
astirpe:
```suggestion
            'default_code': code_xpath and code_xpath[0].text or False,
```

astirpe (changes requested):

base_ubl/__manifest__.py
pedrobaeza: Now called `uom_unece`

Is it community-data-files already in oca_dependencies.txt?

base_ubl/models/ubl.py
```diff
@@ -326,8 +326,8 @@ def _ubl_add_line_item(
     @api.model
     def _ubl_add_item(
             self, name, product, parent_node, ns, type='purchase',
-            seller=False, version='2.1'):
-        '''Beware that product may be False (in particular on invoices)'''
+            seller=None, version='2.1'):
```
pedrobaeza: Why change this? This is not a reference variable

MiquelRForgeFlow: Well, it should be an empty record of `self.env['product.supplierinfo']` to be precise

pedrobaeza: But then `None` and `False` is the same, and you can't put that reference variable as you can imagine. The point is to not touch here if there's no better option.

MiquelRForgeFlow: What problem do you have with `None`? Why should it be `False`? Both functionally are the same. I just preferred `None` because this way in line 340, in `seller.id`, I don't receive stupid PyCharm message of "Unresolved attribute reference 'id' for class 'bool'".

pedrobaeza: I don't have any problem with `None`. The problem is to change things that are not needed. You don't know if dependent modules are checking this in a strict way (`seller is False`), so if no substantial benefit from changing this, then don't do it for avoiding potential problems. I know that other modules while being migrated can change this, but save pain if not needed.

pedrobaeza: About your PyCharm error, is due to the IDE, not because of the code. Python regular linters are not used to ORM framework.

MiquelRForgeFlow: Ok.

> I know that other modules while being migrated can change this, but save pain if not needed.

But it's not a pain, it's just improving code. And also gives entertainment, because migrations are very boring. 😜

base_ubl/__manifest__.py
MiquelRForgeFlow: yes, it is.

base_ubl/models/ubl.py
pedrobaeza: If you promise you'll check all the dependent modules, then keep the change.

base_ubl/models/ubl.py
MiquelRForgeFlow: should this change be backported to v11?

MiquelRForgeFlow:
Ready, changes applied

pedrobaeza (approved):

astirpe (approved):

PR #100, opened 2019-04-03
[12.0][MIG] .travis.yml: remove TRANSIFEX + add MAKEPOT

astirpe:


pedrobaeza:
Let's fast-track this without waiting Travis result, which will be the same as before. The pity is that you have launched a lot of PRs already that will hammer Travis with each extra build. A rebase on them will cancel previous builds if not yet started.

astirpe:
None of the ongoing builds were started, so I stopped them. Going to rebase soon. But seems travis is very slow this morning.

pedrobaeza:
I think Travis has reduce the free slots, and this is happening since several days. I have to wait till the following day to find some CI results. This depends also on the OCA activity of the day.

PR #105, opened 2019-04-30
[12.0][FIX] account_invoice_ubl: related field must be not readonly

astirpe:
This PR fixed the following issue:

- Go to menu *Invoicing > Configuration > Settings > Invoicing*.
- Under *Electronic Invoices*, the field "*Embed PDF in UBL XML Invoice*" is not selectable.

Field "*Embed PDF in UBL XML Invoice*"  is a related field and must be set as `readonly=False`.


pedrobaeza (approved):
Fast-tracking this

PR #108, opened 2019-05-20
[10.0][FIX]Fix supplier search for variant

luc-demeyer:


base_ubl/models/ubl.py
```diff
@@ -344,7 +344,8 @@ def _ubl_add_item(
                 if seller:
                     sellers = self.env['product.supplierinfo'].search([
                         ('name', '=', seller.id),
-                        ('product_tmpl_id', '=', product.product_tmpl_id.id)])
+                        ('product_tmpl_id', '=', product.product_tmpl_id.id),
```
tbaden: is there even a need for the template id if you use the variant id?

luc-demeyer:
@tbaden 
Thanks for your feedback. 
After a bit of investigation:  
product_tmpl_id must always be there, hence I think it's better to keep that part of the domaint (a product.supplierinfo record without product_tmpl_id implies a corrupted data structure).

But while investigating I amended this PR since we need to search on product_id when variants are used and product_id when variants are not used.

BenjaHe (approved):

pedrobaeza:
Please use this method instead:

https://github.com/odoo/odoo/blob/0c912354a959010e51376ad1b9dab9c1fa9e620d/addons/product/models/product.py#L514

luc-demeyer:
@pedrobaeza 
Thanks a lot for this suggestion. I'll have a look and adapt this PR accordingly.

rousseldenis (approved):
Code review.

PR #85, opened 2019-02-20
[12.0] Migrate base_business_document_import*

alexis-via:
Migrate base_business_document_import from v11 to v12
Migrate base_business_document_import_phone from v10 to v12
Convert README to new format

PR #110, opened 2019-05-31
[MIG] 12.0-mig-base-business-document-import

randall-vx:


emagdalenaC2i:
@randall-vx there is already a PR to migrate base_business_document_import. See #85

PR #112, opened 2019-06-06
[10.0][FIX] Fill the account_id (required) for every configuration

acsonefho:
Currently, if you select a "single" method to import invoice and without product, you'll have an error because the account_id is not filled.

### DIFF
--- a/account_invoice_import/wizard/account_invoice_import.py
+++ b/account_invoice_import/wizard/account_invoice_import.py
@@ -617,10 +617,13 @@ class AccountInvoiceImport(models.TransientModel):
             self, diff_amount, invoice, import_config):
         ailo = self.env['account.invoice.line']
         prec = invoice.currency_id.rounding
+        account = import_config.get(
+            'account', self.env['account.account'].browse())
         il_vals = {
             'name': _('Adjustment'),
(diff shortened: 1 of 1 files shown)


rousseldenis (approved):
Code review

RobinetDenisAcsone (approved):
Code review

astirpe (approved):

acsonefho:
@rousseldenis Can you merge this one please? It seems ok now

rousseldenis:
@acsonefho I've no merge rights here. @astirpe

alexis-via:
I don't understand this PR: if the import method is single_*, you should never need an adjustment line ! Because, when you use a single_* method, the total without tax is written in the price field and qty = 1... So, could you tell us more about the exact bug scenario ?

acsonefho:
> I don't understand this PR: if the import method is single__, you should never need an adjustment line ! Because, when you use a single__ method, the total without tax is written in the price field and qty = 1... So, could you tell us more about the exact bug scenario ?

Sorry but I don't understand your comment. I'm talking about a missing `account_id` to create the invoice line (probably because the `product_id` is not filled).

PR #113, opened 2019-07-03
[10.0][FIX] account_invoice_import fix invoice update during import

acsonefho:
**Bug:**
During the import of invoice, it seems that the update of existing invoices was implemented but not triggered automatically.

I also add the `_get_existing_invoices` method to add inherit point to find invoices.


Cedric-Pigeon:
@acsonefho Please check Travis

PR #118, opened 2019-07-25
.travis.yml: fix as per new template

astirpe:
Travis in branch 12.0 is broken. Trying to make it green again.

PR #119, opened 2019-07-25
.travis.yml: Fix it with new template

astirpe:
Backport of https://github.com/OCA/edi/pull/118. Cherry-picked from a2671fa31086f894bb6d665f0cb3c447017faad8

PR #120, opened 2019-07-25
.travis.yml: Fix it with new template

astirpe:
Backport of #118. Cherry-picked from a2671fa.
Also add MAKEPOT.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ python:
   - "2.7"
 
 addons:
+  postgresql: "9.6"
   apt:
     packages:
       - expect-dev  # provides unbuffer utility
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,36 +10,21 @@ python:
 
 addons:
   apt:
-# only add the two lines below if you need wkhtmltopdf for your tests
-    sources:
-      - pov-wkhtmltopdf
-#    Search your sources alias here:
-#      https://github.com/travis-ci/apt-source-whitelist/blob/master/ubuntu.json
     packages:
       - expect-dev  # provides unbuffer utility
       - python-lxml  # because pip installation is slow
       - python-simplejson
       - python-serial
       - python-yaml
-#     Search your packages here:
-#       https://github.com/travis-ci/apt-package-whitelist/blob/master/ubuntu-precise
-      - wkhtmltopdf  # only add if needed and check the before_install section below
-
-# set up an X server to run wkhtmltopdf.
-before_install:
-  - "export DISPLAY=:99.0"
-  - "sh -e /etc/init.d/xvfb start"
 
 env:
   global:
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Same problem about PG 9.6

astirpe:
Now Travis fails for the following error:

```
The XML file is invalid against the XML Schema Definition
XSD Error: Element '{urn:un:unece:uncefact:data:standard:ReusableAggregateBusinessInformationEntity:100}TypeCode': [facet 'enumeration'] The value '31' is not an element of the set {'10', '20', '30', '42', '48', '49', '57', '58', '59', '97', 'ZZZ'}., line 44
No valid XML file found in the PDF
FAIL

```


I think I need help for this, as we don't use factur-x.

pedrobaeza:
The real problem is the `assert` you have changed:
```
2019-07-25 12:56:16,048 5824 ERROR openerp_test odoo.addons.account_invoice_factur-x.tests.test_facturx_invoice: ` AssertionError: None != 'factur-x.xml'
```
https://travis-ci.org/OCA/edi/jobs/563533926#L1198

astirpe:
Before that change on the `assert`, this was the error:

```
2019-07-25 12:13:07,990 5726 ERROR openerp_test odoo.addons.account_invoice_factur-x.tests.test_facturx_invoice: ` AssertionError: None is not true : factur-x.xml
```

https://travis-ci.org/OCA/edi/jobs/563522056#L1198

pedrobaeza:
Well, then the problem seems to be that it should be something, not `None`. @alexis-via please check the problem.

astirpe:
@pedrobaeza Even if this PR ends in red, the travis failure is caused by just one test of module account_invoice_factur-x and not by the travis config.
Still this PR improves the situation: it fixes the error that is actually blocking all the PRs on branch 10.0 (see for example https://travis-ci.org/OCA/edi/builds/554320657?utm_source=github_status&utm_medium=notification). You can notice that the error is preventing every test and lint check to be executed; this PR is a fix for this.

I would ask to merge it as it is, so that the pending PRs can be rebased.
Also consider that PRs #118 and #119 were already merged and this one is just a backport.

pedrobaeza:
Then put the test as it was, or we will have a red Travis.

astirpe:
@pedrobaeza done.

pedrobaeza:
Uhm, but Travis is still failing... so we don't get the green status we need. I can merge it, but we will still have the same red status for the rebased PRs. @alexis-via can you please check the problem?

astirpe:
@pedrobaeza that's what I was trying to say in https://github.com/OCA/edi/pull/120#issuecomment-516723233. Even the this PR will not make the travis green, at least it will allow travis to run the tests.

pedrobaeza:
I would bet for commenting that line for now.

PR #145, opened 2019-11-07
[10.0][IMP] account_invoice_import_ubl remove optional listAgencyID of InvoiceTypeCode to parse the invoice type

acsonefho:
**Current behavior**
We receive XML invoices and the invoice type is not correct because we don't have the tag `listAgencyID`.

**Fix**
As the tag (`listAgencyID`) seems optional, just remove it to read the `InvoiceTypeCode`.
(+ Adapt the unit test to ensure it's correctly parsed.)

I really don't know why Travis fail. I check the issue about partner_id is missing but I think it comes from another module.


acsonefho:
@alexis-via Can I have your opinion about that please? :)
Thanks!

alexis-via (approved):

alexis-via:
Fine for me.

PR #149, opened 2019-12-12
[MIG] base_zugferd: Migration to 11.0 (base_factur-x)

yucer:
#70 with the proper rename

PR #150, opened 2019-12-12
[11.0][MIG] base_zugferd: Migration to 11.0 (base_factur-x) 

yucer:
A new attempt of #149 . Now including only the addon to be migrated. 

The last commit is just the last state of the repository files and can be discarded when merging although it should give no conflicts.


yucer:
Last try in #158.

PR #151, opened 2019-12-12
[FIX] 10.0 bad use of play_onchange

gurneyalex:
The method `play_onchange` in `server-tools/onchange_helper`
only returns the changed keys in the dictionary it receives as first arguement (see https://github.com/OCA/server-tools/blob/10.0/onchange_helper/models/models.py#L106-L138)
The code in `account_invoice_import`, `purchase_order_import` and `sale_order_import`
was written as if all the keys were returned. Maybe the behavior of play_onchange
was updated in a recent version, but this nevertheless needs fixing in this module.
    
This should fix the unit tests which are red on this branch.

Also remove a method defined twice in `purchase_order_import`. 

Also fix purchase_order_import which would not add the unit price (maybe also related to the play_onchange changes)

### DIFF
--- a/account_invoice_import/wizard/account_invoice_import.py
+++ b/account_invoice_import/wizard/account_invoice_import.py
@@ -205,7 +205,7 @@ class AccountInvoiceImport(models.TransientModel):
             'journal_id': journal_id,
             'invoice_line_ids': [],
         }
-        vals = aio.play_onchanges(vals, ['partner_id'])
+        vals.update(aio.play_onchanges(vals, ['partner_id']))
         vals['invoice_line_ids'] = []
         # Force due date of the invoice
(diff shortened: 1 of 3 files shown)


gurneyalex:
@lmignon it seems you worked on https://github.com/OCA/server-tools/commit/3b6dd12bcd4ad6c890a1846ad40877f3f2a8e11f do you confirm that the behaviour of the play_onchange method changed?

lmignon:
@gurneyalex You are right. The behavior of the play_onchange method changed. The new implementation is much faster and take benefit of the cache.

lmignon (approved):
LGTM (code review)

guewen (approved):

alexis-via (approved):

PR #158, opened 2019-12-13
[11.0][MIG] base_zugferd: Migration to 11.0 (base_factur-x)

yucer:
A new attempt of #150 , with the rename of the module references in the translation files and the path of the xsd files to load.


yucer:
Did a mistake here... closing...

PR #170, opened 2020-02-10
[12.0] Port Factur-X generation stack from v10 to v12

alexis-via:
This PR depends on https://github.com/OCA/community-data-files/pull/83 and https://github.com/OCA/edi/pull/169


alexis-via:
One thing I forgot to mention: I updated the name of the module to use "facturx" instead of "factur-x". In fact, it is not a good thing to use a dash in the name of a module because you can't import it:
from odoo.addons.account_invoice_factur-x ==> it will throw an error.
For the module account_e-invoice_generate, it's too late, we should rename it for v13.

PR #173, opened 2020-02-13
[12.0][MIG] account_invoice_import_invoice2data

alexis-via:


alexis-via:
Incredible ! Green Travis with invoice2data on the first Travis run ! I didn't expect it to be so easy this time! Hopefully, Travis has a more recent OS now, which makes things much easier.

PR #187, opened 2020-04-03
Fix travis config

sbidoul:


.travis.yml
```diff
@@ -9,22 +9,8 @@ python:
...
   postgresql: "9.6"
   apt:
-# only add the two lines below if you need wkhtmltopdf for your tests
-    sources:
-      - pov-wkhtmltopdf
-#    Search your sources alias here:
-#      https://github.com/travis-ci/apt-source-whitelist/blob/master/ubuntu.json
     packages:
-      - expect-dev  # provides unbuffer utility
-      - python-lxml  # because pip installation is slow
```
bealdav, line 19: We need lxml in this repo.
You suggest to install it with pip ?

bealdav (approved):
OK

.travis.yml
```diff
@@ -38,20 +24,11 @@ jobs:
...
-    # be installed in the same database. And you get a huge speed penalty in your
-    # tests)
-    #  - TESTS="1.0" ODOO_REPO="odoo/odoo" UNIT_TEST="1"
-    #  - TESTS="1.0" ODOO_REPO="OCA/OCB" UNIT_TEST="1"
 
 env:
   global:
     - VERSION="8.0" TESTS="0" LINT_CHECK="0" MAKEPOT="0"
-    - WKHTMLTOPDF_VERSION="0.12.1"
+    - WKHTMLTOPDF_VERSION="0.12.5"
```
bealdav, line 31: I thought that each Odoo version had its matching wkhtml version, ok for that

.travis.yml
sbidoul: lxml is an Odoo [dependency](https://github.com/odoo/odoo/blob/9e8f70e4849b0eeaca8b5cf51372ecfa23dc561b/requirements.txt#L17), so MQT will install automatically, and pip will cache it so performance should not be an issue..

.travis.yml
sbidoul: Yes, but I've no better solution to offer and I guess for running tests it should be good enough.

bealdav:
@sbidoul Thanks a lot for this fix

bealdav:
@sbidoul if you can please merge this PR. Thanks a lot

PR #192, opened 2020-04-20
[12.0] add missing factur-x dependency

sbidoul:
Let's see if travis is happy with this

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: python
-sudo: true 
+sudo: true
 cache:
   apt: true
   directories:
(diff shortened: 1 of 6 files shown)


sbidoul:
Travis is red saying: `Modules "Account e-invoice Generate" and "Import Vendor Bills From XML" are incompatible.`.

sbidoul:
I guess we need to uninstall Odoo's `account_facturx`, not sure if we've ever had to do that with MQT.

bealdav:
Here is what I have done in l10n-fr https://github.com/OCA/l10n-france/blob/12.0/.travis.yml#L26

As said here https://github.com/OCA/l10n-france/pull/214

> Maybe remove auto_install of account_facturx module in OCB

and then only test with OCB ...

Maybe you have a better idea ?

sbidoul:
@bealdav yes I made a commit to get Odoo's `account_facturx` out of the way.

Now there seem to be an error in demo data: `account_invoice_import_facturx.fr_vat_purchase_20`, trying to fix it.

sbidoul:
Hm, this 12.0 branch is in a sad state... now there is a bug somewhere in `embed_xml_in_pdf`... 

@alexis-via, help!

astirpe:
@sbidoul the error you mentioned should be fixed by https://github.com/OCA/edi/pull/195

alexis-via (approved):

PR #193, opened 2020-04-21
[FIX]Restore new api compatibility to get_pdf method

florian-dacosta:
@sbidoul @alexis-via @bealdav 

This api.v7 break the compatibility with new api call to get_pdf.
To address this issue, I do the same as in the official report module.

### DIFF
--- a/purchase_order_ubl/models/report.py
+++ b/purchase_order_ubl/models/report.py
@@ -37,5 +37,5 @@ class Report(models.Model):
 
     @api.v8  # noqa
     def get_pdf(self, records, report_name, html=None, data=None):
-        return Report.get_pdf(self._model, self._cr, self._uid, records.ids,
+        return self._model.get_pdf(self._cr, self._uid, records.ids,
                               report_name, html=html, data=data, context=self._context)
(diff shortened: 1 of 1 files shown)


bealdav (approved):

bealdav:
please @astirpe could you have a look. Thanks

purchase_order_ubl/models/report.py
```diff
@@ -34,3 +34,8 @@ def get_pdf(
             pdf_content = order.embed_ubl_xml_in_pdf(
                 pdf_content)
         return pdf_content
+
+    @api.v8  # noqa
+    def get_pdf(self, records, report_name, html=None, data=None):
+        return Report.get_pdf(self._model, self._cr, self._uid, records.ids,
```
astirpe:
```suggestion
        return self._model.get_pdf(self._cr, self._uid, records.ids,
```

This is a cleaner way IMO. Anyway it's not blocking.

astirpe (approved):

florian-dacosta:
@astirpe done, thanks!

bealdav:
@alexis-via branch is now red. 
I can fix that. Don't you think that I could completely convert api.v7 to new api here ?

Thanks for your feed back.

bealdav:
solved by https://github.com/OCA/edi/pull/194

PR #197, opened 2020-04-29
[10.0][FIX] account_invoice_import_factur-x: facturx Python module does not exist

SimoRubi:
```
$ pip install facturx
DEPRECATION: Python 2.7 reached the end of its life on January 1st, 2020. Please upgrade your Python as Python 2.7 is no longer maintained. A future version of pip will drop support for Python 2.7. More details about Python 2 support in pip, can be found at https://pip.pypa.io/en/latest/development/release-process/#python-2-support
ERROR: Could not find a version that satisfies the requirement facturx (from versions: none)
ERROR: No matching distribution found for facturx
```
This is also why Travis is currently not able to build the test environment.

### DIFF
--- a/account_invoice_factur-x/__manifest__.py
+++ b/account_invoice_factur-x/__manifest__.py
@@ -16,7 +16,7 @@
         'base_zugferd',
         'base_vat',
         ],
-    'external_dependencies': {'python': ['facturx']},
+    'external_dependencies': {'python': ['factur-x']},
     'data': [
         'views/res_partner.xml',
         'views/account_config_settings.xml',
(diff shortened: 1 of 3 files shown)


account_invoice_factur-x/__manifest__.py
```diff
@@ -16,7 +16,7 @@
         'base_zugferd',
         'base_vat',
         ],
-    'external_dependencies': {'python': ['facturx']},
+    'external_dependencies': {'python': ['factur-x']},
```
sbidoul, line 19: That would be correct but in v13 only, which allows declaring the PyPI distribution name here. Until 12.0, it must be the import name, and the mapping must go in setup.py. You probably need to backport #192

alexis-via:
Thanks for your explanation Stéphane !

PR #198, opened 2020-05-07
[8.0][FIX] purchase_order_ubl : wrong args used in super

bguillot:
Hi

It is not the good argument that is passed to the super().

Thanks for the review.


bealdav (approved):

bguillot:
Actually I think this is needed I close my PR

PR #199, opened 2020-05-08
[10.0] FIX #104 base_ubl: use address of partner, not of commercial_parter

### LINKED ISSUE #104: [CLARIFY] Why is the invoicing address taken from the commercial partner?
Regarding the module `base_ubl`:

In the code used to generate the customer section (the `AccountingCustomerParty`, the partner used for the address details is the partner's commercial partner, not the partner themselves.

We have noticed that this means that if the invoicing address is truly different from the company it is related to (perhaps the company has several different invoicing addresses), then the address selected onto the invoice is not respected: the invoice of the parent is used instead.

The decision is explicitly made [here](https://github.com/OCA/edi/blob/11.0/base_ubl/models/ubl.py#L170), but no comment or docstring has been supplied explaining why it is done like this.

The module was originally authored by @alexis-via : might he be able to shed some light on why things are done this way?

I am trying to figure out whether I need to hack some kind of patch into our own system or try to make a fix PR here. If this is the expected behaviour in the rest of the world, and I'd end up breaking everybody else's installations, I probably shouldn't go for it.

alexis-via:
Add legal entity block when commercial_partner.is_company = True

### DIFF
--- a/base_ubl/models/ubl.py
+++ b/base_ubl/models/ubl.py
@@ -186,10 +186,10 @@ class BaseUbl(models.AbstractModel):
         if partner.lang:
             self._ubl_add_language(partner.lang, party, ns, version=version)
         self._ubl_add_address(
-            commercial_partner, 'PostalAddress', party, ns, version=version)
+            partner, 'PostalAddress', party, ns, version=version)
         self._ubl_add_party_tax_scheme(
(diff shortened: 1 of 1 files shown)


jgrandguillaume (approved):
Agree with this

PR #201, opened 2020-05-14
[13.0][MIG] sale_order_import

sebalix:
Based on 12.0 migration PR: https://github.com/OCA/edi/pull/101

Depends on:
- [x] https://github.com/OCA/edi/pull/200
- [x] https://github.com/OCA/sale-workflow/pull/1023

PR #202, opened 2020-05-14
[13.0][MIG] sale_order_import_ubl

sebalix:
Depends on:
- [x] https://github.com/OCA/edi/pull/201

I have to change the way the ZIP codes are formatted in the demo data, otherwise partners didn't match... This is caused "because" of this line: https://github.com/OCA/edi/blob/13.0/base_ubl/models/ubl.py#L659
but I do not understand why tests were not failing on previous versions as this code is there from at least 8.0... any idea?


sale_order_import_ubl/wizard/sale_order_import.py
```diff
@@ -0,0 +1,137 @@
...
+    @api.model
+    def parse_xml_order(self, xml_root, detect_doc_type=False):
+        start_tag = "{urn:oasis:names:specification:ubl:schema:xsd:"
+        rfq = "RequestForQuotation"
+        if xml_root.tag == start_tag + "Order-2}Order":
+            if detect_doc_type:
+                return "order"
+            else:
+                return self.parse_ubl_sale_order(xml_root)
+        elif xml_root.tag == "{}{}-2}}{}".format(start_tag, rfq, rfq):
```
i-vyshnevska, line 21: what does this "2" for?

Tonow-c2c (approved):

jgrandguillaume (approved):
LGMT

sale_order_import_ubl/wizard/sale_order_import.py
sebalix: I don't know, it respects the UBL specs I suppose, I didn't change that.

desdelinux (approved):
functional review is ok

Issue #196, opened 2020-04-29
XML is invalid against the XML Schema Definition


nitohu:
Hello I was trying to generate a PDF file with an embedded XML (ZUGFeRD 2.0 Standard).
I wanted to use the `account_invoice_facturx` OCA module as a basis for this.

module: account_invoice_facturx
version: 12.0.1.0.0

**Current behaviour:**
When clicking on the "Print" -> "Invoice" button in the Invoice form inside Odoo, I'm getting the error that the XML is invalid against the XML Schema Definition because another field is expected than given.
```
factur-x: The XML file is invalid against the XML Schema Definition
factur-x: XSD Error: Element '{urn:un:unece:uncefact:data:standard:ReusableAggregateBusinessInformationEntity:100}SpecifiedTradeSettlementLineMonetarySummation': This element is not expected. Expected is ( {urn:un:unece:uncefact:data:standard:ReusableAggregateBusinessInformationEntity:100}ApplicableTradeTax )., line 35
```

### Traceback:
```Odoo Server Error
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/dist-packages/facturx/facturx.py", line 144, in check_facturx_xsd
    official_schema.assertValid(t)
  File "src/lxml/etree.pyx", line 3621, in lxml.etree._Validator.assertValid
lxml.etree.DocumentInvalid: Element '{urn:un:unece:uncefact:data:standard:ReusableAggregateBusinessInformationEntity:100}SpecifiedTradeSettlementLineMonetarySummation': This element is not expected. Expected is ( {urn:un:unece:uncefact:data:standard:ReusableAggregateBusinessInformationEntity:100}ApplicableTradeTax )., line 32

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/mnt/c/Users/niklastom.hucke/Dev/odoo-12.0/addons/web/controllers/main.py", line 1671, in report_download
    response = self.report_routes(reportname, docids=docids, converter=converter)
  File "/mnt/c/Users/niklastom.hucke/Dev/odoo-12.0/odoo/http.py", line 519, in response_wrap
    response = f(*args, **kw)
  File "/mnt/c/Users/niklastom.hucke/Dev/odoo-12.0/addons/web/controllers/main.py", line 1612, in report_routes
    pdf = report.with_context(context).render_qweb_pdf(docids, data=data)[0]
[...]

PR #204, opened 2020-06-01
[FIX] Tax is mandatory on invoice lines

BT-aestebanez:
Fix for https://github.com/OCA/edi/issues/196

The standard does require at least one tax per invoice line as defined in the xsd:
```
  <xs:complexType name="LineTradeSettlementType">
    <xs:sequence>
      <xs:element name="ApplicableTradeTax" type="ram:TradeTaxType"/>
      <xs:element name="BillingSpecifiedPeriod" type="ram:SpecifiedPeriodType" minOccurs="0"/>
      <xs:element name="SpecifiedTradeAllowanceCharge" type="ram:TradeAllowanceChargeType" minOccurs="0" maxOccurs="unbounded"/>
      <xs:element name="SpecifiedTradeSettlementLineMonetarySummation" type="ram:TradeSettlementLineMonetarySummationType"/>
      <xs:element name="AdditionalReferencedDocument" type="ram:ReferencedDocumentType" minOccurs="0"/>
      <xs:element name="ReceivableSpecifiedTradeAccountingAccount" type="ram:TradeAccountingAccountType" minOccurs="0"/>
    </xs:sequence>
  </xs:complexType>
```

The element  `<xs:element name="ApplicableTradeTax" type="ram:TradeTaxType"/>` does not have  "minOccurs" set and therefore, the element is required (source https://www.w3schools.com/xml/schema_complex_indicators.asp)

### DIFF
--- a/account_invoice_facturx/models/account_invoice.py
+++ b/account_invoice_facturx/models/account_invoice.py
@@ -629,30 +629,34 @@ class AccountInvoice(models.Model):
         line_trade_settlement = etree.SubElement(
             line_item, ns['ram'] + 'SpecifiedLineTradeSettlement')
 
-        if iline.invoice_line_tax_ids:
-            for tax in iline.invoice_line_tax_ids:
-                trade_tax = etree.SubElement(
-                    line_trade_settlement,
-                    ns['ram'] + 'ApplicableTradeTax')
-                trade_tax_typecode = etree.SubElement(
-                    trade_tax, ns['ram'] + 'TypeCode')
-                if not tax.unece_type_code:
-                    raise UserError(_(
-                        "Missing UNECE Tax Type on tax '%s'")
-                        % tax.display_name)
-                trade_tax_typecode.text = tax.unece_type_code
-                trade_tax_categcode = etree.SubElement(
-                    trade_tax, ns['ram'] + 'CategoryCode')
-                if not tax.unece_categ_code:
-                    raise UserError(_(
-                        "Missing UNECE Tax Category on tax '%s'")
-                        % tax.display_name)
-                trade_tax_categcode.text = tax.unece_categ_code
-                # No 'DueDateTypeCode' on lines
-                if tax.amount_type == 'percent':
-                    trade_tax_percent = etree.SubElement(
-                        trade_tax, ns['ram'] + 'RateApplicablePercent')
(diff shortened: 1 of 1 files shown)


BT-aestebanez:
Hello @alexis-via 
First of all thanks for contributing with this work.

Could you review this PR?

Thanks,
Alvaro

account_invoice_facturx/models/account_invoice.py
```diff
@@ -629,30 +629,34 @@ def _cii_add_invoice_line_block(
...
-                        % tax.display_name)
-                trade_tax_categcode.text = tax.unece_categ_code
-                # No 'DueDateTypeCode' on lines
-                if tax.amount_type == 'percent':
-                    trade_tax_percent = etree.SubElement(
-                        trade_tax, ns['ram'] + 'RateApplicablePercent')
-                    trade_tax_percent.text = '%0.*f' % (2, tax.amount)
+        if not iline.invoice_line_tax_ids:
+            raise UserError(_(
+                "The Factur-X standard specify that each line needs, "
```
thomaspaulb: 'specifies'.

thomaspaulb (approved):

account_invoice_facturx/models/account_invoice.py
BT-aestebanez: Thanks, applied

alexis-via:
in my v14 PR to migrate account_invoice_facturx https://github.com/OCA/edi/pull/288
I solved the problem when there is no tax on the invoice line, but I made big changes to the code that generate the tax block. Maybe we could backport it to v12. Could you have a look at it and tell me what you think ?

alexis-via:
I forgot to mention that backporting is not easy due to the revolution on invoices in v13... tell me what you think first and then we'll decide.

thomaspaulb:
Specifically this part, which adds an empty tax block if tax is not filled, right?

image

I'd say backporting that is an honourable but cumbersome job, and could be done later, on top of this small fix.

BT-aestebanez:
> Specifically this part, which adds an empty tax block if tax is not filled, right?
> 
> image
> 
> I'd say backporting that is an honourable but cumbersome job, and could be done later, on top of this small fix.

I agree :)
Sorry for the late reply

PR #166, opened 2020-01-13
[10.0][FIX] base_ubl: PartyName node is not mandatory in xsd so it should not be mandatory in code either.

SimoRubi:
I have edited one of the test files to allow testing this change, please advise me if I should test it in any other way.

PR #206, opened 2020-06-04
[FIX] base_ubl: PartyName fix migration to 12.0

SimoneVagile:
See: https://github.com/OCA/edi/pull/166


sale_order_import_ubl/tests/files/UBL-Order-2.0-Example_no-party-name.xml
```diff
@@ -0,0 +1,177 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
SimoRubi, line 1: The module `sale_order_import_ubl` does not exist in `12.0` yet, here you are just adding this file and the one below, this is not correct.
You should wait for https://github.com/OCA/edi/pull/102 to be merged before proposing this change

SimoRubi (changes requested):
Thanks for your contribution!
Unfortunately, this has to wait until https://github.com/OCA/edi/pull/102 is merged

SimoneVagile:
Depends on:

#102

SimoneVagile:
Depends on:

- [ ] #102

sale_order_import_ubl/tests/files/UBL-Order-2.0-Example_no-party-name.xml
SimoneVagile: https://github.com/OCA/edi/pull/206#issuecomment-638769051

PR #208, opened 2020-06-09
[13.0][MIG] account_invoice_facturx and account_invoice_facturx_py3o

alexis-via:
Up-port improvements of account_e-invoice_generate from v12 to v13


gurneyalex:
@alexis-via https://github.com/akretion/edi/pull/1 fixes the pre-commit checks on travis

account_invoice_facturx/__manifest__.py
```diff
@@ -0,0 +1,29 @@
...
+    'author': 'Akretion,Odoo Community Association (OCA)',
+    'maintainers': ['alexis-via'],
+    'website': 'https://github.com/OCA/edi',
+    'depends': [
+        'account_e-invoice_generate',
+        'account_payment_partner',
+        'base_facturx',
+        'base_vat',
+        ],
+    'external_dependencies': {'python': ['facturx']},
```
sbidoul:
```suggestion
    'external_dependencies': {'python': ['factur-x']},
```

In v13 we can put the PyPI package name here.

account_invoice_facturx_py3o/__manifest__.py
```diff
@@ -0,0 +1,17 @@
...
+    'name': 'Py3o Factur-x Invoice',
+    'version': '13.0.1.0.0',
+    'category': 'Invoicing Management',
+    'license': 'AGPL-3',
+    'summary': 'Generate Factur-X invoices with Py3o reporting engine',
+    'author': 'Akretion,Odoo Community Association (OCA)',
+    'maintainers': ['alexis-via'],
+    'website': 'https://github.com/OCA/edi',
+    'depends': ['account_invoice_facturx', 'report_py3o'],
+    'external_dependencies': {'python': ['facturx']},
```
sbidoul:
```suggestion
    'external_dependencies': {'python': ['factur-x']},
```

pedrobaeza:
FYI, Odoo has included UNECE codes in v14 for core factur-x.

alexis-via:
Seems I forgot the PR for base_facturx... I have it on my laptop, I'll do the PR soon.

alexis-via:
@pedrobaeza Great, at last. I need to have a look at it.
Their Factur-x implementation is so broken in v13... there is not even the invoice number in factur-x.xml !!! Even once it is fixed, their XML doesn't comply with the XSD... it's a shame.

pedrobaeza:
Ouch, that's a shame... I think proper bug reports can guide the situation to be fixed.

alexis-via:
@pedrobaeza I found these issues a few days ago. I'll try to find some time in the coming days to make some bug reports.

alexis-via:
Depends on PR to migrate base_facturx https://github.com/OCA/edi/pull/209

etobella:
@alexis-via This PR makes runbot fail, Can you review?

alexis-via:
Yes. But when I connect to runbot, I don't get any error in the logs, only some INFO messages. Maybe I don't know how to use Runbot properly. Do you know where I can see an error message ?

etobella:
This module excludes a base module that will be autoinstalled by odoo. This modules and the dependent modules cannot be tested on travis

etobella:
Adding the exclude has made uninstallable 6 modules. Just wondering if we need to revert this PR and propose a new PR without this issue. At the end, this will be a lot of code that will not be tested by CI

PR #209, opened 2020-06-11
[13.0] MIG base_facturx

alexis-via:
I know that this module is almost empty. But I want to keep it, because we'll certainly put more things in it in the future, when the Order-X standard will be released (supposing it is released one day...).


alexis-via:
Travis is red because base_facturx/data/zugferd_codes.xml hasn't been re-formatted. I've run "pre-commit run -a" before commit/push, but I didn't reformat this xml file. If I run it now, same result... strange !

alexis-via:
And Order-X is coming soon: http://fnfe-mpe.org/wp-content/uploads/2020/06/2020-05-31-Order-X-Press-release-EN.pdf So we need to keep this module.

pedrobaeza:
Then it should be `base_x` :stuck_out_tongue_closed_eyes:

alexis-via:
We are in OCA/edi, not in OCA/porn Pedro ! :) :)

jgrandguillaume:
@alexis-via Can you make travis happy please :) ?

PR #216, opened 2020-07-02
[IMP] . amounts decimal places should be no more than 2

badbole:
According to UBL standars and validations 
[BR-DEC-19]-The allowed maximum number of decimals for the VATcategory taxable amount (BT-116) is 2.;
[BR-DEC-20]-The allowedmaximum number of decimals for the VAT category tax amount (BT-117) is 2.    ;
[UBL-DT-01]-Amounts shall be decimal up to two fractiondigits;[UBL-DT-01]-Amounts shall be decimal up to two fraction digits

Currently with accounting precision set to 3 or more decimal places we generate invalid XML file
This fix hardcodes 2 decimal places instead of using any given precission, 

Same fix shoul be applied to base_ubl module

### DIFF
--- a/account_invoice_ubl/models/account_invoice.py
+++ b/account_invoice_ubl/models/account_invoice.py
@@ -95,6 +95,7 @@ class AccountInvoice(models.Model):
             parent_node, ns['cac'] + 'LegalMonetaryTotal')
         cur_name = self.currency_id.name
         prec = self.currency_id.decimal_places
+        prec = prec if prec <= 2 else 2
         line_total = etree.SubElement(
             monetary_total, ns['cbc'] + 'LineExtensionAmount',
             currencyID=cur_name)
@@ -126,7 +127,9 @@ class AccountInvoice(models.Model):
         dpo = self.env['decimal.precision']
         qty_precision = dpo.precision_get('Product Unit of Measure')
         price_precision = dpo.precision_get('Product Price')
+        price_precision = price_precision if price_precision <= 2 else 2
         account_precision = self.currency_id.decimal_places
+        account_precision = account_precision if account_precision <= 2 else 2
         line_id = etree.SubElement(line_root, ns['cbc'] + 'ID')
         line_id.text = str(line_number)
         uom_unece_code = False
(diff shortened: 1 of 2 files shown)


badbole:
@alexis-via could you please review and check tis request... 
if you agree i will make another one to base_ubl and other modules needed for e-invoicing
greetz

badbole:
@astirpe could you also check on this pull request?

account_invoice_ubl/models/account_invoice.py
```diff
@@ -94,7 +94,8 @@ def _ubl_add_legal_monetary_total(self, parent_node, ns, version='2.1'):
         monetary_total = etree.SubElement(
             parent_node, ns['cac'] + 'LegalMonetaryTotal')
         cur_name = self.currency_id.name
-        prec = self.currency_id.decimal_places
+        # prec = self.currency_id.decimal_places
```
astirpe: I would prefer to keep the previous lines not commented, something like this:

```suggestion
        prec = self.currency_id.decimal_places
        prec = prec if prec <= 2 else 2
```

And the same logic for the other changes you made.

astirpe (changes requested):

base_ubl/models/ubl.py
```diff
@@ -394,7 +394,8 @@ def _ubl_add_item(
     def _ubl_add_tax_subtotal(
             self, taxable_amount, tax_amount, tax, currency_code,
             parent_node, ns, version='2.1'):
-        prec = self.env['decimal.precision'].precision_get('Account')
+        # prec = self.env['decimal.precision'].precision_get('Account')
```
astirpe: Same discourse here

bealdav:
and probably commits to merge ! ?

badbole:
squashed into one commit

astirpe (approved):

bealdav:
Thanks @badbole for your contributions in this repo and whole maintainer work on several branch.
For my information and OCA business knowledge growth could you add some comments inside the code, near line 98, to explain why this code (probably copy/paste PR description) and a link where is defined on the web, the rule you explain.

This is only asked in 12.0 branch, past is past

Thanks a lot

Other subject and not mandatory you could refactor a little with this function in base_ubl

```
@api.model
def make_precision_ubl_compatible(self, precision)
    """ docstring with adhoc explanation
    """
    return precision if precision <= 2 else 2
```
Once again only for 12.0

What do you think ?

Really thanks

PR #221, opened 2020-08-03
[13.0][IMP] base_business_document_import partner search

TDu:
As there is no longer a specific separation between customer and
supplier with the `customer_rank` and `supplier_rank` fields on partner.
It is less error prone to order the partner found by there rank type
insteed of filtering.

### DIFF
--- a/base_business_document_import/models/business_document_import.py
+++ b/base_business_document_import/models/business_document_import.py
@@ -71,12 +71,13 @@ class BusinessDocumentImport(models.AbstractModel):
         company_id = self._context.get("force_company") or self.env.user.company_id.id
         domain = ["|", ("company_id", "=", False), ("company_id", "=", company_id)]
         if partner_type == "supplier":
-            domain += [("supplier_rank", ">", 0)]
+            order = "supplier_rank desc"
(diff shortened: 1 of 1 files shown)


base_business_document_import/models/business_document_import.py
```diff
@@ -71,12 +71,13 @@ def _match_partner(self, partner_dict, chatter_msg, partner_type="supplier"):
...
         if partner_type == "supplier":
-            domain += [("supplier_rank", ">", 0)]
+            order = "supplier_rank desc"
             partner_type_label = _("supplier")
         elif partner_type == "customer":
-            domain += [("customer_rank", ">", 0)]
+            order = "customer_rank desc"
             partner_type_label = _("customer")
         else:
+            order = ""
```
astirpe, line 80: Is `order = ""` working fine? Not crashing? :thinking:

TDu: I just tested it in a shell and no crash ( I mean the search with order ="")

astirpe (approved):
LGTM

jbaudoux (approved):
Thanks

PR #223, opened 2020-08-04
8.0 supplier_inventory_import_ubl: misc improvements

bealdav:
- improvement: add fields on prd.template
- fix 
- new tests

### DIFF
--- a/supplier_inventory_import_ubl/models/product.py
+++ b/supplier_inventory_import_ubl/models/product.py
@@ -2,39 +2,33 @@
 # © 2020 David BEAL @ Akretion
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
-from openerp import api, fields, models
+from openerp import _, api, fields, models
 from openerp.addons import decimal_precision as dp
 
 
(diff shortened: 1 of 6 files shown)


supplier_inventory_import_ubl/models/product.py
```diff
@@ -6,20 +6,54 @@
 from openerp.addons import decimal_precision as dp
 
 
-class ProductProduct(models.Model):
-    _inherit = "product.product"
+HELP = "Contextual info on last supplier stock updated with UBL import"
+
+
+class ProductSupplierInventoryUblMixin(models.AbstractModel):
+    _name = "product.supplier.inventory.ubl.mixin"
```
florian-dacosta: I don't really see the point in doing a mixin here since the fields are overriden for product template.

florian-dacosta (approved):
One not blocking comment, the reste seems ok

supplier_inventory_import_ubl/models/product.py
bealdav: I can remove yes

bguillot (approved):
Works in production

PR #225, opened 2020-08-07
[13.0][FIX] account_invoice_ubl customer ref

TDu:
From the little I understand of the specification, it seems that the
OrderReference can contain more than an ID element. See :
https://docs.oasis-open.org/ubl/csprd02-UBL-2.3/mod/summary/reports/UBL-Invoice-2.3.html#Table-OrderReference.Details

But this does not pass the validation and it seems that the only
accepted element in OrderReference is ID.

The goal of this PR is to add the customer reference and ID of the
sale.order being invoiced.

### DIFF
--- a/account_invoice_ubl/__manifest__.py
+++ b/account_invoice_ubl/__manifest__.py
@@ -15,6 +15,7 @@
         "account_payment_partner",
         "base_ubl_payment",
         "account_tax_unece",
+        "sale",
     ],
     "data": ["views/account_move.xml", "views/res_config_settings.xml"],
     "post_init_hook": "set_xml_format_in_pdf_invoice_to_ubl",
--- a/account_invoice_ubl/models/account_move.py
+++ b/account_invoice_ubl/models/account_move.py
@@ -39,10 +39,17 @@ class AccountMove(models.Model):
 
     def _ubl_add_order_reference(self, parent_node, ns, version="2.1"):
         self.ensure_one()
-        if self.name:
-            order_ref = etree.SubElement(parent_node, ns["cac"] + "OrderReference")
+        order_id = self.line_ids.mapped("sale_line_ids.order_id")
+        if len(order_id) > 1:
+            logger.warning("UBL Invoice for multiple sale order")
+            order_id = order_id[0]
+        order_ref = etree.SubElement(parent_node, ns["cac"] + "OrderReference")
+        order_ref_orderid = etree.SubElement(order_ref, ns["cbc"] + "SalesOrderID")
+        order_ref_orderid.text = order_id.name
+        customer_ref = order_id.client_order_ref
+        if customer_ref:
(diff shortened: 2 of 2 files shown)


TDu:
Help on understanding how this could be implemented welcome. Thank you (The UBL specification is quiet a beast!)

account_invoice_ubl/__manifest__.py
```diff
@@ -15,6 +15,7 @@
         "account_payment_partner",
         "base_ubl_payment",
         "account_tax_unece",
+        "sale",
```
astirpe, line 18: Instead of adding a new dependency here, you could wrap your modifications in a block like:

```python
if self.env['ir.module.module'].search([('name', '=', 'sale'), ('state', '=', 'installed')]):
    ...
else:
    ...
```


and in the "else" just keep the old code to avoid change of behavior.

astirpe (changes requested):

account_invoice_ubl/models/account_move.py
```diff
@@ -39,10 +39,17 @@ def _ubl_add_header(self, parent_node, ns, version="2.1"):
 
     def _ubl_add_order_reference(self, parent_node, ns, version="2.1"):
         self.ensure_one()
-        if self.name:
-            order_ref = etree.SubElement(parent_node, ns["cac"] + "OrderReference")
+        order_id = self.line_ids.mapped("sale_line_ids.order_id")
+        if len(order_id) > 1:
+            logger.warning("UBL Invoice for multiple sale order")
+            order_id = order_id[0]
```
astirpe, line 45: Instead of getting one order only, you could get all the orders and merge all the `client_order_ref`'s together

astirpe: Anyway this is just a technical suggestion, not sure if it's compliant with UBL functional specifications :see_no_evil:

account_invoice_ubl/models/account_move.py
```diff
@@ -39,10 +39,17 @@ def _ubl_add_header(self, parent_node, ns, version="2.1"):
...
+        order_ref = etree.SubElement(parent_node, ns["cac"] + "OrderReference")
+        order_ref_orderid = etree.SubElement(order_ref, ns["cbc"] + "SalesOrderID")
+        order_ref_orderid.text = order_id.name
```
astirpe: This line is failing in tests, could you check?

TDu:
@astirpe  Thank you for your review, I fixed the line for Travis, you have a point on the dependencies and I could fix that.
But as long as I can not generate a valid XML file with the customer reference and the sale order id, this PR as no purpose.

TDu:
Now Travis mention the problem I explain in the PR description

astirpe:
> Now Travis mention the problem I explain in the PR description

Yes I see. Thanks!

TDu:
I am closing this, because it is not being used so will not be updated.

PR #229, opened 2020-08-14
8.0 IMP base_ubl: add GLN support

bealdav:


base_business_document_import/models/business_document_import.py
```diff
@@ -103,6 +104,19 @@ def _match_partner(
             partner_dict, chatter_msg, domain, partner_type_label)
         if partner:
             return partner
+        if partner_dict.get("gln"):
+            if "partner_identification_gln" in self.env.registry._init_modules:
+                # Only if partner_identification_gln is installed
+                # this code is executed
+                categ = self.env.ref(
```
astirpe: You can do this with `raise_if_not_found=False` and avoid the check  whether `partner_identification_gln` is installed. If `partner_identification_gln.partner_identification_gln_number_category` is found do the search otherwise skip.

base_business_document_import/models/business_document_import.py
```diff
@@ -103,6 +104,19 @@ def _match_partner(
...
+                    "partner_identification_gln."
+                    "partner_identification_gln_number_category")
+                id_number = self.env["res.partner.id_number"].search([
+                    ("category_id", "=", categ.id),
+                    ("name", "=", partner_dict["gln"])])
+                # GLN unicity is managed by gln module
+                if id_number and id_number[0].name == partner_dict["gln"]:
```
astirpe: You can do the search with `limit=1` and avoid the `id_number[0]`.

base_business_document_import/models/business_document_import.py
```diff
@@ -103,6 +104,19 @@ def _match_partner(
             partner_dict, chatter_msg, domain, partner_type_label)
         if partner:
             return partner
+        if partner_dict.get("gln"):
+            if "partner_identification_gln" in self.env.registry._init_modules:
```
astirpe: I would put the code in a separate method, to be easily extended in case the gln field is provided by a different module, other than `partner_identification_gln`. IMO most of the customers use the `gln` by just adding a field on partner.

So this block of code would be something like:

```python
        if partner_dict.get("gln"):
            partner = _match_partner_by_gln()
            if partner:
                return partner
```

astirpe (changes requested):

base_business_document_import/models/business_document_import.py
bealdav: Thanks for your valuable review. I've have updated
if not installed, search() below will fail because of missing fields

astirpe (approved):
LGTM

hparfr (approved):
LGTM but didn't test.

bealdav:
@hparfr you're right but Odoo doesn't manage really well not mandatory dependencies in tests, at least in older versions. Thanks for your review

base_business_document_import/models/business_document_import.py
```diff
@@ -189,6 +194,23 @@ def _hook_match_partner(
...
+        categ = self.env.ref(
+            "partner_identification_gln."
+            "partner_identification_gln_number_category", raise_if_not_found=False)
+        if categ:
+            # partner_identification_gln is installed
+            id_number = self.env["res.partner.id_number"].search([
+                ("category_id", "=", categ.id),
+                ("name", "=", gln_number_from_file)], limit=1)
+            # GLN unicity is managed by gln module
+            if id_number and id_number.name == gln_number_from_file:
```
astirpe: I think here `id_number.name == gln_number_from_file` is not needed anymore:

```suggestion
            if id_number:
```

bealdav: Oops you're right

apollet (approved):

astirpe:
But you didn't squash... :smile:

bealdav:
ocabot has a squash feature ? I missed it

astirpe:
@bealdav squash should be done manually, before merging.. AFAIK

bealdav:
I love github suggestion feature and it's an old version, it doesn't really matter

PR #230, opened 2020-08-14
[FIX] test if res_ids is int, in this case change it in array

vrenaville:
* Purpose
If we try to validate a bank statement, it send an res_id as int, so length test of array failed.

* dev
test if res_id is an int, in this case transform it into an array


astirpe (approved):
Thanks!
Although it seems to me a bug of the statement import of Odoo, as `res_ids` is supposed to be always a list. :smile:

sebalix (approved):

vrenaville:
@astirpe I agree with you, but I prefer to be conservative because it's called many times in odoo core

guewen (approved):

PR #231, opened 2020-08-17
[13.0][Fix] account_invoice_ubl print error

TDu:
Hide the Generate UBL XML file button when the move is not posted as it
can not be generated in that state.

Also fix some crash when generating pdf for non supported invoice type
and state.

### DIFF
--- a/account_invoice_ubl/models/account_move.py
+++ b/account_invoice_ubl/models/account_move.py
@@ -325,7 +325,7 @@ class AccountMove(models.Model):
 
     def add_xml_in_pdf_buffer(self, buffer):
         self.ensure_one()
-        if self.is_ubl_sale_invoice_posted:
+        if self.is_ubl_sale_invoice_posted():
             version = self.get_ubl_version()
             xml_filename = self.get_ubl_filename(version=version)
             xml_string = self.generate_ubl_xml_string(version=version)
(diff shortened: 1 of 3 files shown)


astirpe:
Ok for hiding the button.
But for removing the asserts, I have some doubts. In what case do you get a crash? Maybe the usability should be improved somewhere else...

TDu:
@astripe Just printing an invoice in a draft state will raise the exception

astirpe:
Did you try by changing the `_post_pdf()` method?
Something like this:

```python

    def _post_pdf(self, save_in_attachment, pdf_content=None, res_ids=None):
        ....
            if self.is_ubl_xml_to_embed_in_invoice():
                invoice = self.env["account.move"].browse(res_ids)
                if invoice.state == "posted":
                    pdf_content = invoice.embed_ubl_xml_in_pdf(pdf_content)
        ....
```

TDu:
I looked at `is_ubl_xml_to_embed_in_invoice` changing in the same file but was not happy about it because it deals with a recordset.
I will try your option...

astirpe:
And also something like:

```python
    def postprocess_pdf_report(self, record, buffer):
        if self.is_ubl_xml_to_embed_in_invoice():
            if record._name != "account.move" or record.state == "posted":
                buffer = record.add_xml_in_pdf_buffer(buffer)
        return super().postprocess_pdf_report(record, buffer)
```

TDu:
@astirpe Thanks for your help.
Finally there was a bug on a function call.
And there is an existing function to test if the XML should be generated or not for an invoice.

astirpe:
Please fix also missing brackets in line: https://github.com/OCA/edi/blob/44bd063628cce8422fb643b16759675da5afb5d4/account_invoice_ubl/models/account_move.py#L337

TDu:
Good catch, fixed that line 337 as well.

astirpe (approved):

guewen (approved):

PR #234, opened 2020-09-01
[10.0] account_import_invoice - Add 2 hooks to get invoice content and partner

yvaucher:
This allow to give an other mean of import of invoice content than from a file.
Which is useful for QR scan in case of l10n_ch_qr_bill_scan module.

Plus add a hook when getting the partner that can accept an action.
This action can be a new wizard step to define the partner.
Also a hook needed for l10n_ch_qr_bill_scan.

https://github.com/OCA/l10n-switzerland/pull/561

### DIFF
--- a/account_invoice_import/wizard/account_invoice_import.py
+++ b/account_invoice_import/wizard/account_invoice_import.py
@@ -488,6 +488,40 @@ class AccountInvoiceImport(models.TransientModel):
             ], limit=1)
         return existing_inv
 
+    def _get_parsed_invoice(self):
+        """Hook to change the method of retrieval for the invoice data"""
+        return self.parse_invoice(self.invoice_file, self.invoice_filename)
+
+    def _hook_no_partner_found(self, partner_dict):
+        """Hook designed to add an action when no partner is found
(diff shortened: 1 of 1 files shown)


account_invoice_import/wizard/account_invoice_import.py
```diff
@@ -488,6 +488,38 @@ def invoice_already_exists(self, commercial_partner, parsed_inv):
             ], limit=1)
         return existing_inv
 
+    def _get_parsed_invoice(self):
+        """Hook to change the method of retrieval for the invoice data"""
+        return self.parse_invoice(self.invoice_file, self.invoice_filename)
+
+    def _hook_no_partner_found(self, partner_dict):
+        """Hook designed to add an action when no partner is found
+        For instance to propose to create the partner based on the partner_dict.
```
gurneyalex, line 497:
```suggestion
        For instance to propose to create the partner based on the partner_dict.
        In that case, the hook is expected to return an action serialized as a dictionary 
        which will be returned to the web client. 
```

gurneyalex (approved):
small docstring improvement suggested.

i-vyshnevska (approved):

PR #236, opened 2020-09-18
[13.0] UBL Party parser refactor proposal

jbaudoux:
**Party parser refactor**
1/ First try to match on explicit partner ref
2/ If country and/or state is provided, make optional on search domain
3/ If vat is provided, make required on search domain
4/ Call search hook
5/ Try to match on email
6/ Try to match on name
7/ Try to match on website or email domain
8/ Try to match on vat

**Do not reformat VAT number** as in Odoo v13 the VAT number is normalized
and UBL provides a normalized VAT number. Moreover, some countries like
Switzerland have spaces in the normalized VAT number that should not be
removed.

**Do not require PartyName** that is optional in UBL

**Add partner_indentification glue module** with https://github.com/OCA/partner-contact/blob/13.0/partner_identification and match partner on cac:PartyIdentification/cbc:ID

**Refactor delivery partner matching**

**Add sales order invoicing partner parsing & matching**

**Add validation of UBL XML content according to schema**

**Suggest to update existing quotation only if the client ref (PO number) is the same**

cc @alexis-via @bealdav @gixxi @lmignon


bealdav:
To complete the subjetc, gln identification is used in some real industry cases https://github.com/OCA/edi/commit/1d19553ec96018d8787e11d94fe28c9fc65c500a

jbaudoux:
> To complete the subjetc, gln identification is used in some real industry cases [1d19553](https://github.com/OCA/edi/commit/1d19553ec96018d8787e11d94fe28c9fc65c500a)

The policy as evolved and now only those schemeID codes should be used:
https://docs.peppol.eu/poacc/upgrade-3/codelist/ICD/
Same with the VAT number that could previously be provided in the same way using the a scheme "<countrycode>:VAT"

See https://github.com/OpenPEPPOL/documentation/blob/master/TransportInfrastructure/PEPPOL-EDN-Policy-for-use-of-identifiers-4.0-2019-01-28.pdf

I will add the PartyIdentification/ID in the partner_dict, then a hook could process it

base_ubl/models/ubl.py
```diff
@@ -633,11 +633,15 @@ def ubl_parse_party(self, party_node, ns):
...
             "vat": vat_xpath and vat_xpath[0].text or False,
-            "name": partner_name_xpath[0].text,
+            "name": partner_name_xpath and partner_name_xpath[0].text or False,
             "email": email_xpath and email_xpath[0].text or False,
             "website": website_xpath and website_xpath[0].text or False,
             "phone": phone_xpath and phone_xpath[0].text or False,
         }
+        id_node.xpath("cac:PartyIdentification/cbc:ID", namespaces=ns)
+        if id_node:
+            partner_dict["id"] = id_node[0].text
```
astirpe: As far as I understand this `"id"` should be the identification document number? Could `"id"` here be confused with the id of the record? In this case better to use another string.

jbaudoux: renamed to id_number as in https://github.com/OCA/partner-contact/blob/13.0/partner_identification/models/res_partner_id_number.py

jbaudoux:
@bealdav Following your remark regarding GLN, I have added a glue module partner_identification_import that should solve it

jgrandguillaume:
Great @jbaudoux Thanks ! I'm in-line with the proposed change. Please fix the test now ;)

base_business_document_import/models/business_document_import.py
```diff
@@ -109,31 +119,20 @@ def _match_partner(self, partner_dict, chatter_msg, partner_type="supplier"):
             )
             if state:
                 domain += ["|", ("state_id", "=", False), ("state_id", "=", state.id)]
+
+        # Append required vat to the domain
         if partner_dict.get("vat"):
-            vat = partner_dict["vat"].replace(" ", "").upper()
```
lmignon, line 113: @jbaudoux Why do you remove the search with 'vat' only as criteria?

base_business_document_import/models/business_document_import.py
```diff
@@ -177,20 +187,21 @@ def _match_partner(self, partner_dict, chatter_msg, partner_type="supplier"):
...
-            if partner:
-                return partner
-        if partner_dict.get("name"):
-            partner = rpo.search(
-                domain + [("name", "=ilike", partner_dict["name"])],
-                limit=1,
-                order=order,
-            )
+
+        if partner_dict.get("vat"):
```
lmignon, line 226: @jbaudoux You already make a check on vat a line #123. Why do you move the search on vat here with a domain that could contain others criteria?

partner_identification_import/models/business_document_import.py
```diff
@@ -0,0 +1,31 @@
...
+                # partner_identification_gln is installed
+                id_number = self.env["res.partner.id_number"].search(
+                    [
+                        ("category_id", "=", categ.id),
+                        ("name", "=", partner_dict["id_number"]),
+                    ],
+                    limit=1,
+                )
+                if id_number:
+                    return id_number.partner_id
```
jbaudoux, line 45: If the id number is explicit, we should not fallback on standard matching and raise an error

base_business_document_import/models/business_document_import.py
jbaudoux: @lmignon I'd like to support complex partner structures. The customer could be a contact child of the company. So I removed first match on commercial entity only and replaced by adding the vat in the domain to restrict to the right partner structure on the following searches.

base_business_document_import/models/business_document_import.py
jbaudoux: @lmignon If I cannot match the right partner inside the commercial entity structure, then I end up matching the one with the highest customer/supplier rank

base_business_document_import/models/business_document_import.py
```diff
@@ -69,7 +107,8 @@ def _match_partner(self, partner_dict, chatter_msg, partner_type="supplier"):
         if partner_dict.get("id"):
             return rpo.browse(partner_dict["id"])
         company_id = self._context.get("force_company") or self.env.user.company_id.id
-        domain = ["|", ("company_id", "=", False), ("company_id", "=", company_id)]
+        domain = domain or []
```
guewen, line 110: `domain` should be copied before mutating, to prevent side effects on the domain uptream

base_business_document_import/models/business_document_import.py
```diff
@@ -46,8 +46,48 @@ def _strip_cleanup_dict(self, match_dict):
...
+                + [
+                    "|",
+                    ("mobile", "=", partner_dict["phone"]),
+                    ("phone", "=", partner_dict["phone"]),
+                ],
+                limit=1,
+                order=order,
+            )
+            if partner:
+                return partner
```
guewen, line 80: not blocking: worth splitting IMO

base_business_document_import/models/business_document_import.py
```diff
@@ -222,97 +261,116 @@ def _hook_match_partner(
         return False
 
     @api.model
-    def _match_shipping_partner(self, shipping_dict, partner, chatter_msg):
+    def _match_shipping_partner(
+        self, partner_dict, partner, chatter_msg, domain=None, raise_exception=True
```
TDu, line 265:
```suggestion
        self, partner_dict, partner, chatter_msg, domain=None
```
Seems not to be used anywhere. And can probably be handled by a `try...except` in function using `super` (if this was added in anticipation)

base_business_document_import/models/business_document_import.py
jbaudoux: In fact, I'm relying on side effect as of here: https://github.com/OCA/edi/pull/236/files#diff-3bde8c0adf791caf40e4df6ec01a1e45fddd4298a4810398f4d559162960b95aR340
Any better way of doing this?

jgrandguillaume (approved):
LGTM, functional test ok

lmignon (approved):
LGTM (Code review only)

base_business_document_import/models/business_document_import.py
```diff
@@ -59,8 +99,6 @@ def _match_partner(self, partner_dict, chatter_msg, partner_type="supplier"):
             'ref': 'C1242',
             'phone': '01.41.98.12.42',
             }
-        The key 'phone' is used by the module
-        base_phone_business_document_import
         """
```
alexis-via, line 102: I don't comment my code a lot, but if you remove the few comments I write...
I find this comment useful... don't you ?

alexis-via (approved):

base_business_document_import/models/business_document_import.py
jbaudoux: @alexis-via I couldn't find that module you were referring and I integrated the phone parsing in this module. As such, I found out that this comment was not relevant anymore.

PR #243, opened 2020-10-19
[13.0][ADD] EDI

etobella:
This module tries to create a standard platform similar to `account_edi` that can be inherited by all modules.

As a demo, `mail` option is created

@pedrobaeza @JordiBForgeFlow @jarroyomorales


edi/models/edi_document.py
```diff
@@ -0,0 +1,102 @@
...
+        [
+            ("outbound", "Outbound"),  # Documents sent
+            ("inbound", "Inbound"),  # Documents received"
+        ],
+        required=True,
+        default="outbound",
+        readonly=True,
+        help="Show direction of the document",
+    )
+    parent_id = fields.Many2one(
```
bealdav, line 47: Thanks for this contribution @etobella. It seems really interesting.
I've have just noticed there is here a feature already implemented in a low level module in v12
https://github.com/OCA/server-tools/tree/12.0/attachment_queue
Could you have a look. Both module share some fields and features: example:
fields: edi_type, m2o attachment_id, state, error
See
https://github.com/OCA/server-tools/blob/12.0/attachment_queue/models/attachment_queue.py

I think there is an opportunity to reuse attachment_queue by inheriting it and complete with your own fields and edi features. What do you think ?

etobella: Well, they are really similar, but the behavior at the end is different.
`edi.documents` are a relation between an attachment (m2o and m2m [Annexes]) and a record.
`attachment.queue` is just a task created over an attachment.

It might be possible to integrate them, but I am not sure if it would be better to integrate `attachment_queue` directly to `edi` and not making this kind of split. What do you think? @pedrobaeza ?

bealdav: FYI i use UBL edi in version 8.0 and in next month in 14.0. I have the same need than you for transportation purpose. I hardly use the feature with this an older module https://github.com/akretion/ak-odoo-incubator/tree/8.0/attachment_base_synchronize.
The idea is that we do not only need file transportation in EDI (this repo). We also need for other data exchange.

etobella: EDI is Electronic data interchange, not a specific protocol, so, If you need to do some kind of transportation between systems, at the end, it can be represented as EDI

bealdav: > attachment.queue is just a task created over an attachment.

You're right and that is a feature we need in many domains. But in some domains we also need extra features as you propose (annexes). It seems there is no real difficulty to support annexes on top of attachment_queue.

Probably it could be really confusing to have too many ways to deal with file exchange I think.

Note: attachment_queue is used by attachment_synchronize in the same repo and be combined with https://github.com/OCA/storage/tree/12.0/storage_backend_sftp or s3 like storage.

> If you need to do some kind of transportation between systems, at the end, it can be represented as EDI

Yes at the end a rename in v13 of an existing module in v12 to stick to edi can be done if you want.

pedrobaeza: I think the purpose of `attachment_queue` is another and can't be used as base for EDI, as not everything is exchanged as "attachments". There are SOAP messages (like in Spanish SII), there are webservices requests like in FacturaE, etc, so a better abstract base is needed.

bealdav: > so a better abstract base is needed.

You're probably right, thanks a lot.

etobella:
Deprecating in favor of #248

PR #244, opened 2020-10-21
[10.0][IMP] purchase_order_import: Takes into account that a purchase order…

lmignon:
… line can be linked to more thant one stock.move when processing the amended qties from the order response document

### DIFF
--- a/purchase_order_import/wizard/order_response_import.py
+++ b/purchase_order_import/wizard/order_response_import.py
@@ -267,9 +267,6 @@ class OrderResponseImport(models.TransientModel):
 
     @api.model
     def _process_conditional(self, purchase_order, parsed_order_document):
-        precision = self.env["decimal.precision"].precision_get(
-            "Product Unit of Measure"
-        )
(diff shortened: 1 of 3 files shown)


lmignon:
@bealdav @jgrandguillaume Since you reviewed the original PR your could be interested by this one.

bealdav:
I'm interested but no time up to now sorry. I've to backport it in v8 in next month

purchase_order_import/wizard/order_response_import.py
```diff
@@ -297,100 +294,132 @@ def _process_conditional(self, purchase_order, parsed_order_document):
...
-        Add the move the picking's backorder
+        Add the moves to the picking's backorder
         return the backorder associated to the current picking. If no backorder
         exists, create a new one.
-        :param move:
+        :param moves:
         """
         StockPicking = self.env["stock.picking"]
-        current_picking = move.picking_id
+        current_picking = moves[0].picking_id
```
PieterPaulussen, line 412: Perhaps it is safer to group the moves by picking and iterate over the combinations?

PR #247, opened 2020-11-19
[13.0] Add base_edi

simahawk:
See DESCRIPTION.rst for details.


bealdav (approved):

pedrobaeza:
There's already a proposal for a good EDI framework base done on #243. You should check it instead of adding another one.

simahawk:
> You should check it instead of adding another one.

@pedrobaeza there's no framework here. This is just a menu and base group to aggregate things.

In any case I'm working on something similar but I missed it as when I checked it was not there and I started working on it already but put it on hold.

Nevertheless this here can be use as a common ground.
I'll ping @etobella to check together that implementation.
Thanks for the ref.

etobella (approved):

pedrobaeza:
Enric has informed me about your joined plans, Simone, so great to find a common ground :smiley:

JordiBForgeFlow (approved):

bealdav:
I'm in favor to have informations on future developments on this topic.

Maybe an RFC issue could be nice, thanks ;-)

simahawk:
@bealdav you are right. I quickly drafted this https://github.com/OCA/edi/issues/249. Maybe not complete, but it's a start :wink:

PR #250, opened 2020-11-23
[IMP] edi: Define Records and type. Define function for creation of edi exchange records

etobella:
@simahawk 

First of the PRs, this is necessary before any of the others :wink:


simahawk:
@etobella ehm... https://github.com/OCA/edi/pull/251 :sweat_smile: 
I've already rebased your fork and added some fixes/imp here https://github.com/OCA/edi/pull/251
Let's get in touch to not overlap anymore :wink:

PR #248, opened 2020-11-19
[13.0] Add edi

simahawk:
Depends on #247 
To be considered for #243.
To be the base for more work to be discussed w/ @etobella

PR #251, opened 2020-11-23
[IMP] edi: define type, record, record create helper 

simahawk:
@etobella 1st round after #248


edi/models/edi_backend.py
```diff
@@ -37,3 +47,83 @@ def _get_component(self, safe=False, work_ctx=None, **kw):
...
+            )
+        if exchange_record.exchange_file:
+            raise exceptions.UserError(
+                _("Exchabge record ID=%d already has a file to process!")
+                % exchange_record.id
+            )
+
+    def _generate_output(self, exchange_record, **kw):
+        """To be implemented"""
+        raise NotImplementedError()
```
etobella, line 129: Maybe it would be good to propose something here.
I have three options on mind:

* Have the backend/type a field that marks a component of usage "generate"
* Have the backend/type a field that marks a function
* Have a selector that allows to define which options of below is used. What do you think?

simahawk: we could look up for a component for the current record model with a usage like `edi.output.{backend_type}.{exchange_type}` and fallback to `edi.output.{backend_type}` if not found.

Anyway, in the context of this PR is not required IMO.

etobella (approved):

edi/models/edi_backend.py
etobella: Leaving to round 2 then :wink:

jarroyomorales (approved):
Code review LGTM :+1: !

simahawk:
Version bumped manually to work around Travis issue.

edi/models/edi_exchange_record.py
```diff
@@ -0,0 +1,133 @@
...
+        help="Sent or received on this date.",
+        compute="_compute_exchanged_on",
+        store=True,
+    )
+    # TODO: use sequence and make it unique
+    exchange_identification_code = fields.Char(
+        track_visibility="onchange",
+        help="Identification of the EDI, useful to search and join other documents",
+    )
+    ack_file = fields.Binary(attachment=True)
```
LoisRForgeFlow, line 53: @simahawk just a question here, wouldn't this be more suitable in a new exchange record type "ack"?

PR #253, opened 2020-11-23
[IMP] edi: add generic send function

etobella:
This PR defines the sending function of exchange records.

Still pending how to define the sending function.
@simahawk @jarroyomorales

### DIFF
--- a/edi/models/edi_backend.py
+++ b/edi/models/edi_backend.py
@@ -127,3 +127,68 @@ class EDIBackend(models.Model):
     def _generate_output(self, exchange_record, **kw):
         """To be implemented"""
         raise NotImplementedError()
+
+    def exchange_send(self, exchange_record):
+        """Send exchange file."""
+        self.ensure_one()
+        exchange_record.ensure_one()
+        # In case already sent: skip sending and check the state
+        check = self._exchange_output_check(exchange_record)
+        if not check:
+            return False
+        try:
+            self._exchange_send(exchange_record)
+        except Exception as err:
+            if self.env.context.get("_edi_send_break_on_error"):
+                raise
+            error = str(err)
+            state = "output_error_on_send"
+            message = exchange_record._exchange_send_error_msg()
+            res = False
+        else:
+            message = exchange_record._exchange_sent_msg()
+            error = None
+            state = "output_sent"
+            res = True
+        finally:
+            exchange_record.edi_exchange_state = state
+            exchange_record.exchange_error = error
+            if message:
+                self._exchange_notify_record(exchange_record, message)
+        return res
+
+    def _exchange_output_check(self, exchange_record):
+        if exchange_record.direction != "output":
+            raise exceptions.UserError(
+                _("Record ID=%d is not meant to be sent!") % exchange_record.id
+            )
+        if not exchange_record.exchange_file:
+            raise exceptions.UserError(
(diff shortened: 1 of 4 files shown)


edi/models/edi_backend.py
```diff
@@ -127,3 +127,62 @@ def _validate_generate_output(self, exchange_record, force=False):
     def _generate_output(self, exchange_record, **kw):
         """To be implemented"""
         raise NotImplementedError()
+
+    def exchange_send(self, exchange_record):
+        """Send exchange file."""
+        self.ensure_one()
+        exchange_record.ensure_one()
+        if exchange_record.direction != "output":
```
simahawk: this part can be moved to `_exchange_output_check`

edi/models/edi_backend.py
```diff
@@ -127,3 +127,62 @@ def _validate_generate_output(self, exchange_record, force=False):
...
+            raise exceptions.UserError(
+                _("Record ID=%d has no file to send!") % exchange_record.id
+            )
+        # In case already sent: skip sending and check the state
+        check = self._exchange_output_check(exchange_record)
+        if not check:
+            return False
+        try:
+            self._exchange_send(exchange_record)
+        except Exception as err:
```
simahawk, line 141: this catchall exception can be dangerous as it can swallow important exceptions.
I would rely on a ctx key to raise the error on demand (eg: handy for tests or for manual test via UI).
Something like:
```suggestion
        except Exception as err:
            if self.env.context.get("_edi_send_break_on_error"):
                raise
```

edi/models/edi_backend.py
```diff
@@ -127,3 +127,62 @@ def _validate_generate_output(self, exchange_record, force=False):
...
+            exchange_record.exchange_error = error
+            if message:
+                self._exchange_notify_record(exchange_record, message)
+        return res
+
+    def _exchange_output_check(self, record):
+        return record.edi_exchange_state in ["output_pending", "output_error_on_send"]
+
+    def _exchange_send(self, exchange_record):
+        component = self._get_component(usage="edi.send.%s" % self.backend_type_id.code)
```
simahawk, line 176:
```suggestion
        # TODO: maybe lookup for an `exchange_record.model` specific component 1st
        component = self._get_component(usage="edi.send.%s" % self.backend_type_id.code)
```

edi/models/edi_backend.py
```diff
@@ -127,3 +127,62 @@ def _validate_generate_output(self, exchange_record, force=False):
...
         """To be implemented"""
         raise NotImplementedError()
+
+    def exchange_send(self, exchange_record):
+        """Send exchange file."""
+        self.ensure_one()
+        exchange_record.ensure_one()
+        if exchange_record.direction != "output":
+            raise exceptions.UserError(
+                _("Record ID=%d is not meant to be sended") % exchange_record.id
```
simahawk:
```suggestion
                _("Record ID=%d is not meant to be sent") % exchange_record.id
```

simahawk (approved):
LG, just some small remarks

simahawk:
@etobella @jarroyomorales this one could be merged IMO.

jarroyomorales (approved):

simahawk:
@etobella can you bump the version manually and merge after?

etobella:
@simahawk I cannot proceed with merge, as I don't have permissions, but I made the bump manually

PR #255, opened 2020-11-23
[14.0][MIG] account_einvoice_generate

alexis-via:
Module has been renamed to remove dash in module name.


account_einvoice_generate/__manifest__.py
```diff
@@ -0,0 +1,18 @@
...
+    "name": "Account e-invoice Generate",
+    "version": "14.0.1.0.0",
+    "category": "Accounting & Finance",
+    "license": "AGPL-3",
+    "summary": "Technical module to generate PDF invoices with " "embedded XML file",
+    "author": "Akretion,Onestein,Odoo Community Association (OCA)",
+    "maintainers": ["alexis-via"],
+    "website": "https://github.com/OCA/edi",
+    "depends": ["account"],
+    "excludes": ["account_edi"],
```
etobella: Why do you need this exclusion?

alexis-via: I want to make things clear that you have to make a choice: either you opt for the einvoicing stack of OCA, or for the einvoicing stack of Odoo SA. You can't use both at the same time.
From a technical point of view, the exclusion is one level higher : account_invoice_factux excludes account_edi_facturx. But I prefer to make things clear at the generic level, so that there is no misunderstanding.

alexis-via:
All unece modules in OCA/community-data-files have been merged, so we can soon merge this one.

StefanRijnhart:
I'd like to see Travis (or github actions) run on this, so a rebase would be good.

fshah-initos:
Hi @alexis-via 
Is it possible to fix the failed checks so that we can get this PR merged soon. As it is depened on other modules too.

StefanRijnhart:
Thanks for the rebase! Tests fail on `odoo.exceptions.UserError: Modules "Account e-invoice Generate" and "Import/Export Invoices From XML/PDF" are incompatible.` and I now see that account_edi, which you exclude in this module, is autoinstalled.

alexis-via:
Yep. I don't know how to solve this. It's the same pb on facturx: we declare a conflict on the facturx module from the official addons, but that module is auto_install = True. Any idea how to solve this?

StefanRijnhart:
I just had a look (again) how to tackle this problem but I don't think there is a way to do this in a module except when it is preloaded when the database is created and keeps on being preloaded throughout the lifecycle of the database.
I also compared the code of the Odoo edi/facturx modules and yours and I don't immediately see the technical incompatibility (although I'm sure that it must be confusing for the user to see both options available). Can you clarify those technical incompatibilities to me?

dsolanki-initos:
Hi @alexis-via Can you please resolve the Failed checks so that we merge the module asap?

dsolanki-initos:
Hi @alexis-via Can you please resolve the Failed checks so that we merge the module asap?

dsolanki-initos:
It would be great if the checks are resolved so that we can merge the PR soon.

StefanRijnhart:
@dsolanki-initos fixing the checks is not trivial. There is a supposed incompatibility with a module in Odoo core that is autoinstalled. @alexis-via do you have a slotion in mind?

mkxmawilix:
Hello there,

I've no tested this trick, and this is just an idea and I'm not sure this is possible but, can we bring with a "pre_init_hook" with "account_einvoice_generate" and this hook will insert a module dependency with a SQL request for the dependencies of the"account_edi" module. This dependency would be an un-installable ghost module that we will never install and this hook would uninstall "account_edi" at the same time.

In this case the "auto_install=True" of "account_edi" will never be triggered as long we will not install this "un-installable ghost module".

As I said this is an idea, I really don't know if this is technically doable, I just wanna help you to solve this.

By the way thanks for all the job you are doing here.

thomaspaulb:
@StefanRijnhart

Assuming for a moment the incompatibility between the Odoo module and the OCA module is indeed real (i have not checked).

Then am I correct to say that this is just a problem of the OCA test suite? Because in real life, people would just install the OCA module, see the incompatibility warning, deinstall the Odoo module and install the OCA module. After that, due to the conflict, the Odoo module is not (auto)installed anymore.

(If you dont know, i can test it myself, but I thought you may already know the answer)

fshah-initos:
Hi @alexis-via
Is it possible to fix the failed checks so that we can get this PR merged soon. As it is depended on other modules too.

alexis-via:
Travis is green.
I eventually removed the "excludes" key on "account_edi". It was not a pure technical incompatibility, it was a way to show the user that he had to make a choice between the OCA edi stack and the Odoo edi stack. But, as account_edi is auto_install=True, it makes things difficult for tests. So I dropped the "excludes" key.

alexis-via:
@fshah-initos If you make an approve review here, I'll merge.

alexis-via:
@dsolanki-initos If you make an approve review here, I'll merge.

StefanRijnhart (approved):

PR #261, opened 2020-11-24
[FIX] travis

etobella:
Disables execution on travis of modules that are not working properly due to PR #208 

This is obviously a patch for other modules on the repo to work properly. It is obviously needed to solve the original issue

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,6 +42,7 @@ install:
     ${HOME}/maintainer-quality-tools
   - export PATH=${HOME}/maintainer-quality-tools/travis:${PATH}
   - travis_install_nightly
+  - rm -fr ${HOME}/{odoo,OCB}-${VERSION}/addons/account_facturx
 
 script:
   - travis_run_tests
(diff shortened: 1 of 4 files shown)


bealdav:
Hi @etobella maybe you might be inspired with that

https://github.com/OCA/l10n-france/blob/12.0/.travis.yml#L32

to exclude only the module in conflict instead of exclude many modules

etobella:
This pr is just a patch. Also, i would have to remove all the modules on the exclude as all depend on the uninstallable module. Not removing one if them would rise an error. Also, this way codecov is saying that is untested.

At the end an exclude from a base autoinstalled module is not great IMO. And that is something that should be fixed by the author

etobella:
At the end, I followed your comments. When this is green, it would be interesting to merge it soon

bealdav:
Yeah, you're pragmatic @etobella , really nice

bealdav (approved):

simahawk (approved):
let's patch it!

PR #262, opened 2020-11-25
[IMP] edi: Add mixins

etobella:
This PR adds a new mixin for exchange records. This should be inherited by records that might create exchange records

### DIFF
--- a/edi/models/edi_exchange_consumer_mixin.py
+++ b/edi/models/edi_exchange_consumer_mixin.py
@@ -1,0 +1,47 @@
+# Copyright 2020 ACSONE SA
+# Copyright 2020 Creu Blanca
+# @author Simone Orsi <simahawk@gmail.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+
+class EDIExchangeConsumerMixin(models.AbstractModel):
+    """Record that might have related EDI Exchange records
+    """
+
+    _name = "edi.exchange.consumer.mixin"
+    _description = "Abstract record where exchange records can be assigned"
+
+    exchange_record_ids = fields.One2many(
+        "edi.exchange.record",
+        inverse_name="res_id",
+        domain=lambda r: [("model", "=", r._name)],
+    )
+    exchange_record_count = fields.Integer(compute="_compute_exchange_record_count")
+
+    def _has_exchange_record(self, exchange_type, backend, extra_domain=False):
+        """This function is useful when generating the configuration"""
+        domain = [
+            ("model", "=", self._name),
+            ("res_id", "=", self.id),
+            ("backend_id", "=", backend.id),
+            ("type_id.code", "=", exchange_type),
+        ]
+        if extra_domain:
+            domain += extra_domain
+        return bool(self.env["edi.exchange.record"].search_count(domain))
+
+    @api.depends("exchange_record_ids")
+    def _compute_exchange_record_count(self):
+        for record in self:
+            record.exchange_record_count = len(record.exchange_record_ids)
+
+    def action_view_edi_records(self):
+        self.ensure_one()
+        action = self.env.ref("edi.act_open_edi_exchange_record_view").read()[0]
+        action["domain"] = [("model", "=", self._name), ("res_id", "=", self.id)]
+        return action
+
+    def _edi_generate_records(self):
+        raise NotImplementedError()
--- a/edi/tests/test_edi_mixin.py
+++ b/edi/tests/test_edi_mixin.py
@@ -1,0 +1,50 @@
+# Copyright 2020 Creu Blanca
+# @author: Enric Tobella
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo_test_helper import FakeModelLoader
+
+from .common import EDIBackendCommonTestCase
+
+
+class EDIBackendTestCase(EDIBackendCommonTestCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        # Load fake models ->/
+        cls.loader = FakeModelLoader(cls.env, cls.__module__)
+        cls.loader.backup_registry()
+        from .models import EdiExchangeConsumerTest
+
+        cls.loader.update_registry((EdiExchangeConsumerTest,))
+        cls.new_record = (
+            cls.env["edi.exchange.consumer.test"]
+            .with_context(test_backend_id=cls.backend.id)
+            .create({})
+        )
+        cls.exchange_type_out.exchange_filename_pattern = "{record.id}"
+
+    @classmethod
+    def tearDownClass(cls):
+        cls.loader.restore_registry()
+        super().tearDownClass()
+
+    def setUp(self):
+        super().setUp()
+
+    def test_mixin(self):
+        self.assertEqual(0, self.new_record.exchange_record_count)
+        vals = {
+            "model": self.new_record._name,
+            "res_id": self.new_record.id,
+        }
+        exchange_type = "test_csv_output"
+        exchange_record = self.backend.create_record(exchange_type, vals)
+        self.new_record.refresh()
+        self.assertEqual(1, self.new_record.exchange_record_count)
+        action = self.new_record.action_view_edi_records()
+        self.new_record.refresh()
+        self.assertEqual(
+            exchange_record, self.env["edi.exchange.record"].search(action["domain"])
+        )
+        self.new_record._has_exchange_record(exchange_type, self.backend)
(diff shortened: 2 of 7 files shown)


edi/models/edi_exchange_record_mixin.py
```diff
@@ -0,0 +1,90 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+
+class EDIExchangeRecordMixin(models.AbstractModel):
+    """Record that might have related EDI Exchange records
+    """
+
+    _name = "edi.exchange.record.mixin"
```
simahawk:
```suggestion
    _name = "edi.exchange.consumer.mixin"
```
I named it like this in the original code as it expresses better the purpose :wink:

edi/models/edi_exchange_record_mixin.py
```diff
@@ -0,0 +1,90 @@
...
+        "edi.exchange.record",
+        inverse_name="res_id",
+        domain=lambda r: [("model", "=", r._name)],
+    )
+    exchange_record_count = fields.Integer(compute="_compute_exchange_record_count")
+    missing_exchange_records = fields.Boolean(
+        compute="_compute_has_missing_exchange_records"
+    )
+
+    def _get_exchange_record_configuration(self):
```
simahawk: Is not clear what this method does/should return.
As per the schema of returned data I'd say it should always be a dict as it makes always clear what you get.
Let's talk about this maybe?

edi/models/edi_exchange_record_mixin.py
```diff
@@ -0,0 +1,90 @@
...
+            ("type_id.code", "=", exchange_type),
+        ]
+
+    def _has_exchange_record(self, exchange_type, backend):
+        return self.env["edi.exchange.record"].search(
+            self._has_exchange_record_domain(exchange_type, backend), limit=1
+        )
+
+    @api.depends("exchange_record_ids")
+    def _compute_has_missing_exchange_records(self):
```
simahawk: I don't get how you determine this reliably.

edi/models/edi_exchange_record_mixin.py
```diff
@@ -0,0 +1,90 @@
...
+        for record in self:
+            missing_exchange_records = False
+            for vals in record._get_exchange_record_configuration():
+                if not record._has_exchange_record(vals[0], vals[1]):
+                    missing_exchange_records = True
+                    break
+            record.missing_exchange_records = missing_exchange_records
+
+    def _generate_exchange_records(self):
```
simahawk: this should be used when/how? And how do you select the backend?

edi/tests/test_edi_mixin.py
```diff
@@ -0,0 +1,136 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+import mock
+
+from odoo import api, models
+
+from .common import EDIBackendCommonTestCase
+
+
+class EdiExchangeRecordMixinTest(models.Model):
```
simahawk: never define test models in a model that is imported at runtime as it gets there forever.
Please use odoo-test-helper (yeah, brand new), Example https://github.com/OCA/search-engine/blob/13.0/connector_search_engine/tests/test_all.py#L21

edi/tests/test_edi_mixin.py
```diff
@@ -0,0 +1,136 @@
...
+
+        self.registry.enter_test_mode(self.cr)
+        self.old_cursor = self.cr
+        self.cr = self.registry.cursor()
+        self.env = api.Environment(self.cr, self.uid, {})
+        self.test_model = self._init_test_model(EdiExchangeRecordMixinTest)
+        self.new_record = self.env["edi.exchange.record.mixin.test"].create({})
+        self.exchange_type_out.exchange_filename_pattern = "{record.id}"
+
+    def _init_test_model(self, model_cls):
```
simahawk: you can trash all this w/ model loader :wink:

simahawk (changes requested):

edi/tests/test_edi_mixin.py
etobella: Thanks!!! I was looking for this code but I was unable to find it :smile:

edi/models/edi_exchange_record_mixin.py
etobella: Done

edi/models/edi_exchange_record_mixin.py
etobella: I added more information there, I hope this is easier now.

edi/models/edi_exchange_record_mixin.py
etobella: Just checking the possible configurations :smile: Now it is clearer

edi/models/edi_exchange_record_mixin.py
etobella: The backend is defined on the configuration, with the new comments I hope it is clearer

edi/models/edi_exchange_consumer_mixin.py
```diff
@@ -0,0 +1,104 @@
...
+                }
+                creation_args.update(configuration.get("create_vals", {}))
+                exchange_record = backend.create_record(exchange_type, creation_args)
+                record._post_generate_exchange_records(exchange_record)
+                exchange_records |= exchange_record
+        return exchange_records
+
+    def _post_generate_exchange_records(self, exchange_record):
+        if (
+            self.env.context.get("force_edi_exchange_generation", False)
```
etobella: I am not sure if this context should be taken from the record or from the exchange_record (would be inherited from the backend of the configuration)

edi/tests/test_edi_mixin.py
```diff
@@ -0,0 +1,101 @@
...
+        cls.loader.update_registry((EdiExchangeConsumerTest,))
+
+    @classmethod
+    def tearDownClass(cls):
+        cls.loader.restore_registry()
+        super().tearDownClass()
+
+    def setUp(self):
+        super().setUp()
+        self.new_record = (
```
simahawk: this should stay in setUpClass. Never re-create records for all tests if not really required (which is almost never) :wink:

edi/tests/test_edi_mixin.py
```diff
@@ -0,0 +1,101 @@
...
+
+import base64
+
+import mock
+from odoo_test_helper import FakeModelLoader
+
+from .common import EDIBackendCommonTestCase
+
+
+class EDIBackendTestCase(EDIBackendCommonTestCase, FakeModelLoader):
```
simahawk: You don't have to inherit FakeModelLoader

edi/models/edi_exchange_consumer_mixin.py
```diff
@@ -0,0 +1,104 @@
...
+            record.missing_exchange_records = bool(
+                record._get_exchange_record_configuration()
+            )
+
+    def _generate_exchange_records(self):
+        """This function will generate the exchange records of consumers.
+
+        For example, this can be added when posting an invoice or validating a picking
+        """
+        exchange_records = self.env["edi.exchange.record"]
```
simahawk: @etobella as we discussed, let's leave here only the hook `_generate_exchange_records`.

simahawk (changes requested):

edi/models/edi_exchange_consumer_mixin.py
etobella: I think, maintaining it has not much sense, I will keep the `_has_exchange_record`

simahawk: It has sense as we still want to support a button to generate it manually.
I'd have `edi_generate_records` (public, can be used by a btn) which calls `_edi_generate_records` which by default here raises a `NotImplementedError`.

edi/models/edi_exchange_consumer_mixin.py
```diff
@@ -0,0 +1,47 @@
...
+        """This function is useful when generating the configuration"""
+        domain = [
+            ("model", "=", self._name),
+            ("res_id", "=", self.id),
+            ("backend_id", "=", backend.id),
+            ("type_id.code", "=", exchange_type),
+        ]
+        if extra_domain:
+            domain += extra_domain
+        return self.env["edi.exchange.record"].search(domain, limit=1)
```
jarroyomorales: Even though this will work in `if self._has_exchange_record(...` I think it makes more sense to return True or False with `bool(self.env["edi...` than returning the first record it finds.

simahawk: yep, and then replace `search` w/ `search_count`

jarroyomorales (approved):

etobella:
With this, we are ready for `edi_account` and an example with spanish e-invoice system

simahawk (approved):

PR #264, opened 2020-11-26
[13.0] EDI Add crons to automate record status

simahawk:


edi/models/edi_backend.py
```diff
@@ -198,34 +206,88 @@ def _exchange_output_check(self, exchange_record):
...
+    def _output_new_records_domain(self):
+        return [
+            ("type_id.exchange_file_auto_generate", "=", True),
+            ("type_id.direction", "=", "output"),
+            ("edi_exchange_state", "=", "new"),
+            ("exchange_file", "=", False),
+        ]
+
+    def _output_pending_records_domain(self):
+        states = ("output_pending", "output_sent", "output_sent_and_error")
```
etobella, line 279: Why output sent should be pending?

I would also add a check on the type in order to proceed with this. What do you think?

simahawk: because `_check_output_exchange_sync` takes care of sending them as well

simahawk: > I would also add a check on the type in order to proceed with this.

what do you mean? The type's direction is "output", what else?

etobella: Maybe the output will be sent on user decision, not by a cron

It is just an idea (similar to output generation??)

simahawk: hmm yeah, why not... 
I could also split sending from sent here

etobella (approved):

edi/models/edi_backend.py
etobella: We can keep it for version two, so, don't worry :smile:

simahawk:
@etobella if it's ok for you I fast track this as it's the base for the next round.

PR #265, opened 2020-11-27
[13.0] Add edi_storage + add unique identifier

simahawk:
~To be rebased on https://github.com/OCA/edi/pull/266~


edi/models/edi_exchange_record.py
```diff
@@ -47,11 +47,6 @@ class EDIExchangeRecord(models.Model):
         store=True,
         readonly=False,
     )
-    # TODO: use sequence and make it unique
-    exchange_identification_code = fields.Char(
```
etobella, line 51: Why removing this field? It is interesting, this is the identification file code from the other side. Maybe it can be renamed if necessary, but it has a lot of sense. At the end, is what we need to consult the other side of the status of this record.

etobella (commented):
Just some comments regarding a deletion of a field.

Checking the storage backend, Code looks great, I will try to make a test locally :smile:

edi/models/edi_exchange_record.py
simahawk: check the rest of the diff: I called it `identifier` simply :wink:

etobella: I know, but `identifier` is your identifier, `exchange_identification_code` was intended for the `identifier` of the other part (sometimes you could receive a notification with their identifier that can be useful for status update.

Maybe `external_identifier` might be a better name for the field I am talking about

simahawk: aah I see then I vote for `external_identifier` :smile:

simahawk: done

simahawk:
@etobella this one should be ready :wink:

etobella (approved):

PR #268, opened 2020-11-30
[IMP] edi: Security

etobella:
This PR adds security on exchange records following a similar logic than odoo `mail.message`.

### DIFF
--- a/edi/models/edi_exchange_consumer_mixin.py
+++ b/edi/models/edi_exchange_consumer_mixin.py
@@ -45,3 +45,30 @@ class EDIExchangeConsumerMixin(models.AbstractModel):
 
     def _edi_generate_records(self):
         raise NotImplementedError()
+
+    @api.model
+    def get_edi_access(self, doc_ids, operation, model_name=False):
+        """Retrieve access policy.
+        The behavior is similar to `mail.thread` and `mail.message`
+        and it relies on the access rules defines on the related record.
+        The behavior can be customized on the related model
+        by defining `_edi_exchange_record_access`.
+
+        By default `write`, otherwise the custom permission is returned.
+        """
+        DocModel = self.env[model_name] if model_name else self
+        create_allow = getattr(DocModel, "_edi_exchange_record_access", "write")
+        if operation in ["write", "unlink"]:
+            check_operation = "write"
+        elif operation == "create" and create_allow in [
+            "create",
+            "read",
+            "write",
+            "unlink",
+        ]:
+            check_operation = create_allow
+        elif operation == "create":
+            check_operation = "write"
+        else:
+            check_operation = operation
+        return check_operation
--- a/edi/tests/test_security.py
+++ b/edi/tests/test_security.py
@@ -1,0 +1,170 @@
+# Copyright 2020 Creu Blanca
+# @author: Enric Tobella
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo_test_helper import FakeModelLoader
+
+from odoo.exceptions import AccessError
+
+from .common import EDIBackendCommonTestCase
+
+
+class TestEDIExchangeRecordSecurity(EDIBackendCommonTestCase):
+    @classmethod
+    def _setup_records(cls):
+        super()._setup_records()
+        # Load fake models ->/
+        cls.loader = FakeModelLoader(cls.env, cls.__module__)
+        cls.loader.backup_registry()
+        from .fake_models import EdiExchangeConsumerTest
+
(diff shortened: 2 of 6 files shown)


edi/tests/test_security.py
```diff
@@ -0,0 +1,176 @@
+# Copyright 2020 ACSONE
+# @author: Simone Orsi <simahawk@gmail.com>
```
simahawk: this should be you :wink:

etobella:
Now this can be tested :smile: @simahawk @jarroyomorales

edi/models/edi_exchange_consumer_mixin.py
```diff
@@ -45,3 +45,27 @@ def action_view_edi_records(self):
...
     def _edi_generate_records(self):
         raise NotImplementedError()
+
+    @api.model
+    def get_edi_access(self, doc_ids, operation, model_name=False):
+        """We are following the same logic applied on mail.thread and mail.message
+
+        In order to decide if we can create a record is defined by _edi_post_access or
+        by default write, otherwise, the same permission is returned
+        """
```
simahawk:
```suggestion
        """Retrieve access policy.

        The behavior is similar to `mail.thread` and `mail.message`
        and it relies on the access rules defines on the related record.

        The behavior can be customized on the related model
        by defining `_edi_exchange_record_access`.
        
        By default `write`, otherwise the custom permission is returned.
        """
```
Is this meaning correct?

_edi_post_access should be renamed to `_edi_exchange_record_access` or similar IMO,

edi/tests/test_security.py
```diff
@@ -0,0 +1,176 @@
...
+            {
+                "name": "Exchange Record rule demo",
+                "model_id": cls.env.ref("edi.model_edi_exchange_consumer_test").id,
+                "domain_force": "[('name', '=', 'test')]",
+                "groups": [(4, cls.group.id)],
+            }
+        )
+        cls.user = (
+            cls.env["res.users"]
+            .with_context({"no_reset_password": True, "mail_notrack": True})
```
simahawk: use kw args to not override the env completely

edi/tests/test_security.py
```diff
@@ -0,0 +1,176 @@
...
+
+    def test_security_user(self):
+        self.user.write({"groups_id": [(4, self.group.id)]})
+        exchange_record = self.create_record()
+        self.assertTrue(exchange_record)
+
+    def test_security_user_no_rule(self):
+        self.user.write({"groups_id": [(4, self.group.id)]})
+        self.consumer_record.name = "no_rule"
+        with self.assertRaises(AccessError):
```
simahawk: always better to check the error message explicitly

edi/tests/test_security.py
```diff
@@ -0,0 +1,176 @@
...
+            exchange_record.with_user(self.user).unlink()
+
+    def test_search_no_group(self):
+        exchange_record = self.create_record()
+        self.assertEqual(
+            0,
+            len(
+                self.env["edi.exchange.record"]
+                .with_user(self.user)
+                .search([("id", "=", exchange_record.id)])
```
simahawk: use search_count instead?

edi/tests/test_security.py
```diff
@@ -0,0 +1,176 @@
...
+            ),
+        )
+
+    def test_write_no_group(self):
```
simahawk: maybe it's better to rename these tests like `test_user_no_write` and alike?

simahawk (changes requested):
great work! Some small remarks.

etobella:
@simahawk Comments attended :smile:

edi/models/edi_exchange_consumer_mixin.py
```diff
@@ -45,3 +45,30 @@ def action_view_edi_records(self):
...
+        """Retrieve access policy.
+        The behavior is similar to `mail.thread` and `mail.message`
+        and it relies on the access rules defines on the related record.
+        The behavior can be customized on the related model
+        by defining `_edi_exchange_record_access`.
+
+        By default `write`, otherwise the custom permission is returned.
+        """
+        DocModel = self.env[model_name] if model_name else self
+        create_allow = getattr(DocModel, "_edi_post_access", "write")
```
simahawk: _edi_post_access shouldn't be `_edi_exchange_record_access`?

etobella: True, I forgot :facepalm:

etobella:
@simahawk I think this is ready for review.

simahawk (approved):

PR #271, opened 2020-12-02
[13.0] EDI refactor ack handling w/ its own type

simahawk:
NOTE: edi_storage ack handling has been put on hold because we still have to implement automatic process of incoming (eg: receive/retrieve files from outside).

### DIFF
--- a/edi/views/edi_exchange_record_views.xml
+++ b/edi/views/edi_exchange_record_views.xml
@@ -53,12 +53,13 @@
                             <field name="exchange_filename" invisible="1" />
                             <field name="direction" invisible="1" />
                         </group>
-                        <group
-                            name="ack"
-                            attrs="{'invisible': [('ack_needed', '=', False)]}"
-                        >
+                        <!-- TODO: move to a page? -->
+                        <group name="related">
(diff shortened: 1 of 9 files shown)


simahawk:
@etobella  rebased and ready for review

edi/views/edi_exchange_record_views.xml
```diff
@@ -53,12 +53,13 @@
                             <field name="exchange_filename" invisible="1" />
                             <field name="direction" invisible="1" />
                         </group>
-                        <group
-                            name="ack"
-                            attrs="{'invisible': [('ack_needed', '=', False)]}"
-                        >
+                        <!-- TODO: move to a page? -->
```
etobella, line 56: Yes, probably better there, maybe invisible if the type has not an ack_type_id defined

simahawk: I'd leave the TODO there for now. We can change it later.
What about the rest? Approve-able? :)

etobella (approved):
Yes, it is, just was a comment, and can be managed later, jejeje

PR #272, opened 2020-12-02
[13.0] EDI add file reception + auto generate ack files when needed

simahawk:
Based on/includes #271 (to be merged 1st).
Only last 4 commits are relevant (excluding a fixup)

### DIFF
--- a/edi/models/edi_backend.py
+++ b/edi/models/edi_backend.py
@@ -425,7 +425,7 @@ class EDIBackend(models.Model):
         for backend in self:
             backend._check_input_exchange_sync(**kw)
 
-    def _check_input_exchange_sync(self):
+    def _check_input_exchange_sync(self, **kw):
         """Lookup for pending input records and take care of them.
 
         First work on records that need to receive input.
--- a/edi/models/edi_backend.py
+++ b/edi/models/edi_backend.py
@@ -358,3 +358,105 @@ class EDIBackend(models.Model):
         if component:
             return component.process()
         raise NotImplementedError()
+
+    def exchange_receive(self, exchange_record):
+        """Retrieve an incoming document.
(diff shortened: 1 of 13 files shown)


edi/models/edi_backend.py
```diff
@@ -358,3 +364,122 @@ def _exchange_process(self, exchange_record):
...
+        component = self._get_component(exchange_record, "receive")
+        if component:
+            return component.receive()
+        raise NotImplementedError()
+
+    def _cron_check_input_exchange_sync(self, **kw):
+        for backend in self:
+            backend._check_input_exchange_sync(**kw)
+
+    def _check_input_exchange_sync(self):
```
etobella: Missing the **kw, as they might be passed by the cron?

edi/models/edi_backend.py
```diff
@@ -358,3 +364,122 @@ def _exchange_process(self, exchange_record):
...
+        ]
+
+    def _exchange_receive(self, exchange_record):
+        component = self._get_component(exchange_record, "receive")
+        if component:
+            return component.receive()
+        raise NotImplementedError()
+
+    def _cron_check_input_exchange_sync(self, **kw):
+        for backend in self:
```
etobella, line 425: I think related cron is commented, with this implementation we might add it

etobella (approved):
Just some minor comments

edi/models/edi_backend.py
simahawk: I prefer to wait when tests will be there :wink:

edi/models/edi_backend.py
simahawk: good catch... you see? No tests... will come soon on next round :wink:

PR #213, opened 2020-06-23
[13.0][MIG] account_invoice_import

yvaucher:
With changes of model from `account.invoice` to `account.move`

TODO:

- [x] Fix dashboard
- [x] Check if account_id is still necessary on taxes (in unit tests)

Replaced by #727

PR #273, opened 2020-12-02
[WIP] [TMP] [13.0] mig account invoice import

NL66278:
DO NOT MERGE

PR to check wether I can solve the problems in the existing PR for this migration.


etobella:
@NL66278 We have been working on an standard interface for integrating all `edi` methods. It might be interesting to do it with `account_invoice_import` If you want, I can do a small PR on your branch integrating it. WDYT?

NL66278:
@etobella I would definitely be interested!

etobella:
Thanks!!!! Working on it :smile:

etobella:
https://github.com/tegin/edi/tree/13.0-mig-account_invoice_import/account_invoice_import
This is how it will be managed. Most part of the code will be stored on the component.

NL66278:
@etobella If you put a pull request on my branch, I can see the differences... What is the general idea?

etobella:
I made a rebase, So I don't know if I can make a PR on your branch :thinking: But I can do something
The main idea is to use `edi.exchange.record`s and `edi.backend`.
Each `account.invoice.import.config` is a backend, a configuration to import data.
The `edi.exchange.record` is a document that has been received (or sent) by odoo (this time it is sent by the user to odoo).
Data of this record is the final dict obtained and it is processed.
I am currently unable to remove or simplify much more the wizard, as it has several choices for the user, but all the logic has been moved to a component, that will manage the changes.

If you have doubts we can make a call in order to share the points and show you the main goal of this framework.

etobella:
https://github.com/NL66278/edi/pull/1 You can check the changes here.

NL66278:
@etobella I looked at your commit.

Can you explain the advantage of making everything a "component"? I have looked at the component module code a view times now and I can not make heads or tails out of it. To me it seems to add abstraction layer on abstraction layer without any benefit I can detect. But maybe it is something very beautiful that this simple mind does not understand, despite working in IT now for over 30 years....

etobella:
@NL66278 I can try to explain it.

Components are sets of code related to a base model, called  ´collection´. There are similarities with models, but also differences:

* They have an inheritance system
* They have no records
* We usually decide which component to use by the `_usage` key and, if necessary, the assigned model (defined as a list on `_apply_on`)
* They might be AbstractComponent (only used for inheritance) or Component

Why they are useful on this case, because we want to create a unique framework that defines the integration between systems (in this particular case, importing invoices).
On the `edi` modules, we define 3 configuration modules:
* `edi.exchange.type` defines the kind of files that can be exchanged, direction(from odoo or to odoo), type of file, etc.
* `edi.backend` defines the exact configuration settings to be used
* `edi.backend.type` defines a common relation between the previous models. They define part of the component required code

I will try to show an example:
The backend type can be UBL files, that is simply a standard.
The exchange type will be each kind of UBL files (purchase.order, sale.order, stock.picking, etc...) or their response (when I send a sale.order I might expect a response from my customer)
The backend will be the exact configuration used to send this data. I might have different backends according to different reasons even for the same exchange type (some customers want to receive it in a WebService, other by an SFTP, etc...)

So, why I am proposing to apply this change in this module (my approach is far from definitive, but it is a first approach)? Because at the end, each received and imported invoice is an exchange record and it might have sense to store it as an exchange record. We could even respond to our supplier that we have received an processed it or that there is an issue. 

I think that my proposal is not complete and we might need to do a complete refactoring to implement it correctly, but, at the end, it is a good initial approach.

I hope that it solved part of your doubts. If you need more explanation I will be happy to try to solve it or do a call to solve them.

etobella:
Just a small comment, I know that all of this could be solved without components creating the expected functions and so on. But, at the end, we are creating small snippets of code with components. With functions, we are complicating models with a set of functions that are hard to inherit and manage.
With small and controlled cases it is better to use functions. IMO, if we try to make a global framework, the components might be easier to use.

account_invoice_import/wizard/account_invoice_import.py
```diff
@@ -0,0 +1,1130 @@
...
+                parsed_inv[entry] *= -1
+            for line in parsed_inv.get("lines", []):
+                line["qty"] *= -1
+                if "price_subtotal" in line:
+                    line["price_subtotal"] *= -1
+        # Handle taxes:
+        self._pre_process_parsed_inv_taxes(parsed_inv)
+        # Handle rounding:
+        for line in parsed_inv.get("lines", []):
+            line["qty"] = float_round(line["qty"], precision_digits=prec_uom)
```
CasVissers-360ERP, line 407: There is an exception possibility here:

Invoice line qty = 0,5 / price = 100
Product Unit of Measure = 0 

When importing the quantity will be set to 1, but the price is not changed accordingly. Result is an unclear debit/credit error from Odoo.

account_invoice_import/wizard/account_invoice_import.py
```diff
@@ -0,0 +1,1130 @@
...
+        "account.invoice.import.config", string="Invoice Import Configuration"
+    )
+    currency_id = fields.Many2one("res.currency", readonly=True)
+    invoice_type = fields.Selection(
+        [("in_invoice", "Invoice"), ("in_refund", "Refund")],
+        string="Invoice or Refund",
+        readonly=True,
+    )
+    amount_untaxed = fields.Float(
+        string="Total Untaxed", digits=dp.get_precision("Account"), readonly=True
```
CasVissers-360ERP, line 48: digits='Account'

account_invoice_import/wizard/account_invoice_import.py
```diff
@@ -0,0 +1,1130 @@
...
+    )
+    amount_total = fields.Float(
+        string="Total", digits=dp.get_precision("Account"), readonly=True
```
CasVissers-360ERP, line 51: digits='Account'

CasVissers-360ERP (changes requested):

yvaucher:
@NL66278 thanks for the work I'll integrate your change and squash my commits in #213 then I'll try to integrate NL66278#1

PR #274, opened 2020-12-03
[IMP] edi: Detection of Pending Manual EDIs and adding it to form view

etobella:
This PR adds a hook on account.move for manual generation of EDIs. Another option is to add directly on EDI Mixin :smile: 

@simahawk

### DIFF
--- a/edi/models/edi_exchange_consumer_mixin.py
+++ b/edi/models/edi_exchange_consumer_mixin.py
@@ -3,7 +3,13 @@
 # @author Simone Orsi <simahawk@gmail.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
+from datetime import datetime, time
+
+import dateutil
+from lxml import etree
+
 from odoo import api, fields, models
+from odoo.tools.safe_eval import safe_eval
 
 
 class EDIExchangeConsumerMixin(models.AbstractModel):
@@ -19,15 +25,127 @@ class EDIExchangeConsumerMixin(models.AbstractModel):
         domain=lambda r: [("model", "=", r._name)],
     )
     exchange_record_count = fields.Integer(compute="_compute_exchange_record_count")
+    expected_edi_configuration = fields.Serialized(
+        compute="_compute_expected_edi_configuration", default={},
+    )
+    has_expected_edi_configuration = fields.Boolean(
+        compute="_compute_expected_edi_configuration"
+    )
 
-    def _has_exchange_record(self, exchange_type, backend, extra_domain=False):
+    def _compute_expected_edi_configuration(self):
+        for record in self:
+            configurations = record._get_expected_edi_configuration()
+            record.expected_edi_configuration = configurations
+            record.has_expected_edi_configuration = bool(configurations)
+
+    def _get_expected_edi_configuration(self):
+        exchange_types = (
+            self.env["edi.exchange.type"]
+            .sudo()
+            .search([("model_ids.model", "=", self._name)])
+        )
+        result = {}
+        for exchange_type in exchange_types:
+            eval_ctx = dict(
+                self._get_eval_context(), record=self, exchange_type=exchange_type
+            )
+            domain = safe_eval(exchange_type.enable_domain or "[]", eval_ctx)
+            if not self.filtered_domain(domain):
+                continue
+            if exchange_type.enable_snippet:
+                safe_eval(
+                    exchange_type.enable_snippet, eval_ctx, mode="exec", nocopy=True
+                )
+                if not eval_ctx.get("result", False):
+                    continue
+            result[exchange_type.id] = exchange_type.display_name
+        return result
+
+    def _get_eval_context(self):
+        """ Prepare the context used when evaluating python code
+            :returns: dict -- evaluation context given to safe_eval
+        """
+        return {
+            "datetime": datetime,
+            "dateutil": dateutil,
+            "time": time,
+            "uid": self.env.uid,
+            "user": self.env.user,
+        }
+
+    @api.model
+    def fields_view_get(
+        self, view_id=None, view_type="form", toolbar=False, submenu=False
+    ):
+        res = super().fields_view_get(
+            view_id=view_id, view_type=view_type, toolbar=toolbar, submenu=submenu
+        )
+        if view_type == "form":
+            doc = etree.XML(res["arch"])
+            for node in doc.xpath("//sheet"):
+                group = False
+                if hasattr(self, "_edi_generate_group"):
+                    group = self._edi_generate_group
+                str_element = self.env["ir.qweb"].render(
+                    "edi.edi_exchange_consumer_mixin_buttons", {"group": group},
+                )
+                node.addprevious(etree.fromstring(str_element))
+            View = self.env["ir.ui.view"]
+
+            # Override context for postprocessing
+            if view_id and res.get("base_model", self._name) != self._name:
+                View = View.with_context(base_model_name=res["base_model"])
+            new_arch, new_fields = View.postprocess_and_fields(self._name, doc, view_id)
+            res["arch"] = new_arch
+            res["fields"] = new_fields
+        return res
+
+    def _edi_create_exchange_record_vals(self, exchange_type):
+        return {
+            "model": self._name,
+            "res_id": self.id,
+        }
+
+    def _edi_create_exchange_record(self, exchange_type, backend):
+        exchange_record = backend.create_record(
+            exchange_type.code, self._edi_create_exchange_record_vals(exchange_type)
+        )
+        self._event("on_edi_generate_manual").notify(self, exchange_record)
+        return exchange_record.get_formview_action()
+
+    def edi_create_exchange_record(self, exchange_type_id):
+        self.ensure_one()
+        exchange_type = self.env["edi.exchange.type"].browse(exchange_type_id)
+        backend = exchange_type.backend_id
+        if (
+            not backend
+            and self.env["edi.backend"].search_count(
+                [("backend_type_id", "=", exchange_type.backend_type_id.id)]
+            )
+            == 1
+        ):
+            backend = self.env["edi.backend"].search(
+                [("backend_type_id", "=", exchange_type.backend_type_id.id)]
+            )
+        if backend:
+            return self._edi_create_exchange_record(exchange_type, backend)
+        action = self.env.ref("edi.edi_exchange_record_create_act_window").read()[0]
+        action["context"] = {
+            "default_res_id": self.id,
+            "default_model": self._name,
+            "default_exchange_type_id": exchange_type_id,
+        }
+        return action
+
+    def _has_exchange_record(self, exchange_type, backend=False, extra_domain=False):
         """This function is useful when generating the configuration"""
         domain = [
             ("model", "=", self._name),
             ("res_id", "=", self.id),
-            ("backend_id", "=", backend.id),
             ("type_id.code", "=", exchange_type),
         ]
+        if backend:
+            domain.append(("backend_id", "=", backend.id))
         if extra_domain:
             domain += extra_domain
         return bool(self.env["edi.exchange.record"].search_count(domain))
@@ -42,9 +160,6 @@ class EDIExchangeConsumerMixin(models.AbstractModel):
         action = self.env.ref("edi.act_open_edi_exchange_record_view").read()[0]
         action["domain"] = [("model", "=", self._name), ("res_id", "=", self.id)]
         return action
-
-    def _edi_generate_records(self):
-        raise NotImplementedError()
 
     @api.model
     def get_edi_access(self, doc_ids, operation, model_name=False):
--- a/edi/readme/CONFIGURE.rst
+++ b/edi/readme/CONFIGURE.rst
@@ -1,0 +1,29 @@
+This module aims to provide an infrastructure to simplify interchangability of documents
+between systems providing a configuration platform.
+It will be inherited by other modules in order to define the proper implementations of
+components.
+
+In order to define a new Exchange Record, we need to configure:
+
+* Backend Type
+* Exchange Type
+* Backend
+* Components
+
+Component definition
+~~~~~~~~~~~~~~~~~~~~
+
+The component usage must be defined like `edi.{direction}.{kind}.{code}` where:
+
+* direction is `output` or `input`
+* kind can be: `generate`, `send`, `check`, `process`, `receive`
+* code is the `{backend type code}` or `{backend type code}.{exchange type code}`
+
+User EDI generation
+~~~~~~~~~~~~~~~~~~~
+
+On the exchange type, it might be possible to define a set of models, a domain and a
+snippet of code.
+After defining this fields, we will automatically see buttons on the view to generate
+the exchange records.
+This configuration is useful to define a way of generation managed by user.
--- a/edi/readme/USAGE.rst
+++ b/edi/readme/USAGE.rst
@@ -1,0 +1,31 @@
+After certain operations or manual execution, Exchange records will be generated.
+This Exchange records might be input records or outputs records.
+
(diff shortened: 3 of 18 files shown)


edi_account/models/account_move.py
```diff
@@ -29,3 +46,9 @@ def action_invoice_paid(self):
         if self:
             self._event("on_paid_account_move").notify(self)
         return result
+
+    def action_generate_edi(self):
+        """This launches a notification for generating missing manual EDIs"""
+        self._event("on_generate_account_edi").notify(
```
simahawk: This is going to trigger the event even if you have not records.
Also what would be the behavior?
Maybe we can discuss how to do it generally...

simahawk: As discussed in call, we could have:
1. model_ids on exc type
2. enable_domain to filter records matching the type
3. enable_snippet here you could do what you want and return true/false when it applies or not to the current record

The next step would be: automatically show buttons depending on this based on a serialized computed field that tells you per exchange type which btn show.

etobella:
@simahawk First approach of what we talked about. Stil pending some tests and some extra documentation :smile:

etobella:
Tests have been improved, the error on runbot is solved on #268

etobella:
image

This is how it will be shown :smile:

jarroyomorales (approved):
Tested and works as expected! :+1:

simahawk:
@etobella seems amazing! Thanks! I'll review later or tomorrow.

edi/models/edi_exchange_consumer_mixin.py
```diff
@@ -19,6 +25,104 @@ class EDIExchangeConsumerMixin(models.AbstractModel):
...
+        for record in self:
+            configurations = record._get_expected_edi_configuration()
+            record.expected_edi_configuration = configurations
+            record.has_expected_edi_configuration = bool(configurations)
+
+    def _get_expected_edi_configuration(self):
+        exchange_types = (
+            self.env["edi.exchange.type"]
+            .sudo()
+            .search([("model_ids.model", "=", self._name), ("backend_id", "!=", False)])
```
simahawk: why backend_id False? This excludes all types that are not assigned to a specific backend but only to a backend type.

edi/models/edi_exchange_consumer_mixin.py
```diff
@@ -19,6 +25,104 @@ class EDIExchangeConsumerMixin(models.AbstractModel):
...
+                    "record": self,
+                    "exchange_type": exchange_type,
+                }
+                eval_ctx.update(self._get_eval_context())
+                safe_eval(
+                    exchange_type.enable_snippet, eval_ctx, mode="exec", nocopy=True
+                )
+                if not eval_ctx.get("result", False):
+                    continue
+            result[exchange_type.id] = exchange_type.display_name
```
simahawk: do not search for the record, you have it.  Use `filtered_domain` to apply the domain on it:
```suggestion
        for exchange_type in exchange_types:
            eval_ctx = dict(
                self._get_eval_context(),
                record=self,
                exchange_type=exchange_type
            )
            domain = safe_eval(
                exchange_type.enable_domain or "[]", eval_ctx
            )
            if not self.filtered_domain(domain):
                continue
            if exchange_type.enable_snippet:
                safe_eval(
                    exchange_type.enable_snippet, eval_ctx, mode="exec", nocopy=True
                )
                if not eval_ctx.get("result", False):
                    continue
            result[exchange_type.id] = exchange_type.display_name
```

edi/models/edi_exchange_consumer_mixin.py
```diff
@@ -19,6 +25,104 @@ class EDIExchangeConsumerMixin(models.AbstractModel):
         domain=lambda r: [("model", "=", r._name)],
     )
     exchange_record_count = fields.Integer(compute="_compute_exchange_record_count")
+    expected_edi_configuration = fields.Serialized(
+        compute="_compute_expected_edi_configuration"
```
simahawk:
```suggestion
        compute="_compute_expected_edi_configuration"
        default={}
```

edi/models/edi_exchange_consumer_mixin.py
```diff
@@ -19,6 +25,104 @@ class EDIExchangeConsumerMixin(models.AbstractModel):
...
+        exchange_type = self.env["edi.exchange.type"].browse(exchange_type_id)
+        backend = exchange_type.backend_id
+        exchange_record = backend.create_record(
+            exchange_type.code, self._generate_edi_configuration_vals(exchange_type)
+        )
+        component = backend._get_component(exchange_record, "manual")
+        # TODO: Define which key the component should have
+        if component:
+            # TODO: Define the function name properly
+            component.manual_generation()
```
simahawk: why another component? What's manual? I think the action should depend on what is missing:
1. missing a record? create it
2. missing data? Receive it
3. missing send? send it
etc...

simahawk: All in all it's not clear what's the purpose of this method :thinking: :sweat_smile:

edi/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,26 @@
+This module is intended to be inherited by other modules in order to define the proper
```
simahawk: yes and no... the aim is also to allow to not have to code anything and just work w/ configuration

edi/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,26 @@
...
+implementation.
+
+In order to define a new Exchange Record, we need to configure:
+
+* Backend Type
+* Exchange Type
+* Backend
+* Connectors
+
+Connector definition
```
simahawk: Connector != component :wink:

edi/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,26 @@
...
+~~~~~~~~~~~~~~~~~~~~
+
+The connector usage must be defined like `edi.{direction}.{kind}.{code}` where:
+
+* direction is `output` or `input`
+* kind can be: `generate`, `send`, `check`, `process`, `receive`
+* code is the `{backend type code}` or `{backend type code}.{exchange type code}`
+
+Manual EDI generation
```
simahawk: To be clarified

edi/readme/USAGE.rst
```diff
@@ -0,0 +1,31 @@
...
+An output record is intended to be used for exchange information from Odoo to another
+system.
+
+The flow of an output record should be:
+
+* Creation
+* Generation of data
+* Checking of data
+* Sending data
+* Reviewing if the data was processed properly by the other party
```
simahawk: Validation

edi/readme/USAGE.rst
```diff
@@ -0,0 +1,31 @@
...
+Input Exchange records
+~~~~~~~~~~~~~~~~~~~~~~
+
+An input record is intended to be used for exchange information another system to odoo.
+
+The flow of an input record should be:
+
+* Creation
+* Reception of data
+* Checking data
```
simahawk, line 30: Validation

edi/tests/test_consumer_mixin.py
```diff
@@ -7,13 +7,15 @@
...
+from odoo.tests.common import Form
+
+from .common import EDIBackendCommonComponentRegistryTestCase
 
 
 # This clashes w/ some setup (eg: run tests w/ pytest when edi_storage is installed)
 # If you still want to run `edi` tests w/ pytest when this happens, set this env var.
 @unittest.skipIf(os.getenv("SKIP_EDI_CONSUMER_CASE"), "Consumer test case disabled.")
-class TestConsumerMixinCase(EDIBackendCommonTestCase):
+class TestConsumerMixinCase(EDIBackendCommonComponentRegistryTestCase):
```
simahawk: the component savepoint case is enough as you don't register any new component on the fly

edi/tests/test_consumer_mixin.py
```diff
@@ -49,3 +51,50 @@ def test_mixin(self):
...
+        self.assertFalse(self.consumer_record.has_expected_edi_configuration)
+        self.assertEqual({}, self.consumer_record.expected_edi_configuration)
+        self.exchange_type_out.write(
+            {
+                "model_ids": [
+                    (
+                        4,
+                        self.env["ir.model"]
+                        .search([("model", "=", self.consumer_record._name)], limit=1)
+                        .id,
```
simahawk:
```suggestion
                        self.env['ir.model']._get_id(self.consumer_record._name),
```

edi/tests/test_consumer_mixin.py
```diff
@@ -49,3 +51,50 @@ def test_mixin(self):
...
+        self.consumer_record.refresh()
+        self.assertFalse(self.consumer_record.has_expected_edi_configuration)
+        self.assertEqual({}, self.consumer_record.expected_edi_configuration)
+        self.assertTrue(self.consumer_record.exchange_record_ids)
+        self.assertEqual(
+            self.consumer_record.exchange_record_ids.type_id, self.exchange_type_out
+        )
+
+    def test_form(self):
+        """Testing that the form has inherited the fields and inserted them
```
simahawk, line 132: I would test the the form xml instead :wink:

simahawk (changes requested):

edi/models/edi_exchange_consumer_mixin.py
etobella: Because I need a backend for generating the exchange record and I only have the record and the exchange type :thinking:

edi/models/edi_exchange_consumer_mixin.py
etobella: Yes, you are right, I will add an event just in case and keeping it the simplest way :wink:

etobella:
@simahawk comments attended

edi/models/edi_exchange_consumer_mixin.py
simahawk: that's why we should have a wizard IMO to select a backend or auto assign it IF there's only one backend for that specific type.backend_type_id.

edi/models/edi_exchange_consumer_mixin.py
```diff
@@ -19,6 +25,96 @@ class EDIExchangeConsumerMixin(models.AbstractModel):
...
+            res["fields"] = new_fields
+        return res
+
+    def _generate_edi_configuration_vals(self, exchange_type):
+        return {
+            "model": self._name,
+            "res_id": self.id,
+        }
+
+    def generate_edi_configuration(self, exchange_type_id):
```
simahawk: this should be named `edi_create_exchange_record` and their related methods renamed accordingly.

edi/models/edi_exchange_consumer_mixin.py
```diff
@@ -19,6 +25,96 @@ class EDIExchangeConsumerMixin(models.AbstractModel):
...
+        self.ensure_one()
+        exchange_type = self.env["edi.exchange.type"].browse(exchange_type_id)
+        backend = exchange_type.backend_id
+        exchange_record = backend.create_record(
+            exchange_type.code, self._generate_edi_configuration_vals(exchange_type)
+        )
+        self._event("on_generate_manual_edi").notify(self, exchange_record)
```
simahawk: I don't think you need this event: you can already hook to `on_create` no?
What's the goal you had in mind?

edi/models/edi_exchange_consumer_mixin.py
etobella: True, I will work on it :wink:

edi/models/edi_exchange_consumer_mixin.py
etobella: The goal of this listener is to allow configuration to process the record if needed (it depends on the listener). Why not using the on_create, because the on_create does not know why the record was generated, so it is always better to reuse it (different behavior on standard creation and manual creation?)

edi/models/edi_exchange_consumer_mixin.py

edi/models/edi_exchange_consumer_mixin.py

etobella:
@simahawk Comments attended or answered :wink:

edi/models/edi_exchange_consumer_mixin.py
```diff
@@ -19,13 +25,114 @@ class EDIExchangeConsumerMixin(models.AbstractModel):
...
+        exchange_record = backend.create_record(
+            exchange_type.code, self._edi_create_exchange_record_vals(exchange_type)
+        )
+        self._event("on_generate_manual_edi").notify(self, exchange_record)
+        return exchange_record.get_formview_action()
+
+    def edi_create_exchange_record(self, exchange_type_id):
+        self.ensure_one()
+        exchange_type = self.env["edi.exchange.type"].browse(exchange_type_id)
+        if exchange_type.backend_id:
```
simahawk: you can do the same if you find only one backend for exchange_type.backend_type_id :wink:

edi/models/edi_exchange_consumer_mixin.py
simahawk: ok, call it `on_edi_generate_manual`

edi/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,28 @@
...
+The component usage must be defined like `edi.{direction}.{kind}.{code}` where:
+
+* direction is `output` or `input`
+* kind can be: `generate`, `send`, `check`, `process`, `receive`
+* code is the `{backend type code}` or `{backend type code}.{exchange type code}`
+
+Manual EDI generation
+~~~~~~~~~~~~~~~~~~~~~
+
+Also, the Exchange can be accessed on debug mode and edit `model_ids`, `enable_domain`
```
simahawk: this should be rephrased...

edi/views/edi_exchange_type_views.xml
```diff
@@ -35,6 +35,15 @@
                             <field name="exchange_file_auto_generate" />
                             <field name="ack_type_id" />
                         </group>
+                        <group name="manual" group="base.group_no_one">
```
simahawk: when #275  is merged we can use the specific group for advanced settings I added there :wink:

etobella:
Comments attended, still missing some tests for increasing coverage

edi/models/edi_exchange_consumer_mixin.py
etobella: OK

edi/views/edi_exchange_type_views.xml
etobella: OK

etobella:
@simahawk Comments attended and tests added

etobella:
Rebased and fixed conflicts, I think that the main issue here is the readme. If you want, I can remove it from here and create a PR just for this

@simahawk

simahawk:
> Rebased and fixed conflicts, I think that the main issue here is the readme. If you want, I can remove it from here and create a PR just for this
> 
> @simahawk

yes please :pray:

etobella:
Done

edi/wizards/edi_exchange_record_create.py
```diff
@@ -0,0 +1,24 @@
...
+    _name = "edi.exchange.record.create"
+    _description = "Create an Exchange Record"
+
+    res_id = fields.Integer(required=True)
+    model = fields.Char(required=True)
+    exchange_type_id = fields.Many2one("edi.exchange.type", required=True)
+    backend_type_id = fields.Many2one(
+        "edi.backend.type", related="exchange_type_id.backend_type_id"
+    )
+    backend_id = fields.Many2one("edi.backend", reaquired=True)
```
simahawk:
```suggestion
    backend_id = fields.Many2one("edi.backend", required=True)
```

simahawk (approved):
small typo

edi/wizards/edi_exchange_record_create.py
etobella: fixed

PR #275, opened 2020-12-03
[13.0] edi: misc improvements

simahawk:
commits are self explanatory


edi/models/edi_backend.py
```diff
@@ -374,8 +399,12 @@ def exchange_receive(self, exchange_record):
         check = self._exchange_receive_check(exchange_record)
         if not check:
             return False
+        state = exchange_record.edi_exchange_state
+        error = False
+        message = None
+        content = None
         try:
-            self._exchange_receive(exchange_record)
+            content = self._exchange_receive(exchange_record)
```
simahawk, line 419: move validate data here and raise a specific exception and catch it, same for output generation
then add 1 new state for output like `output_generate_error` and switch to it.

etobella:
Travis is failing due to:
* xmlunittest
* An error on a test that has not been modified :thinking:

simahawk:
@etobella ready here! I fixed / improved tests.

etobella (approved):

Issue #276, opened 2020-12-03
Switch to MQT=PIP?


simahawk:
Anyone against?

simahawk:
@bealdav @etobella @alexis-via ping :)

bealdav:
+1 for pip at least on v14

etobella:
No problem for me

simahawk:
What about v13? I'd like to start from there. If not, at least we can use the new feature from setuptools odoo to generate requirements.txt automatically https://github.com/OCA/oca-addons-repo-template/pull/31/
I'll sync this repo w/ copier which will help w/ future updates as well.

simahawk:
There we go https://github.com/OCA/edi/pull/277 (no switch yet)

etobella:
Closing as the PR was merged

PR #277, opened 2020-12-06
Sync repo w/ oca-addons-repo-template + switch to MQT=PIP

simahawk:
I'm keeping MQT=OCA Instead of PIP as mentioned in #276 but if there's enough consensus we can switch that too.


requirements.txt
```diff
@@ -1,3 +1,3 @@
-freezegun
 factur-x
-odoo_test_helper
```
simahawk, line 3: @sbidoul I had to enable requirements generation manually as it seems to be considered for v14 only in the template.
Also, it wipes existing test-only dependency: are we supposed to edit `setup/$addon/setup.py` and use `test_require` in such cases?

sbidoul: @simahawk You can put test dependencies in [test-requirements.txt](https://github.com/OCA/maintainer-quality-tools/blob/c9864146f4ee19dd5a7c7bf5a0ea7001ff4700cd/travis/travis_install_nightly#L130).

simahawk: @sbidoul yeah I know but if you don't use MQT=PIP it won't be taken into account AFAIK.
Here I was trying and hybrid approach: have requirements.txt auto-generated but keep MQT=OCA for v13 only.
I'm happy to switch to PIP mode though.

simahawk: Wondering: can't we introduce a new key in manifest to handle this?
It has been discussed here https://github.com/odoo/odoo/issues/40188 but led to nothing so far.
Then we can generate test-requirements.txt automatically if anybody needs it (not needed for pure pip install but handy for documentation or other kind of deployments).
I think it won't hurt anybody...

sbidoul: @simahawk yes, such a new key would be the best. We could try reviving that Odoo issue.
If approved I'd be happy to extend setuptools-odoo to generate a `test` extra with test dependencies.

Other than that I don't see a good solution. IMHO it's best that `requirements.txt' at the repo level does not contain test dependencies so users of the repo don't install stuff they don't actually need at runtime.

simahawk: > @simahawk yes, such a new key would be the best. We could try reviving that Odoo issue.
> If approved I'd be happy to extend setuptools-odoo to generate a `test` extra with test dependencies.

Likely Odoo won't go in that direction soon. I've ping'ed ppl there. Let's wait but meanwhile we can propose an OCA path for this to contributors.
 
> Other than that I don't see a good solution. IMHO it's best that `requirements.txt' at the repo level does not contain test dependencies so users of the repo don't install stuff they don't actually need at runtime.

Sure, requirements.txt should stay clean.

simahawk:
build fails for missing pyyaml dependency https://github.com/OCA/server-tools/pull/1946

simahawk:
@bealdav @etobella ready for review here.
Runbot is orange, not red, for a warning that I fixed already in #275

bealdav (approved):
I don't really understand how does it works. I prefer let others approve

etobella (approved):
It seems that is working as expected, and I don't see any issues on the code review. So good for me, however, we should need to wait the PR fixing Runbot error. Isn't it?

simahawk:
> It seems that is working as expected, and I don't see any issues on the code review. So good for me, however, we should need to wait the PR fixing Runbot error. Isn't it?

nope, is not red as I mentioned. And this change is going to fix the build on that PR :wink:

PR #278, opened 2020-12-08
[8.0][FIX] base_ubl: Backport recent 10.0 fixes

thomaspaulb:
Backport to base_ubl v8 of changes made in base_ubl V10:

- [x] https://github.com/OCA/edi/pull/58
- [x] https://github.com/OCA/edi/pull/199


StefanRijnhart:
Is there a reason that all those back and forward ports of #58 don't include the bit about the payment identifier, which seems the main topic of that PR?

thomaspaulb:
@astirpe Good question, why isn't it in the 11.0 and 12.0 ports?

PR #279, opened 2020-12-08
[13.0] edi: add jobs conf

simahawk:


edi/models/edi_backend.py
```diff
@@ -131,7 +131,7 @@ def _get_exchange_type_domain(self, code):
         ]
 
     # TODO: rename to `exchange_generate` to match other methods
-    def generate_output(self, exchange_record, store=True, force=False, **kw):
+    def exchange_generate(self, exchange_record, store=True, force=False, **kw):
```
etobella, line 161: Good to harmonize names, maybe you can rename the comment now :wink:

etobella (approved):
LGTM :+1: 

Right now you are only adding the job configuration, no actually using it. Do you intend to do this here?

simahawk:
> Right now you are only adding the job configuration, no actually using it. Do you intend to do this here?

Well, maybe I'll use them in crons. Everywhere else the developer should take care of it IMO. Buttons should always use them of course.

What do you think?

etobella:
Good for me :smile:

edi/models/edi_backend.py
```diff
@@ -158,7 +158,7 @@ def _get_exchange_type_domain(self, code):
         ]
 
     # TODO: rename to `exchange_generate` to match other methods
```
etobella: This is not needed anymore

edi/models/edi_backend.py
```diff
@@ -323,7 +322,7 @@ def _check_output_exchange_sync(self, skip_send=False):
             len(new_records),
         )
         for rec in new_records:
-            self.generate_output(rec)
+            self.with_delay().exchange_generate(rec)
```
etobella, line 326: The only thing that makes me crazy with this change is that in order to send a record, it must enter two times on the cron, one for generating and one for sending :upside_down_face:

simahawk: what do you mean? The cron will run once and schedule one job per each record.

etobella: When the cron is launched he will check if a record is generated, and then he will create the jobs, then, it will check the pending to send, and will create the jobs. The jobs will not be executed until all has finished.
If we have a exchange record not generated, it will enter on the first task, but not on the second, as it is not generated yet (it will be once all has finished)

simahawk: as discussed: we don't care. At worse the record will be sent 1h later. If this is an issue you can tweak the cron to run every 15 or 30m.

acsonefho (approved):
Code LGTM!

PR #280, opened 2020-12-09
[10.0][IMP] add despatch advice import

lmarion-source:
includes #244


bealdav:
Hi @lmarion-source and @lmignon 

Thanks to share your work on despatch-advice as a new UBL flow. Did you considered to put this flow in an other module than purchase_order_import ? Maybe a module `despatch_advice_ubl` or other name relative to delivery.

If yes, I think backport and forward port could be made simple.

What do you think ?

bealdav:
perfect

bealdav (approved):

lmignon:
@lmarion-source Can you fix pytlint PLZ https://travis-ci.com/github/OCA/edi/jobs/482189881#L326
Thanks!

despatch_advice_import/wizard/despatch_advice_import.py
```diff
@@ -136,16 +136,22 @@ def process_document(self):
...
         po_name = parsed_order_document.get("ref")
-        order = self.env["purchase.order"].search([("name", "=", po_name)])
-
-        if not order:
-            bdio.user_error_wrap(_("No purchase order found for name %s.") % po_name)
 
         lines = parsed_order_document.get("lines")
         lines_by_id = {int(line["line_id"]): line for line in lines}
 
         for line_id, line_info in lines_by_id.iteritems():
```
lmignon: @lmarion-source You must avoid to search the lines and orders line by line....
```
lines = self.env["purchase.order.line"].browse(lines_by_id.keys())
```

despatch_advice_import/wizard/despatch_advice_import.py
```diff
@@ -136,16 +136,22 @@ def process_document(self):
...
-            line = order.order_line.search([("id", "=", line_id)])
+            line = self.env["purchase.order.line"].search([("id", "=", line_id)])
+
+            # Check PO exist
+            order = self.env["purchase.order"].search([("name", "=", line_info["ref"])])
```
lmignon: The order is available on the line `order = line.order_id`. Yous should only check:
 * if the order ref is provided, the line.order_id.ref must == order_ref
 * if no order ref provided: line.order_id.name == po_name

lmignon (changes requested):

despatch_advice_import_ubl/README.rst
```diff
@@ -0,0 +1,91 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+==========================
+Despatch Advice Import Ubl
+==========================
+
+Import Despatch Advice files
```
bealdav, line 9: IMHO it seems there are 2 workflows with despatch:

1. yours: your receive file about goods that'll receive soon
2. other flow: you define a delivery order for your customer (picking out is created at sale confirmation) and your warehouse site/manager send you a file to indicate which lots are really sent, then have to match with an existing picking

Probably interesting to define here that you only implements the first one. Thanks

bealdav: I have to implement last case in v12, for information

bealdav: Also a despatch, couldn't be linked to an order (sale or purchase) even if UBL require it.
In that case, we only deal with picking, for example between warehouses

bealdav: Please @lmignon @lmarion-source are you OK with the different possible use cases ? Sure I don't want you implement them here. Just I would know if I'm not in error functionally speaking. Thanks

PR #283, opened 2020-12-09
[FIX] edi: Rule not correctly created

etobella:
Rule not correctly formulated

@simahawk Please fast-track

### DIFF
--- a/edi/security/ir_model_access.xml
+++ b/edi/security/ir_model_access.xml
@@ -83,7 +83,7 @@
     <record id="rule_edi_exchange_record_manager" model="ir.rule">
         <field name="name">Manager EDI exchange records</field>
         <field name="model_id" ref="edi.model_edi_exchange_record" />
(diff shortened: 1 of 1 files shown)


simahawk (approved):

edi/security/ir_model_access.xml
```diff
@@ -83,7 +83,7 @@
     <record id="rule_edi_exchange_record_manager" model="ir.rule">
         <field name="name">Manager EDI exchange records</field>
         <field name="model_id" ref="edi.model_edi_exchange_record" />
-        <field name="domain_force">[('1', '=', '1')]</field>
+        <field name="domain_force">[(1, '=', 1)]</field>
```
simahawk, line 86: it should work anyway no? The point is that the domain should always be TRUE.

etobella: As it is an string, it tries to use the field names '1', and it obiously does not exist.

jarroyomorales (approved):

PR #284, opened 2020-12-09
[8.0][ADD] account_invoice_ubl: DueDate

thomaspaulb:
Add the payment due date to UBL invoice.

[PEPPOL BR-CO-25](https://docs.peppol.eu/poacc/billing/3.0/rules/BR-CO-25) specifies that if the invoice amount is positive, either the due date or the payment terms are filled, so this is pretty much mandatory.

### DIFF
--- a/account_invoice_ubl/models/account_invoice.py
+++ b/account_invoice_ubl/models/account_invoice.py
@@ -24,6 +24,9 @@ class AccountInvoice(models.Model):
         doc_id.text = self.number
         issue_date = etree.SubElement(parent_node, ns['cbc'] + 'IssueDate')
         issue_date.text = self.date_invoice
+        if version >= '2.1':
+            due_date = etree.SubElement(parent_node, ns['cbc'] + 'DueDate')
+            due_date.text = self.date_due
         type_code = etree.SubElement(
             parent_node, ns['cbc'] + 'InvoiceTypeCode')
         if self.type == 'out_invoice':


thomaspaulb:
@etobella @StefanRijnhart @gjotten While we're at PEPPOL, maybe this one as well?

StefanRijnhart:
Does it make sense to check the sign of the invoice amount (c.q. invoice type), or are you positive this is harmless in the case of credit notes?

gjotten:
@StefanRijnhart since date_due is always getting filled AFAICS, can we think of any situation where we don't want to include DueDate in the .xml anyway? I wouldn't mind adding it by default, PEPPOL says nothing about there being too much data in this case, otherwise the below points apply:

Out_invoice with a negative total will parse as a negative amount in the UBL .xml, so yes, we should exclude amount_total <= 0 then per _strict_ interpretation of PEPPOL. 

Out_refund nets us a _positive_ amount in the .xml so we'd need to control for that as well @thomaspaulb

StefanRijnhart (approved):
Again, I am not familiar with these standards so if mentioning the due date for negative invoices is harmless this is looking good to me.

gjotten (approved):
+1

thomaspaulb:
We double-checked by putting a credit invoice with DueDate through the validator on https://peppol-test.nl/, and there was no error. (At least not related to this - there's lots more to do regarding PEPPOL-UBL in Odoo, but lets' leave that to future PR's hehe)

@hbrunn Can you do the PSC-honors.

StefanRijnhart:
Perfect, thanks!

I hear @hbrunn is busy at the moment so I'll relieve him in this matter.

/ocabot merge patch

PR #285, opened 2020-12-09
[8.0][IMP] base_ubl_payment: Setting for default payment mode

thomaspaulb:
In some countries (Netherlands) the current default of 31 is invalid, so an option is given to change it.

### DIFF
--- a/base_ubl_payment/README.rst
+++ b/base_ubl_payment/README.rst
@@ -11,7 +11,7 @@ This module is a small module above the *base_ubl* module; it adds the generatio
 Configuration
 =============
 
-No configuration is needed.
+By default, if no payment mode is set on an invoice, the UNECE code 31 is passed as PaymentMeans/PaymentMeansCode. If you want to override this, select a default payment mode on the UBL tab of the Company form view and set the UNECE code of its payment.mode.type to something else.
 
 Usage
 =====
--- a/base_ubl_payment/__openerp__.py
+++ b/base_ubl_payment/__openerp__.py
@@ -13,6 +13,9 @@
     'depends': [
         'account_payment_unece',
         'base_ubl',
-        ],
+    ],
+    'data': [
+        'views/company.xml',
+    ],
     'installable': True,
 }
--- a/base_ubl_payment/models/__init__.py
+++ b/base_ubl_payment/models/__init__.py
@@ -1,3 +1,4 @@
 # -*- coding: utf-8 -*-
 
+from . import company
 from . import ubl
--- a/base_ubl_payment/models/company.py
+++ b/base_ubl_payment/models/company.py
@@ -1,0 +1,14 @@
+# -*- coding: utf-8 -*-
+# © 2016 Akretion (Alexis de Lattre <alexis.delattre@akretion.com>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from openerp import models, fields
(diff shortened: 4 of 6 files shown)


thomaspaulb:
@astirpe For the final goal of having NLCIUS-consistent UBL's coming from Odoo 13.0, I'd need to get a couple of these 8.0 changes reviewed, then forward-port them to 13.0 as a whole bunch when they're deemed acceptable. Can I bother you with them? If someone with another country's perspective then steps in, he could make the final call.

StefanRijnhart:
Looking at UBL specs for the first time. Do I understand correctly that NLCIUS recommends 30 (credit transfer) where 31 (debit transfer) is taken as the default in these OCA modules? Aren't these complete opposites?

thomaspaulb:
@gjotten can you comment on this

gjotten:
@StefanRijnhart & @thomaspaulb AFAICS the generic 30 and 31 payment means codes correspond with the new(er) additions '58 SEPA credit transfer' and '59 SEPA direct debit'. 

In other words, what we would consider an 'overboeking' vs. an 'incasso' in Dutch. But, 31 is neither automatic, direct, nor periodic and I believe it is on its way out or already unsupported in SEPA countries anyway.

Now NLCIUS supports 30, 58, and 59 as well as: 

- 48 (card payment)
- 49 (non-SEPA direct debit)
- 57 (standing agreement) 

I think we can safely say 48 is probably a PoS situation in the vast majority of cases, 49 is not relevant in SEPA countries, and 57 can be relevant but is kind of an outlier and best supported through a payment mode.

As for choosing a default, 30 as a generic option should be fine, but my choice would be 58.

gjotten:
OTOH, since Odoo is used in non-SEPA countries as well I guess 30 would be the most logical default anyway.

StefanRijnhart:
Thanks! I'm still slightly puzzled why the current default is a debit mode while you state that a credit mode makes more sense as a default (and support for direct debit in the current code is lacking (e.g. no mentioning of the mandate number)) but again, I'm very much overwhelmed reading bits and pieces of these specifications for the first time.

About covering the situation in which the payment mode is missing, I believe the vision of the original author is to ensure that the mode is present as much as possible, having provided modules us with modules such as 'account_payment_partner' and 'account_payment_sale'. Is there not a case to be made to deploy such modules rather than adding a global default payment mode *just for the UBL export of the invoice*?

gjotten:
@StefanRijnhart without payment_mode_id being a required field on customer invoices, something has to set _a_ PaymentMeansCode in the .xml or UBL Business Rule 49 nets you a fatal error (i.e. the vendor's accounting software will most likely reject Odoo's invoice).

Account_payment_partner nets you nothing more than the chance to specify what it needs to be since payment mode is optional. So with the way it's set up now, a default is needed if you don't want users easily creating broken UBL data.

StefanRijnhart:
OK. I'll abstain from further review because I don't have the required experience and my intuitions do not align with this change.

thomaspaulb:
@StefanRijnhart Would your intuitions align better with a change that:

- Adds a related boolean field on invoice to check whether the invoice partner requires Dutch UBL (this field is computed on partner, by default when country = NL but you can override it)
- Uses this boolean to determine required=True or False for payment mode on invoice
- Adds a warning when you select a payment mode that has no UNECE code defined

PR #286, opened 2020-12-10
[ADD] edi_webservice

etobella:
Defines webservice integration from EDI Exchange records

@simahawk


pedrobaeza:
Shouldn't it be a bit more specific (SOAP, JSON, ReST...)?

etobella:
You need to implement each case, as a small demo, I made POST and PUT for REST

pedrobaeza:
Yeah, but I suppose you will need extra module with library dependencies, and that's why I wonder if this abstraction layer serves or it can be directly put on each specific one.

etobella:
Well, this is a PoC yet, usually all WS (REST, SOAP, AXIS...) send some data (the file in this case) to an Endpoint. So, we need to create a method for each case (depending on the library). However, once we have created the component, we don't need to create the code anymore. For example, all the POST and PUT methods should be solved with specific configuration using this base module. For SOAP we could create edi_webservice_soap that implements the soap component and everything could be solved.

Also this module is made so we can divide it in `backend_webservice` and `edi_webservice`. Similar to `storage` and `edi_storage`.

Even, sometimes, we will be unable to use this as their specifications are weird and need to be solved differently (like spanish FACe...)

pedrobaeza:
OK, I just don't want to have an over-engineered framework with lots of modules with few real code.

simahawk:
I agree w/ Pedro on "Shouldn't it be a bit more specific (SOAP, JSON, ReST...)?".
What I had in mind when we talked a bit about this was something like:

- have base modules for webservices (maybe to be put into a new repo OCA/webservice)
- there we should do something similar to OCA/storage whereas the type of the backend defines the type of the protocol (eg: XMLRPC, HTTP, etc) and adapters are looked up by that value
- adapters have to expose the basic api for those protocols (eg: put/get/delete/etc)
- specific implementations, like this one, should define their own adapters only if needed
- an edi_webservice should only take care of relating an edi_backend w/ a ws backend and call ws method using the right params (similar to what I did for edi_storage)

@etobella we can have a call to discuss details if you want :wink:

pedrobaeza:
About the repo, I think this one should host everything though, unless more abstraction from EDI can be provided, but that abstraction can be overkilling.

simahawk:
> About the repo, I think this one should host everything though, unless more abstraction from EDI can be provided, but that abstraction can be overkilling.

yeah, for starting - at least - keep it in a separated module

etobella:
OK, I will split it

etobella:
It has been splitted in two modules:

- `webservice`: Adds a generic webservice configuration and a first case for the library requests
- `edi_webservice`: Link between `webservice` and `edi`, not much code there

webservice/components/webservice_component.py
```diff
@@ -0,0 +1,71 @@
...
+            cookies=cookies,
+            verify=verify,
+            headers=self._get_headers(**kwargs),
+            timeout=timeout,
+        )
+        request.raise_for_status()
+        return request.content
+
+    def post(self, data, timeout=20, verify=None, cookies=None, **kwargs):
+        request = requests.post(
```
simahawk: just use `request.request` for all kind of requests and move the logic to `_make_request` method. See https://requests.readthedocs.io/en/master/api/#main-interface

webservice/tests/test_webservice.py
```diff
@@ -0,0 +1,130 @@
+# Copyright 2020 Creu Blanca
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+import mock
+from requests import exceptions
+
+from odoo.tests.common import tagged
+
+from odoo.addons.component.tests.common import SavepointComponentRegistryCase
```
simahawk: you don't need this class if you are not registering other components. Use SavepointComponentCase

webservice/tests/test_webservice.py
```diff
@@ -0,0 +1,130 @@
...
+
+
+class DemoResponse:
```
simahawk: use https://pypi.org/project/responses/ instead

edi_webservice/models/edi_backend.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+class EdiBackend(models.Model):
+
+    _inherit = "edi.backend"
+
+    webservice_backend_id = fields.Many2one("webservice.backend")
+
+    def _get_component_usage_candidates(self, exchange_record, key):
+        candidates = super()._get_component_usage_candidates(exchange_record, key)
```
simahawk, line 15: am I wrong or this would be required only for `send` or `receive` actions?

simahawk (changes requested):

etobella:
Still pending the first comment

webservice/components/webservice_component.py
etobella: I made some changes. I hope it fills your expectations :wink:

webservice/tests/test_webservice.py
```diff
@@ -0,0 +1,131 @@
...
+
+import responses
+from requests import auth, exceptions
+
+from odoo.tests.common import tagged
+
+from odoo.addons.component.tests.common import SavepointComponentCase
+
+
+class DemoResponse:
```
simahawk: this is not needed anymore

edi_webservice/models/edi_backend.py
```diff
@@ -0,0 +1,26 @@
...
+from odoo import fields, models
+
+
+class EdiBackend(models.Model):
+
+    _inherit = "edi.backend"
+
+    webservice_backend_id = fields.Many2one("webservice.backend")
+
+    def _get_component_usage_candidates(self, exchange_record, key):
```
simahawk, line 14: can we think about how this can be simplified by using this new approach? https://github.com/OCA/edi/pull/295

Maybe we can extend the component match to `_webservice_protocol`. No?

If this works for you I'll make that piece a bit more flexible.

etobella: You are right, but maybe I should need to wait for the approval of the commit, isn't it? We can keep this PR as DRAFT until the merge, but it seems right to me :wink:

etobella:
Hi @simahawk We might need this for OCA/l10n-spain#1590

simahawk:
> Hi @simahawk We might need this for [OCA/l10n-spain#1590](https://github.com/OCA/l10n-spain/pull/1590)

@etobella ciao! Sorry, 1st holidays, then the EDI project got frozen and I left the whole thing behind :sweat_smile: 
I make a tour now!

webservice/components/webservice_component.py
```diff
@@ -0,0 +1,52 @@
...
+
+class BaseWebServiceAdapter(AbstractComponent):
+    _name = "base.webservice.adapter"
+    _collection = "webservice.backend"
+
+
+class BaseRestRequestsAdapter(Component):
+    _name = "base.requests"
+    _usage = "webservice.httprequest"
+    _inherit = "base.webservice.adapter"
```
simahawk: so, by looking at https://github.com/OCA/edi/pull/295 you should be able to define this like

```suggestion
class BaseRestRequestsAdapter(Component):
    _name = "base.requests"
    _usage = "webservice.request"
    _webservice_protocol = "http"
    _inherit = "base.webservice.adapter"
```

webservice/components/webservice_component.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def request(self, method, **kwargs):
```
simahawk:
```suggestion
    def _request(self, method, **kwargs):
```

edi_webservice/models/edi_backend.py
simahawk: you are right. Go ahead if this unlocks your work but add a TODO for refactoring.
Look at my comment on the adapter class.

simahawk (changes requested):

etobella:
@simahawk Changes made to have the same logic.

webservice/__manifest__.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2020 Creu Blanca
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Edi WebService",
```
simahawk: "edi" to remove from this module?

webservice/components/webservice_component.py
```diff
@@ -0,0 +1,62 @@
+# Copyright 2020 Creu Blanca
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+import requests
+
+from odoo.addons.component.core import AbstractComponent, Component
+
+
+class BaseWebServiceAdapter(AbstractComponent):
```
simahawk: can you split this 2 classes in 2 files? (`base_adapter.py` and `request_adapter.py` shall be good)

webservice/models/webservice_backend.py
```diff
@@ -0,0 +1,50 @@
...
+
+from odoo import fields, models
+
+
+class WebserviceBackend(models.Model):
+
+    _name = "webservice.backend"
+    _inherit = ["collection.base", "server.env.mixin"]
+    _description = "WebService Backend"
+    _server_env_section_name_field = "code"
```
simahawk: you can use https://github.com/OCA/server-env/blob/13.0/server_environment/models/server_env_tech_name_mixin.py

webservice/tests/test_webservice.py
```diff
@@ -0,0 +1,120 @@
...
+import responses
+from requests import auth, exceptions
+
+from odoo.tests.common import tagged
+
+from odoo.addons.component.tests.common import SavepointComponentCase
+
+
+@tagged("-at_install", "post_install")
+class CommonWebService(SavepointComponentCase):
```
simahawk: move it to `common.py`

webservice/tests/test_webservice.py
```diff
@@ -0,0 +1,120 @@
...
+                "code": "demo_ws",
+            }
+        )
+
+    def test_web_service_not_found(self):
+        with self.assertRaises(exceptions.ConnectionError):
+            self.webservice.call("get")
+
+    @responses.activate
+    def test_web_service_get_mocked(self):
```
simahawk: remove `_mocked` from every test

webservice/views/webservice_backend.xml
```diff
@@ -0,0 +1,77 @@
...
+                    </div>
+                    <group name="config">
+                        <field name="protocol" />
+                        <field name="url" />
+                        <field name="username" />
+                        <field name="password" password="True" />
+                        <field name="content_type" />
+                    </group>
+                </sheet>
+                <div class="oe_chatter" />
```
simahawk:
```suggestion

```

edi_webservice/models/edi_backend.py
```diff
@@ -0,0 +1,35 @@
...
+    _webservice_actions = ("send", "receive")
+
+    def _get_component_usage_candidates(self, exchange_record, key):
+        candidates = super()._get_component_usage_candidates(exchange_record, key)
+        if not self.webservice_backend_id or key not in self._webservice_actions:
+            return candidates
+        return ["webservice.{}".format(key)] + candidates
+
+    def _component_match_attrs(self, exchange_record, key):
+        # Override to inject storage_backend_type
```
simahawk:
```suggestion
        # Override to inject `webservice_protocol` as match attribute
```
and replace all occurrences :wink:

webservice/__manifest__.py
```diff
@@ -0,0 +1,15 @@
...
+    "summary": """
+        Defines webservice integration from EDI Exchange records""",
+    "version": "13.0.1.0.0",
```
simahawk, line 8: alpha?

simahawk (changes requested):
Thanks! Much better! :)

etobella:
Thanks for the review, comments attended

simahawk (approved):

PR #287, opened 2020-12-10
[8.0][MIG] Backport of account_invoice_ubl_email_attachment from 11.0

thomaspaulb:


thomaspaulb:
@astirpe What do you think about the last commit? In 8.0 but I think also for other versions of the UBL module, every time you press the "Generate XML" button a new attachment is generated. I don't see why we should send all of them to the customer.

astirpe (approved):
Looks good :+1: 



> @astirpe What do you think about the last commit? In 8.0 but I think also for other versions of the UBL module...

I agree: this fix should be also ported to other versions.

PR #291, opened 2020-12-11
[13.0][FIX] account_invoice_ubl_email_attachment: attach only the most recently generated UBL file

thomaspaulb:
Forward port of https://github.com/OCA/edi/pull/287

### DIFF
--- a/account_invoice_ubl_email_attachment/models/mail_template.py
+++ b/account_invoice_ubl_email_attachment/models/mail_template.py
@@ -21,24 +21,26 @@ class MailTemplate(models.Model):
             invoice = self.env["account.move"].browse(res_id)
             version = invoice.get_ubl_version()
             ubl_filename = invoice.get_ubl_filename(version=version)
-            ubl_attachments = self.env["ir.attachment"].search(
+            ubl_attachment = self.env["ir.attachment"].search(
                 [
                     ("res_model", "=", "account.move"),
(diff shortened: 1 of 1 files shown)


astirpe (approved):

StefanRijnhart (approved):

account_invoice_ubl_email_attachment/models/mail_template.py
```diff
@@ -26,7 +26,9 @@ def generate_email(self, res_ids, fields=None):
                     ("res_model", "=", "account.move"),
                     ("res_id", "=", res_id),
                     ("name", "=", ubl_filename),
-                ]
+                ],
+                order="create_date desc",
+                limit=1,
```
etobella, line 31: If you are adding the limit, then the `if len(ubl_attachments) == 1` has no sense.

Can you explain why the limit=1 is important?

thomaspaulb: See https://github.com/OCA/edi/pull/287, if we dont have the limit then all the previous versions of the UBL will also be attached, which makes no sense. I agree the code can be shortened though, will push a fixup.

etobella:
@thomaspaulb can you squash your commits and then proceed with merge :wink:

thomaspaulb:
@etobella Done.

PR #292, opened 2020-12-13
EDI: Some fixes

etobella:
This PR adds two changes:

- Fixes the fields_view_get, as it was loosing the specific views on some fields (for example, on invoice lines)
- Adds a new function to search a exchange record (similar to _has_exchange_record, but returning the record)

### DIFF
--- a/edi/models/edi_exchange_consumer_mixin.py
+++ b/edi/models/edi_exchange_consumer_mixin.py
@@ -138,7 +138,19 @@ class EDIExchangeConsumerMixin(models.AbstractModel):
         return action
 
     def _has_exchange_record(self, exchange_type, backend=False, extra_domain=False):
-        """This function is useful when generating the configuration"""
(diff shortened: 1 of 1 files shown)


edi/models/edi_exchange_consumer_mixin.py
```diff
@@ -139,6 +142,17 @@ def edi_create_exchange_record(self, exchange_type_id):
 
     def _has_exchange_record(self, exchange_type, backend=False, extra_domain=False):
         """This function is useful when generating the configuration"""
```
simahawk: can you change the docstrings as you are here? :wink:

simahawk (approved):
LG, small remark

edi/models/edi_exchange_consumer_mixin.py

jarroyomorales (approved):
LGTM :+1:

PR #294, opened 2020-12-16
8.0 FIX crash when pivot format had a 'note' key (and no attachment) backport from v10

bealdav:
Add method to create SO in sale.order.import accessible via JSON-RPC


thomaspaulb (approved):
Identical to the [10.0](https://github.com/OCA/edi/commit/730ef0dc939db5129df4129f61ebbf602a566eee) commit and not seeing any issues.

PR #295, opened 2020-12-16
[13.0] edi: refactor component lookup

simahawk:
Depends on https://github.com/OCA/connector/pull/392

The match is now done based on specific control keys: backend type, exchange type.
This makes components more clean and their behavior more understandable.

@etobella this is what I was talking about last time :wink:

### DIFF
--- a/edi/components/base.py
+++ b/edi/components/base.py
@@ -9,6 +9,9 @@ class EDIBackendComponentMixin(AbstractComponent):
 
     _name = "edi.component.mixin"
     _collection = "edi.backend"
+    _usage = None
+    _backend_type = None
+    _exchange_type = None
 
     @property
     def exchange_record(self):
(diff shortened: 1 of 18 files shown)


simahawk:
@etobella can you review this and https://github.com/OCA/connector/pull/392 ?

etobella (approved):
The only thing I don't like is that old implementations will not work anymore with this changes. I have implemented one module (controlled, so no real problem) and I don't know if it will affect existent PRs

simahawk:
> The only thing I don't like is that old implementations will not work anymore with this changes. I have implemented one module (controlled, so no real problem) and I don't know if it will affect existent PRs

well, if that's the case: override the usage candidates method and add them back where you need them.
If you have several implementation I can add backward support and log the deprecation.

etobella:
Ok! @simahawk as soon as it is green, IMO we can proceed with merge :wink:

alexis-via (approved):

alexis-via:
I restarted Travis, let's see if it becomes green now that the other PR has been merged.

etobella:
@simahawk it seems related to the exchange_template

etobella:
I made a PR on your branch fixing the test. I don't like the approach, but I didn't know a better option. Can you check it please?

etobella:
@simahawk

simahawk:
@etobella fixed in last commit in the right way :wink: 
Thanks for the reminder! :stuck_out_tongue:

etobella:
I knew you will do it in the right way :stuck_out_tongue_closed_eyes: but this way you had an idea why the error was raised (And I was obviously saying it was a wrong approach :laughing: )

simahawk:
> I knew you will do it in the right way stuck_out_tongue_closed_eyes but this way you had an idea why the error was raised (And I was obviously saying it was a wrong approach laughing )

jeje! Much appreciated :wink: Muchas gracias!

PR #297, opened 2020-12-22
8.0 IMP base_ubl: add UUID node

bealdav:
UUID  can be used to match line with Order Response

### DIFF
--- a/purchase_order_ubl/models/purchase.py
+++ b/purchase_order_ubl/models/purchase.py
@@ -69,7 +69,8 @@ class PurchaseOrder(models.Model):
         self._ubl_add_line_item(
             line_number, oline.name, oline.product_id, 'purchase',
             oline.product_qty, oline.product_uom, line_root, ns,
-            seller=self.partner_id.commercial_partner_id, version=version)
+            seller=self.partner_id.commercial_partner_id, uuid=oline.id,
+            version=version)
 
     @api.multi
     def _ubl_add_order_line(
@@ -85,7 +86,7 @@ class PurchaseOrder(models.Model):
             seller=self.partner_id.commercial_partner_id,
             currency=self.currency_id, price_subtotal=oline.price_subtotal,
             qty_precision=qty_precision, price_precision=price_precision,
-            version=version)
+            uuid=oline.id, version=version)
 
     @api.multi
     def get_delivery_partner(self):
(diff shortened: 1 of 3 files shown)


purchase_order_ubl/models/purchase.py
```diff
@@ -85,7 +86,7 @@ def _ubl_add_order_line(
             seller=self.partner_id.commercial_partner_id,
             currency=self.currency_id, price_subtotal=oline.price_subtotal,
             qty_precision=qty_precision, price_precision=price_precision,
-            version=version)
+            uuid=oline.id, version=version)
```
StefanRijnhart, line 89: Isn't it unwise to use a numeric id that is not 'universally unique'?

bealdav: Thanks for your review @StefanRijnhart 

Selection_137

Semantically you're right, but contextually to the exchange from a system A to a system B for the flow, purchase line item here, the UID is unique because of unique id in the table.

You're right it's more a UID than UUID. It's a compromise.
Do you think there is more convenient attribute ?

http://www.datypic.com/sc/ubl22/e-cac_LineItem.html

StefanRijnhart: If there is already an `identifier for this line item asigned by the seller` then that's where the line id should go, not into the UUID.

I can imagine applying a mix-in class to add an automatically populated UUID column to Odoo models. Isn't there already a module that provides that?

bealdav: > If there is already an identifier for this line

I carefully check all other attributes on line item and there is no other matching the case. There nearest is uuid.
I propose to add an overridable method which returns this UUID with default value as False, what do you think ?
If someone need for its usage the proposed mixin mechanism, it can be implemented with this method but in this case it should be decoupled of these EDI module (not depends on it).

StefanRijnhart: I do not know my way around these specifications. But in the screenshot above, where it says 'SalesOrderID', the comment mentions "the identifier for the *line* item". I see references to `cbc:SalesOrderLineID` if I google around. Is there an error in the specs above?

bealdav: Yes that's quite confusing because we know both that SalesOrderID is very different from SalesOrderLineID.
Specifications are not perfect and our projects specially in v8 are not perfect too ;-)

StefanRijnhart:
I would very much like to have clarity on the specs, but if this cannot be had, you could consider creating a UUID from the MD5 sum of a representation of the line id that is unique in the context of your database.
```
>>> import uuid
>>> print uuid.uuid3(uuid.NAMESPACE_DNS, 'sale.order.line#123')
df5bf1b3-cb87-351d-b21a-8ff51fe06c08
```
(`123` here being the internal ID of the line in your database)

bealdav:
Thanks to your review @StefanRijnhart and explanations
Finally a more convenient PR https://github.com/OCA/edi/pull/311 for the use case

Issue #246, opened 2020-11-05
RFC 14.0 migration, call for ideas


bealdav:
Here is some a collection of ideas to refactor this repo to next version

Feel free to add you own ideas

pedrobaeza:
Proposal A: Embrace #243 as base for everything

bealdav:
## base_ubl mixin

https://github.com/OCA/edi/blob/13.0/base_ubl/models/ubl.py

This mixin contains a lot of methods which could be separated in several mixins.

example here https://github.com/OCA/edi/blob/13.0/base_ubl/models/ubl.py#L294
This method is related to line models (sale.order.line and purchase.order.line) but is attached to sale or purchase. Then it's difficult to override lineItem generation depending of your order.line. Possible but it could be made better.

Proposal B : split in different mixins matching real models

@pedrobaeza you can rename your proposal to A

simahawk:
> base_ubl mixin
> This mixin contains a lot of methods which could be separated in several mixins.

If you want to keep python code it would be probably better to use components (convention to define).

In general, for all modules generating outputs we could rework them  - partially or completely - on top of https://github.com/OCA/edi/pull/259. I will soon add support for GS1 which outputs XML. That will give you a real world example.

etobella:
We have also integrated spanish e-invoice with `edi`.  I think that it would be great to integrate all modules with this base. WDYT?
https://github.com/OCA/l10n-spain/pull/1519
@OCA/edi-maintainers

nilshamerlinck:
Hi @OCA/edi-maintainers 

This components-based edi-pocalypse effort looks great ;-)

However, I'm starting to think that it could make sense to introduce a new repo to handle the transition more smoothly: `OCA/edi-legacy`, that would host the "old" modules, eventually migrated to 14.0 without refactoring, for users that would prefer to stick  to the old approach.

What do you think?

simahawk:
> Hi @OCA/edi-maintainers
> 
> This components-based edi-pocalypse effort looks great ;-)
> 
> However, I'm starting to think that it could make sense to introduce a new repo to handle the transition more smoothly: `OCA/edi-legacy`, that would host the "old" modules, eventually migrated to 14.0 without refactoring, for users that would prefer to stick to the old approach.
> 
> What do you think?

[...]

PR #301, opened 2021-01-02
[MIG] base_business_document_import: Migration to 14.0

phuctranfxvn:
Depends module:
- `account_payment_unece` migrated in https://github.com/OCA/community-data-files/pull/100

* Note: I included the folk of the `account_payment_unece`'s repo because the original PR includes no setup file for this module.


etobella:
Just a comment, maybe now is the moment to integrate it with `edi` module from 13.0. WDYT? We have worked with @simahawk with this modules and we think they are a good approach for `edi`, so it would be interesting to make an approach of all `edi` modules in order to simplify the environment.

pedrobaeza:
I agree on unifying everything under `edi` family of modules.

pedrobaeza:
Do you think it's worth to keep this module instead of having main base on `edi` module and directly inherit from it from `account_invoice_import` and `sale_order_import`? Other option if keeping the module is to call this new one `edi_import` for not mixing things.

etobella:
Well, the main functions could be pushed directly to `edi`, but I think it is better to make a new module just for this. But I prefer the name `edi_import`.

alexis-via:
FYI, I'm updating this module to take into account the datamodel change in account_tax_unece and other small updates (self.env.company instead of self.env.user.company_id, etc....).
The discussion about dependency on edi should be continued in https://github.com/OCA/edi/issues/246 (I gave my point of view there).

base_business_document_import/models/business_document_import.py
```diff
@@ -720,7 +720,9 @@ def _match_tax(
                 ("unece_due_date_code", "=", tax_dict["unece_due_date_code"]),
                 ("unece_due_date_code", "=", False),
             ]
-        taxes = ato.search(domain, order="unece_due_date_code")
+        # taxes = ato.search(domain, order="unece_due_date_code")
```
fshah-initos: Please remove commented code.

phuctranfxvn: removed, thanks

rconjour:
PR works techinically correct, used it also for my own migration of purchase-order-import. Code also looks great!

rconjour (approved):

fshah-initos:
@phuctranerp please squash the last commit.

phuctranfxvn:
> @phuctranerp please squash the last commit.

done, thanks

rconjour:
@fshah-initos can you find some time to review this PR again?

base_business_document_import/models/business_document_import.py
```diff
@@ -720,7 +720,8 @@ def _match_tax(
                 ("unece_due_date_code", "=", tax_dict["unece_due_date_code"]),
                 ("unece_due_date_code", "=", False),
             ]
-        taxes = ato.search(domain, order="unece_due_date_code")
+        taxes = ato.search(domain)
```
fshah-initos, line 723: Why the order attribute is removed?

fshah-initos:
PLease resolve the conflicts.

base_business_document_import/models/business_document_import.py
phuctranfxvn: @fshah-initos because the records will be sorted with _get_unece_due_date_type_code, see the next line

phuctranfxvn:
> PLease resolve the conflicts.

done

bosd:
Which one of the 2 migration PR's can be merged?

StefanRijnhart (approved):

StefanRijnhart:
Let's merge this one first as it's purely the migration.

/ocabot merge nobump

PR #318, opened 2021-01-11
[12.0] account_invoice_facturx: filter out invoice lines with display_type

alexis-via:
Catch exception when xsd validation fails and display a nice error message


etobella (approved):
Code Review LGTM

marcelsavegnago (approved):

PR #323, opened 2021-02-11
[13.0][ADD] account_invoice_ubl_peppol

thomaspaulb:
References:

[Example PEPPOL invoice](https://github.com/OpenPEPPOL/peppol-bis-invoice-3/blob/master/rules/examples/base-example.xml)

[Validator](https://peppol-test.nl/validate)

Todo:

- [x] ~~Unit test, using PEPPOL XML or XSD definitions~~ ROADMAP item, describing better unit test
- [x] ~~Aside from using DueDate, fill PaymentTerms with the combined due-date/terms field that Odoo 13.0 now has~~ Added to ROADMAP as well
- [x] ~~Test with credit and purchase invoices~~ Added to ROADMAP


account_invoice_ubl_peppol/models/account_move.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+class AccountMove(models.Model):
+    _inherit = "account.move"
+
+    def _account_invoice_ubl_use_peppol(self):
+        """ Returns True when we should use PEPPOL """
+        domain = self.company_id.get_ubl_domain_peppol()
+        domain.append(('id', '=', self.id))
+        if self.search(domain):
```
etobella: You can use filtered_domain for this
https://github.com/odoo/odoo/blob/13.0/odoo/models.py#L5346

account_invoice_ubl_peppol/README.rst
```diff
@@ -0,0 +1,105 @@
...
+
+- Go to menu *Invoicing > Configuration > Settings > Invoicing*, under *Electronic Invoices*.
+
+- Formulate a domain for which invoices the dialect should become PEPPOL.
+  By default it is *[]*, so all UBL invoices will be PEPPOL.
+  If you want this only for Belgian partners, then you can fill here for example:
+  `[('partner_id.country_id.code', '=', 'BE')]`
+  Or you can choose to enable this only for selected partner ids.
+
+- Install OCA module 'partner_coc' and fill the field `coc_registration_number`
```
etobella: If this is required, maybe you can add it in the manifest on the `depends`

etobella (commented):
Two base commennts. Also review pre-commit please

account_invoice_ubl_peppol/README.rst
thomaspaulb: I'm reluctant to make this required - installations that use another field to store the COC or any other kind of PEPPOL address identifier such as the "organization number", should be able to use this module without having to install `partner_coc`.

On the other hand, if those installations dont need the partner_coc fields they can also use a custom module to hide the partner_coc fields to avoid confusion.

What do you think?

etobella: I would prefer to add it on the depends, and special cases should make the changes (they would need to do it anyway)

account_invoice_ubl_peppol/models/account_move.py
thomaspaulb: Thanks! I didn't know about that.

thomaspaulb:
@etobella This is now ready for a more thorough review.

gjotten:
@thomaspaulb my test-invoice is triggering fatal errors on BR-CL-17 and BR-CL-18 per https://peppol-test.nl/validate, complaining about "Invoice tax categories MUST be coded using UNCL5305 code list". Which is funny because it _does_ contain 

```
      <cac:TaxCategory>
        <cbc:ID>H</cbc:ID>
```
Which to me at least would imply a bog standard [H]igher category of VAT being levied. Any idea why this error is popping up?

thomaspaulb:
@gjotten PEPPOL does not allow `H`: https://docs.peppol.eu/poacc/billing/3.0/codelist/UNCL5305/

gjotten:
If that's the case, [then.](https://github.com/OCA/l10n-netherlands/blob/13.0/l10n_nl_account_tax_unece/data/account_tax_template.xml#L9) [we.](https://github.com/OCA/l10n-netherlands/blob/13.0/l10n_nl_account_tax_unece/data/account_tax_template.xml#L17) [have.](https://github.com/OCA/l10n-netherlands/blob/13.0/l10n_nl_account_tax_unece/data/account_tax_template.xml#L21) [problems.](https://github.com/OCA/l10n-netherlands/blob/13.0/l10n_nl_account_tax_unece/data/account_tax_template.xml#L67) in l10n_nl_account_tax_unece for example, no? 

Does *a* specification imply that all 'standard' rates need to be put under TaxCategory ID [S] then? If so, that implies either an extension of this module tackling the above problems for l10n-netherlands, or a change in that particular dataset. 

[l10n_fr_account_tax_unece](https://github.com/OCA/l10n-france/blob/13.0/l10n_fr_account_tax_unece/data/account_tax_template.xml) seems to adher to the [PEPPOL business rule](https://docs.peppol.eu/poacc/billing/3.0/rules/ubl-tc434/BR-CL-17/) so I guess we're dealing with the latter. Maybe @astirpe can comment?

thomaspaulb:
I'd say those are wrong yes, also in [NLCIUS](https://www.softwarepakketten.nl/wiki_uitleg/10&bronw=7/UBL_en_BTW_algemene_informatie.htm)

alexis-via:
Thanks for this PR, PEPPOL is a very interesting topic and we need to move forward on it.
My question: do you think that it would be reasonable to have peppol compliance directly in account_invoice_ubl ? I mean account_invoice_ubl would generate a PEPPOL-compliant UBL, and then other modules can extend if they want additionnal fields that are not allowed/required by PEEPOL. This could be an option for v14, as account_invoice_ubl hasn't been ported to v14 yet, but we need to decide this rapidly.

I haven't read the PEEPOL specs yet (and I don't have time for that in the very short term). When reading the comments in your PR, I'm surprised by comments that say that some XML tags are not allowed (like WebsiteURI). CIUS (Common Invoice Usage Specifications) usually specify tags that are required, not tags that are not allowed! I've never seen a CIUS that disallow the use of some XML tags that are present in the underlying standard.

thomaspaulb:
@alexis-via If you want a very quick and dumb overview of the differences between what `account_invoice_ubl` generates, and what PEPPOL requires, just generate an invoice from the Runbot or from your customer's server, and paste it in the Validator linked at the very top of the PR. That will take you no more than 15 minutes and you see the errors and warnings. Or if that is still too much, I can paste the image for you here.

`WebsiteURI` was not an error but a Warning, but I decided to go clean on this and also fix the warnings.

The reason that I didn't modify `account_invoice_ubl` directly is basically two reasons:

1. I don't know what standard you were following when you made it, apparently not PEPPOL, and I didnt want to break things
2. I dont want a very long review process, there are few OCA members that are enough into the topic to be able to review

You could be the key though to fixing point 1: if you send me the specs you followed, my customer has a contact at a PEPPOL agency and can ask questions as to why the specs are different and which one to follow for UBL.

alexis-via:
@thomaspaulb I didn't follow any CIUS when I made the initial implementation for UBL. I had a look at the available XML tags and made an implementation. On the opposite, for CII, I followed the Factur-X CIUS.
My proposal to make PEPPOL the reference CIUS for the OCA UBL stack is for Odoo v14, not for v13, so this PR is still fine. I propose to continue the discussion for v14 on https://github.com/OCA/edi/issues/246
In fact, I'm trying to figure out if some users could be bothered if we make PEPPOL the default CIUS. If we don't find any, then I would be in favor of make PEPPOL the default CIUS for v14.

rvalyi:
Hello guys, if I can add an input, I strongly advise you to take a look at https://github.com/tefra/xsdata for doing the Python bindings for the UBL or PEPPOL xsd gramar. It's is quite new, but it its really awesome, very clean and well tested. I did tests for electronic invoicing in Brazil and it works well, chances are we will move our bindings from generateDS to xsdata soon.

thomaspaulb:
@alexis-via If you didnt follow any particular CIUS then I agree with your proposal to dumb down `account_invoice_ubl` to PEPPOL; it's cleaner to add tags than to add them, and then remove them again.

I don't think this has to be decided rapidly; the PR of migrating this `_peppol` module to v14 could also include the removal of several tags from the `account_invoice_ubl` module, if it is already ported by then.

thomaspaulb:
@alexis-via About the forbidding of certain tags: there's lots of forbidding rules such as [this one](https://docs.peppol.eu/poacc/billing/3.0/rules/UBL-CR-143/), that are electronically enforced by tests such as [this one](https://github.com/OpenPEPPOL/peppol-bis-invoice-3/blob/master/rules/sch/CEN-EN16931-UBL.sch#L527). I don't know about other CIUS.

thomaspaulb:
@gjotten Since your above comment was not related to this module but to a problem in `l10n_nl_account_tax_unece`, maybe you can leave your positive functional review? :-)

thomaspaulb:
@NL66278 @gjotten @etobella @astirpe Your review would be appreciated :-)

KKamaa:
pre-commit issue unrelated, see below

```
************* Module sale_order_import_ubl_http.tests.test_sale_order_import_endpoint
sale_order_import_ubl_http/tests/test_sale_order_import_endpoint.py:12: [W7950(odoo-addons-relative-import), ] Same Odoo module absolute import. You should use relative import with "." instead of "openerp.addons.sale_order_import_ubl_http"
```

thomaspaulb:
@KKamaa 

As for the `pre-commit` issue, I fixed in https://github.com/OCA/edi/pull/545 and if you rebase this branch on that branch, or just cherry-pick the commit, it should turn green here too.

As for the Travis issue, it's shown [here](https://app.travis-ci.com/github/OCA/edi/jobs/563624040) and seems legitimate:

```
odoo.exceptions.UserError: ("The UBL XML file is not valid against the official XML Schema Definition. The XML file and the full error have been written in the server logs. Here is the error, which may give you an idea on the cause of the problem : Element '{urn:oasis:names:specification:ubl:schema:xsd:CommonBasicComponents-2}BuyerReference': This element is not expected. Expected is one of ( {urn:oasis:names:specification:ubl:schema:xsd:CommonBasicComponents-2}TaxCurrencyCode, {urn:oasis:names:specification:ubl:schema:xsd:CommonBasicComponents-2}PricingCurrencyCode, {urn:oasis:names:specification:ubl:schema:xsd:CommonBasicComponents-2}PaymentCurrencyCode, {urn:oasis:names:specification:ubl:schema:xsd:CommonBasicComponents-2}PaymentAlternativeCurrencyCode, {urn:oasis:names:specification:ubl:schema:xsd:CommonBasicComponents-2}AccountingCostCode, {urn:oasis:names:specification:ubl:schema:xsd:CommonBasicComponents-2}AccountingCost, {urn:oasis:names:specification:ubl:schema:xsd:CommonBasicComponents-2}LineCountNumeric, {urn:oasis:names:specification:ubl:schema:xsd:CommonAggregateComponents-2}InvoicePeriod, {urn:oasis:names:specification:ubl:schema:xsd:CommonAggregateComponents-2}OrderReference, {urn:oasis:names:specification:ubl:schema:xsd:CommonAggregateComponents-2}BillingReference )., line 8.", '')
```

Seems `BuyerReference` is added in the wrong place, I'll figure out where it should be and leave a comment.

account_invoice_ubl/models/account_move.py
```diff
@@ -248,6 +280,8 @@ def generate_invoice_ubl_xml_etree(self, version="2.1"):
         xml_root = etree.Element("Invoice", nsmap=nsmap)
         self._ubl_add_header(xml_root, ns, version=version)
         self._ubl_add_order_reference(xml_root, ns, version=version)
+        self._ubl_add_buyer_reference(xml_root, ns, version=version)
+        self._ubl_add_invoice_order_reference(xml_root, ns, version=version)
```
thomaspaulb: You're adding `OrderReference` here, but it's also added in `_ubl_add_order_reference` two lines above. You cannot add it twice, and it also should be added in the right order; so you'll have to remove this and modify the existing `_ubl_add_order_reference` instead.

account_invoice_ubl/models/account_move.py
```diff
@@ -248,6 +280,8 @@ def generate_invoice_ubl_xml_etree(self, version="2.1"):
         xml_root = etree.Element("Invoice", nsmap=nsmap)
         self._ubl_add_header(xml_root, ns, version=version)
         self._ubl_add_order_reference(xml_root, ns, version=version)
+        self._ubl_add_buyer_reference(xml_root, ns, version=version)
```
thomaspaulb: In the [XSD document specification](https://github.com/sunflowerit/edi/blob/13.0-add-account_invoice_ubl_peppol/base_ubl/data/xsd-2.2/maindoc/UBL-Invoice-2.2.xsd#L407) which the unit test checks for (and the online PEPPOL validators as well) it's specified that BuyerReference should come before OrderReference. So this one should go one line up, and then it might pass the test. If not, we still might have to shuffle things around a bit.

account_invoice_ubl/models/account_move.py
KKamaa: @thomaspaulb there is only one occurrence `_ubl_add_order_reference ` of this, the one I added and the other is in `sale_order_ubl` so we are only using `account_invoice_ubl` to add for invoice however I noticed its not required for validation and still green even after adding it. Think main issue was re-arrangement `_ubl_add_buyer_reference` coming before order

account_invoice_ubl/models/account_move.py
KKamaa: I think your were right about this coming before order all seems well now got a green validation after uploading attached sample.
image

account_invoice_ubl/models/account_move.py
thomaspaulb: Unfortunately this is still an issue, the guy who tested manages to create a condition where OrderReference is being added twice, and then he gets an error "OrderReference is not expected here" because [maxOccurrence is 1](https://github.com/sunflowerit/edi/blob/13.0-add-account_invoice_ubl_peppol/base_ubl/data/xsd-2.2/maindoc/UBL-Invoice-2.2.xsd#L441).

I think instead of adding a new function `_ubl_add_invoice_order_reference` we should just modify the existing `_ubl_add_order_reference` and choose which one we add. Both functions are in the same module anyway, in `account_invoice_ubl`, and even in the same file. (What does `sale_order_ubl` have to do with this?)

image

KKamaa: @thomaspaulb yeah you were right it creates a duplicate when we get an invoice from a sale order where the sale order name (i.e Origin in Invoice) is available. The invoice details already exist so there was no need to add them again. So I removed the piece code. Think its ok now.

thomaspaulb: Cool, asking them to test again.

bosd:
Thanks for all your work on this module!
This is a very welcomed addition 😃 

Results from functional test on runboat. 

[UBL-Invoice-2.1.zip](https://github.com/OCA/edi/files/8832837/UBL-Invoice-2.1.zip)

Validation with
https://test.peppolautoriteit.nl/validate


Results in:
image

bosd:
Propably worthwhile to mention. My first attempt to generate an invoice failed.
Because of the unece tax code match on the ECOM-07 product failed.
After clicking "print & send button"
Got an user error
image

Which is ok. Thereafter the whole invoice generation failed.
Meaning: There was no PDF file generated, the e-mail message was empty.

The user error was not on top, but behind the email window:.
image

Maybe not fully related with this pr.. but thinking if this is the desired way of handling such exceptions 🤔

bosd:
In addition..
The UX is not very clear in the settings area.
It is very hard to see, that it was actually an area where one could type in between the `[` `]` characters.
Maybe use an regular input field??
Screenshot_2022-06-04_22-29-45

Typing text in there is also error prone.
Not an export on this, is there an widget we can utilize to handle this in a more gui way?
Maybe [this](https://github.com/OCA/web/tree/13.0/web_widget_domain_editor_dialog)??

bosd:
Coming back to my UX comment.
Maybe easier to set per partner which UBL dialect should be used.
Likewise the transmission method can be set on the partner view.
image

account_invoice_ubl_peppol/__manifest__.py
```diff
@@ -0,0 +1,21 @@
...
+    "website": "https://github.com/OCA/edi",
+    "license": "AGPL-3",
+    "depends": ["account_invoice_ubl", "partner_coc"],
+    "data": [
+        "security/ir.model.access.csv",
+        "views/res_config_settings.xml",
+        "views/peppol_eas_list.xml",
+        "data/peppol.eas.list.csv",
+        "views/res_country.xml",
+        "views/peppol_eas_list.xml",
```
bosd: This line is same as line 15

account_invoice_ubl_peppol/data/peppol.eas.list.csv
```diff
@@ -0,0 +1,77 @@
...
+account_invoice_ubl_peppol.code_0196,0196,Kennitala - Iceland legal id for individuals and legal entities
+account_invoice_ubl_peppol.code_0198,0198,ERSTORG
+account_invoice_ubl_peppol.code_0199,0199,Legal Entity Identifier (LEI)
+account_invoice_ubl_peppol.code_0200,0200,Legal entity code (Lithuania)
+account_invoice_ubl_peppol.code_0201,0201,Codice Univoco Unità Organizzativa iPA
+account_invoice_ubl_peppol.code_0202,0202,Indirizzo di Posta Elettronica Certificata
+account_invoice_ubl_peppol.code_0204,0204,Leitweg-ID
+account_invoice_ubl_peppol.code_0208,0208,Numero d'entreprise / ondernemingsnummer / Unternehmensnummer
+account_invoice_ubl_peppol.code_0209,0209,GS1 identification keys
+account_invoice_ubl_peppol.code_9901,9901,Danish Ministry of the Interior and Health
```
bosd, line 33: Before this line the data looks to be coming from: [https://docs.peppol.eu/poacc/billing/3.0/codelist/ICD/](https://docs.peppol.eu/poacc/billing/3.0/codelist/ICD/)
ISO 6523 ICD list.

Maybe offload this to a module in the community-data-files repo? So other modules can access this as well?

EDIT: The 9000 numbers seems to be coming from:
https://github.com/OpenPEPPOL/peppol-bis-invoice-3/blob/master/structure/codelist/eas.xml

account_invoice_ubl_peppol/README.rst
```diff
@@ -0,0 +1,122 @@
...
+  satisfy rule BR-23.
+
+- Go to menu *Contacts*
+  Fill the field `coc_registration_number` for your own company's partner record and for
+  those partners that you want to send e-invoices to.
+
+- Go to menu *Contacts > Configuration > Localization > Countries*
+  On any country relevant for invoice traffic, configure the correct PEPPOL EAS id.
+  For the Netherlands, this is for example `0106`, which stands for Dutch chamber of
+  commerce number.
```
bosd, line 68: Clear instructions. But how does this actually work? Does a mapping need to be generated for each EAS ID to an odoo field? Can the mapping be created by a script?
In this example the Dutch Chamber Of Commerce number is added by oca module partner_coc (from V15 and up in odoo core). So could it be mapped automatically?

There is only one field for the mapping. But according to the standard more options are available and allowed.
Those are in this current version not included in the output.
Some of them could be constructed automatically like dutch `VAT` number, `9944`
(If partner country code is NL then VAT number code is  `9944`. If partner country code is DE then VAT number code is `9930`.

account_invoice_ubl_peppol/views/res_config_settings.xml
```diff
@@ -0,0 +1,47 @@
...
+                        />
+                    </div>
+                    <label for="ubl_default_tax" />
+                    <div class="text-muted">
+                        Specify a default tax to use for invoice lines with no tax.
+                    </div>
+                    <div>
+                        <field name="ubl_default_tax" />
+                    </div>
+                    <label for="ubl_default_tax" />
```
bosd: Should be `ubl_default_uom_id`
image

account_invoice_ubl_peppol/readme/USAGE.rst
```diff
@@ -0,0 +1,4 @@
+In the invoice form click on button `Send & Print`.
+
+If the invoice matches the configured domain for PEPPOL, the invoice will be generated
+and validated according to the stricter PEPPOL 3.0 BIS standard.
```
bosd, line 4: For convenience, please add a link to the validator. And add as instruction to validate the generated file.
While reviewing this module I noticed that in the current workflow a couple of parameters needs to be configured to generate a correct peppol invoice.
(In current workflow there are no warnings if misconfigured something which results that your generated file is not compliant to the peppol standard.)

bosd (changes requested):
Played a bit with this module to do a functional review.
As you can tell, I have no knowledge on this subject. Hopefully my monkey test is of any use.

After some of my user errors. I was able to generate a "almost" peppol compliant outcome.
Keep getting the BR-61 error.
And warnings on the inclusion of Delivery location adress.
image
[Invoice_INV_2022_0008(1).xml.tar.gz](https://github.com/OCA/edi/files/8840085/Invoice_INV_2022_0008.1.xml.tar.gz)

thomaspaulb:
> Typing text in there is also error prone. Not an export on this, is there an widget we can utilize to handle this in a more gui way? Maybe [this](https://github.com/OCA/web/tree/13.0/web_widget_domain_editor_dialog)??

Good comment! I've refactored it to use the Odoo domain widget:

image

account_invoice_ubl_peppol/README.rst
thomaspaulb: I think these could be automatically mapped, but since that is a long and tedious work, it fell out of the scope of the budget for this project. I'll add it to the roadmap.

For the `99XX` codes, I'm not sure these are actually allowed, because when I used one I got this validation error:

image

Seems only the lower codes are allowed in the standard.

account_invoice_ubl_peppol/data/peppol.eas.list.csv
thomaspaulb: Good idea, but not crucial to get this module merged IMO. I've added a ROADMAP item for this as well.

thomaspaulb:
> Keep getting the BR-61 error. 

The BR-61 can be fixed by going to Payment modes, setting the "link to bank account" to "fixed", then going into the journal, selecting the bank account there, and the bank account chosen should be of type "IBAN", which comes from module `base_iban` which was not depended on by this module, so not present on Runboat probably. I've added the dependency.

On my local, I managed to produce the following demo invoice in that way, which passes all the errors and warnings.

> And warnings on the inclusion of Delivery location adress.

I've added this to the roadmap also, since it's a point of note but not crucial either.

[Uploading UBL-Invoice-2.1.zip…]()

thomaspaulb:
@gjotten @bosd Perhaps a renewed review here, so that we can get the required approvals to get this merged! :rocket:

thomaspaulb:
@StefanRijnhart @NL66278 Gentle request for a review here :-) It's a useful module for the EU ecosphere.

gjotten:
@OCA/edi-maintainers can we get a "No Stale" label on this PR please? Adding accounting standardization/compliancy features can be a stop-start thing since needs tend to be highly (localization-) specific.

bosd:
> @gjotten @bosd Perhaps a renewed review here, so that we can get the required approvals to get this merged! :rocket: 

Sure, but it will take a while. I'm already enjoying some Belgium beers. 
Will hope to be able to do so next week :)

thomaspaulb:
@bosd Same here:)

bosd:
@thomaspaulb  Hope you've enjoyed the beers at oxp!
Thanks for all your work on this module!!

Played around a bit with module on runboat.
Created a dutch company "NL Bedrijf" to make this module easier to test.

I Like the UI Improvements!! :+1: 

Probably my bad, but somehow the settings default_tax and `default_uom` were not set/saved.
Around that time I've generated this invoice file.
[Invoice_INV_2022_0003 .txt](https://github.com/OCA/edi/files/9896661/Invoice_INV_2022_0003.txt)
(renamed to txt so I can upload here)
image
It looked like there is no tax line included when the default_tax is not set.
Thus it ignores the tax which has been set on the invoice line. Is that intentional or my bad in the config :question: 

After setting the `default_tax` I was able to generate this. 
[Invoice_INV_2022_0004.txt](https://github.com/OCA/edi/files/9896674/Invoice_INV_2022_0004.txt)

image
Only 1 red item, getting there! :fireworks: 

There seems to be an empty `<cac:Contact/>` tag on line 36.
I've deleted that line and the red error is gone. Now it is peppol compliant :sparkles: :tada: :sparkle:  


Edit: Setting an e-mail address on the company in odoo, fills the empty `<cac:Contact/>`  with an e-mail adress. Then the generated peppol invoices are compliant. As it is common to have an e-mail adress set. I consider this non blocking and approving this PR!

bosd (approved):
LGTM!!

bofiltd (approved):
Functional testing LGTM!

thomaspaulb:
@OCA/edi-maintainers Maybe good to now merge this? There are more improvements possible but these can then be handled in forward-ports and separate PR's.

bosd:
Lets merge ! :fireworks:

bosd:
@simahawk  Can you merge this? :pray:

simahawk:
can't we squash most of the commits after [[ADD] account_invoice_ubl_peppol](https://github.com/OCA/edi/pull/323/commits/3fe2398940a9ee208e91983331334809eef0f76b) ?

account_invoice_ubl_peppol/models/account_move.py
```diff
@@ -0,0 +1,404 @@
...
+
+    def _account_invoice_ubl_use_peppol(self):
+        """ Returns True when we should use PEPPOL """
+        domain = self.company_id.get_ubl_domain_peppol()
+        return bool(self.filtered_domain(domain))
+
+    def generate_ubl_xml_string(self, version="2.1"):
+        self.ensure_one()
+        if self._account_invoice_ubl_use_peppol():
+            this = self.with_context(account_invoice_ubl_use_peppol=True)
```
etobella, line 20: Shouldn't be better to add this to a field computed in post and store it on the invoice? 

With the current configuration, generation of the ubl is not consistent over time, as it depends on some rare configuration defined on a domain field inside the company. If we change this value, the generated document will change.

Also, I don't understand the need of the domain. Which is the standard use? Only for some journals or customers?

thomaspaulb:
> can't we squash most of the commits after [[ADD] account_invoice_ubl_peppol](https://github.com/OCA/edi/pull/323/commits/3fe2398940a9ee208e91983331334809eef0f76b) ?

Yes, you're right, I squashed them.

account_invoice_ubl_peppol/models/account_move.py
thomaspaulb: I agree with your point that this should become a stored computed field.

My idea with the domain was that I assumed that the current UBL implementation was being used by someone somewhere, who made it that way for a reason, eg because it's how UBL should be done in France, and in come I with my PEPPOL hacking and slashing, and perhaps that's good for NL invoices but not for all countries. So the main idea was for it to be a method whereby people can say hey, we use this for NL, or even for whole EU, but for the other invoices we use the normal `_ubl` implementation.

However, earlier during review Alexis already kicked in and said that there wasn't any specific scheme used and that we might as well make PEPPOL the default implementation and let any exceptions identify themselves as child modules that ADD something, instead of now the PEPPOL module mostly REMOVING things. But he suggested doing that in the 14.0+ version.

I can refactor, but we can also leave that for the 14.0+ version. Is this blocking for you?

bosd:
Can you re-trigger the runboat build??
What needs to be done to get this one merged??

simahawk:
This commit is broken https://github.com/OCA/edi/pull/323/commits/f52f93396f2b041a8b1a23231a9b93ecc21ed084 it has a rebase conflict. Can you check pls?

thomaspaulb:
@simahawk I checked, but that's an old commit, it's not part of the current PR anymore.

I pushed an empty commit to retrigger runboat, let's see what it does.

thomaspaulb:
Looks like it went green, removed the commit and force-pushed again.

simahawk:
@etobella fine for you?

etobella (approved):

PR #325, opened 2021-02-17
[13.0][IMP] edi: Add some dependencies in requirements.txt

victoralmau:
Add some dependencies in `requirements.txt` that already exists in `test_requirements.txt` (to prevent future errors)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,3 +3,4 @@ lxml
 PyPDF2
 pyyaml
 xmlschema
+freezegun


victoralmau:
Ping @pedrobaeza

simahawk (changes requested):
it's already here https://github.com/OCA/edi/blob/13.0/test-requirements.txt
Why do we need them again?
IMO we should fix travis-ci setup if this is needed.

victoralmau:
It's similar to https://github.com/OCA/stock-logistics-workflow/pull/786 (warning appear in `edi` tests)

simahawk:
> It's similar to [OCA/stock-logistics-workflow#786](https://github.com/OCA/stock-logistics-workflow/pull/786) (warning appear in `edi` tests)

yes but here we have test-requirements.txt because MQT=PIP.

So, if this is not used by travis-ci setup we should make it use it.

@sbidoul do you see any drawback in running [this line](https://github.com/OCA/maintainer-quality-tools/blob/master/travis/travis_install_nightly#L136) for both kind of setup?

sbidoul:
> @sbidoul do you see any drawback in running this line for both kind of setup?

At first glance that would make complete sense and I don't see any drawback.

simahawk:
quick untested patch here https://github.com/OCA/maintainer-quality-tools/pull/687

PR #327, opened 2021-02-18
[13.0] account_invoice_facturx: backport no-tax-bugfix from v14

alexis-via:
It will avoid an error when there are no taxes on invoice lines.

Also backport other small improvements


Chabagutor (approved):
Works fine thanks !

PR #328, opened 2021-02-19
[13.0][ADD] sale_order_customer_free_ref

TDu:
The goal of this module is to improve on the `client_order_ref` on `sale.order`.

By default Odoo only has one field to handle the customer reference of a sale order.
But when using EDI some specifications allow to have two fields for this, the customer
order Id and a customer (free) reference.

To help with this, this module adds two specific fields for them and transform the
`client_order_ref` standard field into a computed one.

The two new fields are also passed on to created invoices.

### DIFF
--- a/sale_order_customer_free_ref/__manifest__.py
+++ b/sale_order_customer_free_ref/__manifest__.py
@@ -1,0 +1,17 @@
+# Copyright 2021 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+{
+    "name": "Sale Order Customer Free Reference",
+    "version": "13.0.1.0.0",
+    "category": "Sale",
+    "license": "AGPL-3",
+    "summary": "Splits the Customer Reference on sale orders into two fields. "
+    "An Id and a Free reference. The existing field is transformed "
+    "into a computed one.",
+    "author": "Camptocamp SA, Odoo Community Association (OCA)",
+    "website": "https://github.com/oca/edi",
+    "depends": ["sale"],
+    "data": ["views/sale_order.xml", "views/account_move.xml"],
+    "installable": True,
+}
--- a/sale_order_customer_free_ref/models/sale_order.py
+++ b/sale_order_customer_free_ref/models/sale_order.py
@@ -1,0 +1,45 @@
+# Copyright 2021 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+from odoo import api, fields, models
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    _client_order_ref_separator = " - "
+
+    client_order_ref = fields.Char(
+        compute="_compute_client_order_ref",
+        inverse="_inverse_client_order_ref",
+        string="Customer Reference",
+        store=True,
+        copy=False,
+    )
+    customer_order_number = fields.Char(string="Customer Order Number", copy=False)
+    customer_order_free_ref = fields.Char(string="Customer Free Reference", copy=False)
+
+    @api.depends("customer_order_number", "customer_order_free_ref")
+    def _compute_client_order_ref(self):
+        for order in self:
+            refs = [order.customer_order_number, order.customer_order_free_ref]
+            refs = [ref for ref in refs if ref and ref.strip()]
+            order.client_order_ref = self._client_order_ref_separator.join(refs)
+
+    def _inverse_client_order_ref(self):
+        for order in self:
+            if not order.client_order_ref:
+                order.customer_order_number = ""
+                order.customer_order_free_ref = ""
+            refs = order.client_order_ref.split(self._client_order_ref_separator, 1)
+            order.customer_order_number = refs[0]
+            if len(refs) == 2:
+                order.customer_order_free_ref = refs[1]
+            else:
+                order.customer_order_free_ref = ""
+
+    def _prepare_invoice(self):
+        res = super()._prepare_invoice()
+        res["customer_order_number"] = self.customer_order_number
+        res["customer_order_free_ref"] = self.customer_order_free_ref
+        return res
--- a/sale_order_customer_free_ref/models/sale_order.py
+++ b/sale_order_customer_free_ref/models/sale_order.py
@@ -31,12 +31,13 @@ class SaleOrder(models.Model):
             if not order.client_order_ref:
                 order.customer_order_number = ""
                 order.customer_order_free_ref = ""
-            refs = order.client_order_ref.split(self._client_order_ref_separator, 1)
-            order.customer_order_number = refs[0]
-            if len(refs) == 2:
-                order.customer_order_free_ref = refs[1]
             else:
-                order.customer_order_free_ref = ""
+                refs = order.client_order_ref.split(self._client_order_ref_separator, 1)
+                order.customer_order_number = refs[0]
+                if len(refs) == 2:
+                    order.customer_order_free_ref = refs[1]
+                else:
+                    order.customer_order_free_ref = ""
 
     def _prepare_invoice(self):
         res = super()._prepare_invoice()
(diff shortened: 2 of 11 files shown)


sale_order_customer_free_ref/__manifest__.py
```diff
@@ -0,0 +1,17 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+{
+    "name": "Sale Order Customer Free Reference",
+    "version": "13.0.1.0.0",
+    "category": "Sale",
+    "license": "AGPL-3",
+    "summary": "Splits the Customer Reference on sale orders into two fields."
+    "An Id and a Free reference. The existing field is transformed"
+    "into a computed one.",
```
carlosserratoro: Those strings will be concatenated so no space will be left between words being in the tails of each line.

```suggestion
    "summary": "Splits the Customer Reference on sale orders into two fields. "
    "An Id and a Free reference. The existing field is transformed "
    "into a computed one.",
```

sale_order_customer_free_ref/models/sale_order.py
```diff
@@ -0,0 +1,31 @@
...
+        readonly=True,
+    )
+    customer_order_number = fields.Char(string="Customer Order Number", copy=False)
+    customer_order_free_ref = fields.Char(string="Customer Free Reference", copy=False)
+
+    @api.depends("customer_order_number", "customer_order_free_ref")
+    def _compute_client_order_ref(self):
+        for order in self:
+            refs = [order.customer_order_number, order.customer_order_free_ref]
+            refs = [ref for ref in refs if ref]
```
simahawk:
```suggestion
            refs = [ref for ref in refs if ref and ref.strip()]
```

sale_order_customer_free_ref/models/sale_order.py
```diff
@@ -0,0 +1,31 @@
...
+            order.client_order_ref = " - ".join(refs)
```
simahawk:
```suggestion
            order.client_order_ref = self._client_order_ref_separator.join(refs)

    _client_order_ref_separator = " - "
```
to ease customization

sale_order_customer_free_ref/models/sale_order.py
```diff
@@ -0,0 +1,31 @@
...
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    client_order_ref = fields.Char(
+        compute="_compute_client_order_ref",
+        string="Customer Reference",
+        store=True,
+        copy=False,
+        readonly=True,
```
simahawk: shouldn't be `readonly=False` to preserve std behavior? :thinking:

simahawk (changes requested):

sale_order_customer_free_ref/models/sale_order.py
```diff
@@ -0,0 +1,31 @@
...
+    )
+    customer_order_number = fields.Char(string="Customer Order Number", copy=False)
+    customer_order_free_ref = fields.Char(string="Customer Free Reference", copy=False)
```
carlosserratoro: Shouldn't this be the other way?, i.e. to make ```customer_order_number``` and ```customer_order_free_ref``` computed depending on the value set in the standard field ```client_order_ref```?

carlosserratoro (commented):
I have some doubts on this one.

sale_order_customer_free_ref/models/sale_order.py
carlosserratoro: But ```ref``` may be the empty string **or** a falsy value, and bools don't have the method ```strip()```.

simahawk: the 1st check will fail and you won't execute `strip` :wink:

sale_order_customer_free_ref/__manifest__.py
TDu: Thanks fixed.

sale_order_customer_free_ref/models/sale_order.py
TDu: Added a unit test to highlight this. The ref.strip() is executed only if ref is True.
Fixed.

sale_order_customer_free_ref/models/sale_order.py
TDu: Added, thank you.

sale_order_customer_free_ref/models/sale_order.py
TDu: No the idea is `customer_order_number` and `customer_free_ref` are modified and `client_order_ref` is readonly.
Or that is what was requested.

sale_order_customer_free_ref/models/sale_order.py
TDu: Not what was requested, but I can hear that `client_order_ref` being readonly now could be problematic.

carlosserratoro (approved):

TDu:
After discussion, the field `client_order_ref` is kept editable (still computed) but set as readonly on the view (see last fixup)

simahawk (dismissed):

simahawk:
@TDu maybe `partner-contact` is a better repo for this? CC @OCA/edi-maintainers

sale_order_customer_free_ref/models/sale_order.py
```diff
@@ -0,0 +1,33 @@
...
+    _inherit = "sale.order"
+
+    _client_order_ref_separator = " - "
+
+    client_order_ref = fields.Char(
+        compute="_compute_client_order_ref",
+        string="Customer Reference",
+        store=True,
+        copy=False,
+        readonly=False,
```
sebalix: By doing that, no need for an `inverse`? What happens if we write a value in `client_order_ref`?

sebalix (approved):
One question, pre-approving

sale_order_customer_free_ref/models/sale_order.py
sebalix: Maybe the failing job answers my question: can you check?

sebalix (commented):
I prefer to not approve then.

sale_order_customer_free_ref/models/sale_order.py
TDu: The last fixup adds an inverse method...

TDu:
@sebalix Adding the `inverse`method fixed the tests, thanks for the hint

sale_order_customer_free_ref/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,10 @@
+The goal of this module is to improve on the `client_order_ref` on `sale.order`.
+
+By default Odoo only has one field to handle the customer reference of a sale order.
```
jbaudoux:
```suggestion
By default, Odoo only has one field to handle the customer reference of a sales order.
```

sale_order_customer_free_ref/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,10 @@
...
+But when using EDI some specifications allow to have two fields for this, the customer
```
jbaudoux:
```suggestion
However, an order provided by the buyer can contain an id (PO number) and a free reference. When exchanging with the other parties, you sometimes need to differentiate them. For instance, this is required by some EDI flows.
```

sale_order_customer_free_ref/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,10 @@
...
+order Id and a customer (free) reference.
+
+To help with this, this module adds two specific fields for them and transform the
```
jbaudoux:
```suggestion
To help with this, this module adds two specific fields for them and transforms the
```

sale_order_customer_free_ref/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,10 @@
...
+`client_order_ref` standard field into a computed one.
+
+The two new fields are also passed on to created invoices.
```
jbaudoux:
```suggestion
The two new fields are also given when generating invoices.
```

jbaudoux (approved):
Just some English rephrasing

sale_order_customer_free_ref/readme/DESCRIPTION.rst
TDu: Thanks for all the suggestion. Updated.
And squashed.

PR #267, opened 2020-11-30
[13.0] [MIG] account invoice import ubl

NL66278:

PR #334, opened 2021-03-05
[14.0][MIG] account_invoice_import_ubl

alexis-via:


CasVissers-360ERP:
@alexis-via Can you see if my feedback in #267 can be considered?

dsolanki-initos:
Hello @alexis-via can you please resolve the failed checks?

fshah-initos:
@alexis-via  It would be great if you can resolve the failed checks

StefanRijnhart:
There is a superseding PR here: https://github.com/OCA/edi/pull/390 based on the 13.0 port (https://github.com/OCA/edi/pull/267)

The only difference, as far as I can tell, is that it has XML linting applied on the UBL XML files, and demo_data.xml has gone through a round trip of having its indentation level modified in the 13.0 port and restored in the 14.0 port.

@alexis-via do you feel that the UBL XML files should be left alone by the XML linter? We could consider having them ignored in that case.

alexis-via:
@StefanRijnhart I prefer if the demo/test XML files are left unchanged (but it's not a big deal). I don't know how to make them ignored by pre-commit/flake8 (and I need that for other modules too). I remember you mentioned the solution in another comment... but I can't find it at the moment.

PR #339, opened 2021-03-10
[13.0][IMP] account_invoice_export send button visibility

TDu:
This change the visibility of the send eBill button on the invoice.
It is now only shown if the option is checked on the transmit method.
Also improve the configuration on the transmit method adding
information on the process.

### DIFF
--- a/account_invoice_export/views/transmit_method.xml
+++ b/account_invoice_export/views/transmit_method.xml
@@ -14,6 +14,17 @@
                     attrs="{'invisible': [('customer_ok', '=', False)]}"
                 >
                     <field name="send_through_http" />
+                    <div
+                        role="alert"
+                        class="alert alert-info"
+                        colspan="2"
+                        attrs="{'invisible': [('send_through_http', '=', False)]}"
+                    >
+                        <p
+                        >By default the PDF of the invoice will be sent using the connection parameters below (basic authenticaiton).</p>
+                        <p
+                        >Handling specific connection needs and/or exporting other files can be done through code.</p>
+                    </div>
                     <field
                         name="destination_url"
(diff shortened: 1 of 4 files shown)


account_invoice_export/__manifest__.py
```diff
@@ -3,7 +3,7 @@
 
 {
     "name": "Account Invoice Export",
-    "version": "13.0.1.0.0",
+    "version": "13.0.1.1.0",
```
etobella: Don't change the number, it will be changed automatically by the bot.

TDu: Oh, I did not know that, thanks. Will fix

account_invoice_export/views/transmit_method.xml
```diff
@@ -14,6 +14,16 @@
...
                 >
                     <field name="send_through_http" />
+                    <div
+                        role="alert"
+                        class="alert alert-info"
+                        colspan="2"
+                        attrs="{'invisible': [('send_through_http', '=', False)]}"
+                    >
+                      <span
+                        >By default the PDF of the invoice will be send using the connection parameters below (basic authenticaiton).<br
```
simahawk:
```suggestion
                        >By default the PDF of the invoice will be sent using the connection parameters below (basic authenticaiton).<br
```
also, I'd remove the BR and use 2 P so that you won't have any markup in the translation

simahawk (commented):
LG, minor change.
Can you squash on next push?

account_invoice_export/views/transmit_method.xml
TDu: fixed

simahawk (approved):

PR #340, opened 2021-03-10
12.0 add base edifact

MPoch-PlanetaTIC:
This module is the base to generate and parse Edifact files.
This module does nothing by itself, but is used by others, for example: sale_order_import_edifact, that imports Edifact sale orders.

### DIFF
--- a/base_edifact/models/edifact.py
+++ b/base_edifact/models/edifact.py
@@ -200,7 +200,7 @@ QTYLIN|47|%s|%s\n' % (int(invoice_line.quantity), uom_code,
         discounted_price = invoice_line.price_unit - (
             invoice_line.price_unit * invoice_line.discount / 100.0)
         res = 'PRILIN|AAA|%.2f\n\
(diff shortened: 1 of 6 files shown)


base_edifact/models/edifact.py
```diff
@@ -0,0 +1,364 @@
...
+    def edifact_date(self, date):
+        return '%s%s%s' % (date.year,
+                           '{0:02d}'.format(date.month),
+                           '{0:02d}'.format(date.day))
+
+    # Methods to generate: invoice
+    @api.model
+    def edifact_invoice_init(self):
+        return 'INVOIC_D_93A_UN_EAN007\n'
+
```
bealdav, line 45: Thanks for the initiative on this matter.
Don't you think method with `invoice` should move to your edifact module dedicated to invoice ?

bealdav:
Are these development are on a specific version of EDIFACT. 
Could you document it ?

I don't know if these kind of module are in production, but when they'll it could be nice to merge them.

Thanks a lot

PR #343, opened 2021-03-10
[CHG] edi: Use more permissive licence: AGPL-> LGPL

etobella:
This change affects to
* base_edi
* edi
* edi_account
* edi_exchange_template
* edi_xml

It does not affect `edi_storage` as it's main module is AGPL.

@simahawk


simahawk:
@sbidoul is this fine for you?

sbidoul (approved):
Yes. My opinion is that low level modules (base_rest, graphql, queue_job, edi framework etc) are best with a permissive license so we can use them with custom modules that depend on OE.

simahawk:
@etobella regarding `edi_storage` let's ask @sebastienbeau @lmignon if we want to change license for storage base modules?

simahawk (approved):

JordiBForgeFlow (approved):

lmignon (approved):
I also agree that we should also re licence `storage` addons to LGPL

simahawk:
@etobella could you propose a change there too?

etobella:
Sure!

marcelsavegnago (approved):

PR #345, opened 2021-03-10
[ADD] Account Invoice Factur-X Buyer Reference

ivantodorovich:
This PR adds a facturx_ref field to res.partner:

image

When set, it's used as BuyerReference in facturx xml.

This is required to comply with some countries regulations. For instance, when selling to German Public Organizations this field need to be set with a "routing number" (Leitweg ID)

I considered adding this to l10n-germany AFAIK this affects any company from any eu country invoicing to a german public organization

@alexis-via your opinion is most welcomed :)

### DIFF
--- a/account_invoice_facturx_buyer_ref/models/account_invoice.py
+++ b/account_invoice_facturx_buyer_ref/models/account_invoice.py
@@ -1,0 +1,17 @@
+# Copyright 2021 Camptocamp
+# @author: Iván Todorovich <ivan.todorovich@gmail.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import api, models
+
+
+class AccountInvoice(models.Model):
+    _inherit = "account.invoice"
+
+    @api.model
+    def _cii_trade_agreement_buyer_ref(self, partner):
+        return (
+            partner.facturx_ref
+            or partner.commercial_partner_id.facturx_ref
+            or super()._cii_trade_agreement_buyer_ref(partner)
+        )
--- a/account_invoice_facturx_buyer_ref/__init__.py
+++ b/account_invoice_facturx_buyer_ref/__init__.py
@@ -1,0 +1,1 @@
+from . import models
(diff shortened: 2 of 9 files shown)


i-vyshnevska (approved):

alexis-via:
I agree with the need to implement this, but we won't implement it that way. We can't add a field on res.partner that will be used by 0.1% of companies using account_invoice_facturx. The location of the field for Leitweg ID needs to be adressed in the german localization (maybe via a module above partner_identification from OCA/partner-contact) and then a module l10n_de_account_invoice_facturx will add that info in the proper XML tag. There is already a similar module l10n_fr_account_invoice_facturx in OCA/l10n-france.

ivantodorovich:
Hello @alexis-via ,

Thanks for you insight.

I thought about implementing it in the German localization, but our customer is not based in Germany himself, however he's selling to a German Public organization that requires this information to be in the invoices.

This regulation may only be required by German organizations, but it affects all EU companies:

>Companies that act as suppliers to the administration must be able to generate and send XRechnungen. **It does not matter whether the contractor comes from Germany or another EU country - all invoices to the authorities must comply with the new standard**. On the other hand, companies that are exclusively active in the B2C or B2B sector are not affected.

The problem I see with implementing this (and possible other similar things) in localization repositories is that a company carrying inter-EU sales will need modules from all localization repositories to be compliant, even if this company is not based in that country. That may lead to them not even being aware that the module exist or that they need it.

I agree with you that this will only be used by only a very limited number of companies, so what do you think if we isolate this in a separate module `account_invoice_facturx_buyer_reference`, in this repository? This way it can be installed by anyone who needs to set the facturx BuyerReference (I'm not sure but there could be other countries or situations that require it, now or in the future, apart from german public organizations)

ivantodorovich:
@alexis-via I moved this to a separate module
Would you please check it out now? :pray:

account_invoice_facturx_buyer_ref/models/account_invoice.py
```diff
@@ -0,0 +1,17 @@
...
+
+class AccountInvoice(models.Model):
+    _inherit = "account.invoice"
+
+    @api.model
+    def _cii_trade_agreement_buyer_ref(self, partner):
+        return (
+            partner.facturx_ref
+            or partner.commercial_partner_id.facturx_ref
+            or super().remove_facturx_xml_format_in_pdf_invoice(partner)
```
i-vyshnevska: I see "remove_facturx_xml_format_in_pdf_invoice" defined in hooks but not in the model

ivantodorovich: OMG thanks !! I honestly don't know where that came from, wrong copy pasting I think :/

Fixed now :)

i-vyshnevska (approved):

PR #346, opened 2021-03-11
[13.0] Add edi_gs1 + edi_gs1_stock

simahawk:


edi/models/edi_exchange_type.py
```diff
@@ -121,24 +121,36 @@ def _check_backend(self):
...
+            "record_name": record_name,
+            "type": self,
+            "dt": dt,
+            "ext": ext,
+        }
+
+    def _make_exchange_filename(self, exchange_record):
+        """Generate filename."""
+        pattern = self.exchange_filename_pattern
+        pattern = pattern + ".{ext}"
```
simahawk, line 161: this should come from the value defined by the user. You should not enforce ext here.

simahawk (commented):
@acsonefho can we flag this PR as ready?

edi/models/edi_exchange_type.py
acsonefho: It's just to match with this: https://github.com/OCA/edi/pull/346/files#diff-a7911660a866fa99e022eaca3b199af8df77070ec6de5f00a5021421a27a6500R8
The extension should be added into the code after the file name pattern.

acsonefho:
> @acsonefho can we flag this PR as ready?

I still have some pending code but IMO we can already mark this one as ready.
As every tests are green.

simahawk:
@acsonefho can we consider this ready? If yes, please, ~convert it to a real PR~ I can convert it to real PR :sweat_smile: .

BTW  you might be interested in this https://github.com/OCA/edi/pull/402 :wink:

PR #350, opened 2021-03-18
[13.0][IMP] account_invoice_export custom url

TDu:
Add the option to customize the url used to export dynamically.

### DIFF
--- a/account_invoice_export/models/transmit_method.py
+++ b/account_invoice_export/models/transmit_method.py
@@ -26,3 +26,11 @@ class TransmitMethod(models.Model):
         auth = "{}:{}".format(self.destination_user, self.destination_pwd,)
         auth64 = base64.encodebytes(auth.encode("ascii"))[:-1]
         return {"Authorization": "Basic " + auth64.decode("utf-8")}
+
+    def get_transmission_url(self):
+        """Returns the base url used to export.
+
+        Override it to add variable parameters.
+        """
+        self.ensure_one()
+        return self.destination_url or ""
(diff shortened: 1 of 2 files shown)


simahawk (approved):

TDu:
I added one fix if url is empty.
And the same for user and password

account_invoice_export/models/transmit_method.py
```diff
@@ -21,6 +21,14 @@ def get_transmission_http_header(self):
 
         """
         self.ensure_one()
-        auth = "{}:{}".format(self.destination_user, self.destination_pwd,)
+        auth = "{}:{}".format(self.destination_user or "", self.destination_pwd or "")
```
simahawk, line 26: wouldn't be better to raise a specific error?

TDu: Well my thought was letting the call be made and let `request` or the distant server raise an error.
From what I know the password could be empty ?

simahawk: yeah but the error might be not so clear for the user. I guess destination user is mandatory? If yes, raising a UserError here or before should be better. Unless I miss something else.

TDu: From what  I read online and what people understand from the  specs https://datatracker.ietf.org/doc/html/rfc1945#section-11 Basic Authentication allows for an empty user name and password. That is sending just the semi column is a valid option.
In that light the code is correct.

TDu:
@simahawk  I answered our discussion above and rebased the pr

PR #351, opened 2021-03-18
[13.0][ADD] account_invoice_export_server_env

TDu:
This module is based on the `server_environment` module to use files for
configuration. So we can have a different configuration for each
environment (dev, test, integration, prod).  This module define the config
variables for the `account_invoice_export` module.

Exemple of the section to put in the configuration file::

    [transmit_method.transmition_method_name]
    destination_pwd: password,
    destination_user: user,

### DIFF
--- a/account_invoice_export_server_env/models/transmit_method.py
+++ b/account_invoice_export_server_env/models/transmit_method.py
@@ -1,0 +1,19 @@
+# Copyright 2021 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import models
+
+
+class TransmitMethod(models.Model):
+    _name = "transmit.method"
+    _inherit = ["transmit.method", "server.env.mixin"]
+
+    _server_env_section_name_field = "code"
+
+    @property
+    def _server_env_fields(self):
+        return {
+            "destination_user": {},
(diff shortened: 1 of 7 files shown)


account_invoice_export_server_env/models/transmit_method.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2021 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import models
+
+
+class TransmitMethod(models.Model):
+    _name = "transmit.method"
+    _inherit = ["transmit.method", "server.env.mixin"]
```
simahawk, line 9: I'd use https://github.com/OCA/server-env/blob/13.0/server_environment/models/server_env_tech_name_mixin.py as messing up conf by changing display names is quite easy ;)

TDu: As the `transmit_method` model as already a field named `code` that act as a technical field and is unique.
I did a change to use that one (see line below)

simahawk (approved):

TDu:
@simahawk I think this one is ready, if you pass by

nilshamerlinck:
Hi @simahawk seems a second attempt to merge could help on this one :)

PR #235, opened 2020-09-18
[13.0] base_ubl: Add support for ubl-2.2

jbaudoux:
Port support for UBL 2.2 from v10.0 to 13.0

PR #358, opened 2021-04-03
[13.0][IMP] base_ubl: Add support for ubl-2.2

TDu:
This adds the schema for the UBL version 2.2

The files have been linted, but...

The linter only complains about some end of lines and end of files
chars.
Also I don't think it is a good idea to change anything in those
files.
But the only way I know about preventing the linter to access them is to
add some exclude statment in the `.pre-commit-config.yaml` file, and
I guess those changes could be overwritten during next OCA pre-commit
changes.


TDu:
This is to follow up on #235
It has been squashed, so it can be merged
@simahawk

simahawk:
@TDu as per formatting: yes, we should skip them. Regarding the template: when we sync we should not merge blindly ;)

PR #362, opened 2021-04-07
[FIX] edi & edi_storage

etobella:
This PR solves three issues:
1- Security on edi
2- Filtering on edi (the records not assigned could not be found)
3- Sorting of components after edi storage was giving an error when the backend has a storage id.

@simahawk

### DIFF
--- a/edi/models/edi_exchange_record.py
+++ b/edi/models/edi_exchange_record.py
@@ -151,6 +151,12 @@ class EDIExchangeRecord(models.Model):
 
     @property
     def record(self):
+        # In some case the res_model (and res_id) could be empty so we have to load
+        # data from parent
+        if not self.model and not self.parent_id:
+            return None
+        if not self.model and self.parent_id:
+            return self.parent_id.record
         return self.env[self.model].browse(self.res_id)
 
     def _set_file_content(
--- a/edi/models/edi_exchange_record.py
+++ b/edi/models/edi_exchange_record.py
@@ -375,28 +375,27 @@ class EDIExchangeRecord(models.Model):
         super(EDIExchangeRecord, self).check_access_rule(operation)
         if self.env.is_superuser():
             return
-        for exchange_record in self.sudo():
-            if not exchange_record.model or not exchange_record.res_id:
+        default_checker = self.env["edi.exchange.consumer.mixin"].get_edi_access
+        by_model_rec_ids = defaultdict(set)
+        by_model_checker = {}
+        for exc_rec in self.sudo():
+            if not exc_rec.model or not exc_rec.res_id:
                 continue
-            record = (
-                self.env[exchange_record.model]
-                .browse(exchange_record.res_id)
-                .with_user(self._uid)
-            )
-            if hasattr(record, "get_edi_access"):
-                check_operation = record.get_edi_access(
-                    [exchange_record.res_id], operation
+            by_model_rec_ids[exc_rec.model].add(exc_rec.res_id)
+            if exc_rec.model not in by_model_checker:
+                by_model_checker[exc_rec.model] = getattr(
+                    self.env[exc_rec.model], "get_edi_access", default_checker
                 )
-            else:
-                check_operation = self.env[
-                    "edi.exchange.consumer.mixin"
-                ].get_edi_access(
-                    [exchange_record.res_id],
-                    operation,
-                    model_name=exchange_record.model,
+
+        for model, rec_ids in by_model_rec_ids.items():
+            records = self.env[model].browse(rec_ids).with_user(self._uid)
+            checker = by_model_checker[model]
+            for record in records:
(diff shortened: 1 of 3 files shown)


edi/models/edi_exchange_record.py
```diff
@@ -369,10 +372,16 @@ def check_access_rule(self, operation):
         if self.env.is_superuser():
             return
         for exchange_record in self:
-            if not exchange_record.record:
+            sudo_record = exchange_record.sudo()
```
etobella: This is the same change applied on 12.0

edi/models/edi_exchange_record.py
```diff
@@ -151,6 +151,8 @@ def needs_ack(self):
 
     @property
     def record(self):
+        if not self.model or self.model not in self.env:
```
etobella: The same change applied on 12.0

edi/models/edi_exchange_record.py
```diff
@@ -329,6 +331,7 @@ def _search(
             )
             for eid, res_id, model in self._cr.fetchall():
                 if not model:
+                    result.append(eid)
```
etobella, line 338: If no model, the result was lost

edi_storage/models/edi_backend.py
```diff
@@ -73,4 +73,6 @@ def _component_sort_key(self, component_class):
         # Override to give precedence by storage_backend_type when needed.
         if not self.storage_id:
             return res
-        return (1 if component_class._storage_backend_type else 0,) + res
+        return (
+            1 if getattr(component_class, "_storage_backend_type", False) else 0,
```
etobella, line 77: The same was applied on `edi_webservice` https://github.com/OCA/edi/blob/13.0/edi_webservice/models/edi_backend.py#L34

edi/models/edi_exchange_record.py
simahawk: I think we fixed this on https://github.com/OCA/edi/pull/346 CC @acsonefho

edi/models/edi_exchange_record.py
simahawk: wouldn't be better do it before? `for exchange_record in self.sudo()`?

edi/models/edi_exchange_record.py
```diff
@@ -369,10 +372,16 @@ def check_access_rule(self, operation):
         if self.env.is_superuser():
             return
         for exchange_record in self:
-            if not exchange_record.record:
+            sudo_record = exchange_record.sudo()
+            if not sudo_record.model or not sudo_record.res_id:
                 continue
-            if hasattr(exchange_record.record, "get_edi_access"):
-                check_operation = exchange_record.record.get_edi_access(
+            record = (
```
simahawk: same, would be better to do it for the whole loop IMO

simahawk (changes requested):

edi/models/edi_exchange_record.py
etobella: Sure

edi/models/edi_exchange_record.py
etobella: Yep, Cherry picked the commit, as the PR is still on draft and this solves some issues :wink:

edi/models/edi_exchange_record.py
etobella: The problem is that some records could have the edi_access and others not, so it is hard to remove the logic from here :thinking:

simahawk: I'd do something like this (completely untested :stuck_out_tongue: :


```py
        default_checker = self.env[
            "edi.exchange.consumer.mixin"
        ].get_edi_access
        exchange_records = self.sudo().filtered(lambda x: x.model and x.res_id)
        by_model_rec_ids = defaultdict(set)
        by_model_checker = {}
        for exc_rec in exchange_records:
            by_model_rec_ids[exc_rec.model].add(exc_rec.id)
            checker = getattr(self.env[exc_rec.model], "get_edi_access", default_checker)

        for model, rec_ids in by_model_rec_ids.items():
            records = self.env[model].browse(rec_ids).with_user(self._uid)
            checker = by_model_checker[model]
            for record in records:
                check_operation = checker([record.id], operation, model_name=record._name)
                record.check_access_rights(check_operation)
                record.check_access_rule(check_operation)
```

sounds good?

simahawk: BTW would you mind adding a test for permissions? Otherwise please add a TODO

etobella: Tests where already there :wink:  https://github.com/OCA/edi/blob/13.0/edi/tests/test_security.py

I adapted your code, as it seems cleaner :smile:

simahawk (approved):

albariera (approved):

PR #363, opened 2021-04-09
[14.0][MIG] base_business_document_import

alexis-via:
This PR is an improvement of PR #301 with additionnal tests and improvement in error handling. The migration of account_invoice_import is based on it.


rconjour:
@alexis-via can you rebase against https://github.com/OCA/edi/pull/301 because the merge conflict is resolved there.

rconjour (approved):
LGTM, however, i prefer to use the `default-paramater` on `dict.get`, but the `or` also does the trick!

StefanRijnhart:
Can we agree to merge #301 as is and then immediately merge this one afterwards? I think your work falls under 'improvements' so we can merge it separately from the migration PR and give credits where credits are due.

alexis-via:
@StefanRijnhart I kept the commits of @phuctranerp without changes.

StefanRijnhart:
Yes, I know that! So we can merge that PR first without any impact, right?

alexis-via:
yes, sure, we can do that. Merging this one directly would make the same result. Both scenarios are OK.

StefanRijnhart:
Oh, I did not expect that a rebase would be necessary but it is.

alexis-via:
OK, I'll try to find time to work on this in the coming days. I also need to finish the PR for account_invoice_import ; I'm almost ready.

rconjour:
@alexis-via i'll create a superseding PR for this, so you can continue on account_invoice_import

rconjour:
Created superseding PR: https://github.com/OCA/edi/pull/366/commits

StefanRijnhart:
Green on the superseding PR. Closing here

PR #364, opened 2021-04-13
[14.0] [MIG] purchase_order_import

rconjour:
[14.0] [MIG] purchase_order_import


purchase_order_import/wizard/order_response_import.py
```diff
@@ -376,7 +373,7 @@ def _check_picking_status(self, picking):
         :param picking:
         :return:
         """
-        if any(operation.qty_done != 0 for operation in picking.pack_operation_ids):
+        if any(operation.qty_done != 0 for operation in picking.move_line_ids):
```
StefanRijnhart: `move_line_ids` by itself represent reservations, not packing. The packing is represented by the `result_package_id` of each `stock.move.line`.

purchase_order_import/tests/test_order_response_import.py
```diff
@@ -555,29 +567,26 @@ def test_15(self):
...
+            lambda s: s.state == "assigned" and s.product_qty == 2
         )
         self.assertTrue(move_confirmed)
         self.assertEqual(
             _("%s items should be delivered into a next delivery.") % "2",
             move_confirmed.note,
         )
         move_cancel = move_ids.filtered(
-            lambda s: s.state == "cancel" and s.product_qty == 1
+            lambda s: s.state == "cancel" and s.product_qty == 8
```
StefanRijnhart: I'm a bit worried that the cancelled qty is now the original quantity minus the backorder quantity. While the total delivered and backorder quantities are correct, this should not be necessary given that recent version of Odoo allow to update the quantities of confirmed stock moves. Do you think that we can use that to end up with a cancelled quantity of 1 again (which seems the best outcome for an original quantity of 10 minus 7 delivered and 2 backordered).

purchase_order_import/wizard/order_response_import.py
StefanRijnhart: I have to come back to this. I misunderstood the functionality of this part of the code, and it may be better to check the move line's qty_done rather than that of the package. But before we can determine what the scope of this sanity check is, the whole amendment flow needs to be fixed.

The result of an amendment flow with backorder (test13) is:
- A new picking with the original stock moves in state 'assigned' and the split backorder stock move in state 'draft' (leading to the picking being in state 'draft' as is validated in that test), and
- The original picking without any stock moves (which is now no longer linked to the purchase order because the linkage goes though the stock moves).

This is because stock.picking's `_create_backorder` will already move the untransferred move lines to the backorder, and does not allow to pass specific moves anymore (cf. https://github.com/OCA/edi/pull/364/commits/d4cd2b730fdd967c278f8231e64bf2600e42621b#diff-d08823307907b2e25c3460d4d1e71ef3fec25281f073f35e6a942e750c6309f5L364)

I wonder what scenarios this module is supposed to support. The tests only assume one message per purchase order, but is it the case that the supplier announces backorder deliveries in one or more incremental updates? Or can a fully accepted purchase order be amended to create a backorder in an update message? This will affect the solution direction of the problem described above.

i-vyshnevska:
@StefanRijnhart @rconjour can we reopen this PR and continue work or better to open a new PR?

StefanRijnhart:
@i-vyshnevska if you think this code is good to go you can reopen and approve, but if it needs more changes (or a rebase), I would open a new one.

PR #371, opened 2021-05-04
[MIG] edi_oca: Migration to 14.0

etobella:
This PR migrates edi to 14.0 and changes the name to edi_oca. The commit history has been changed in order to have the proper names using `git filter-branch`.

Depends on:

- [x] #370
- [x] https://github.com/OCA/connector/pull/404

@simahawk @pedrobaeza


edi_oca/security/ir_model_access.xml
```diff
@@ -72,17 +72,26 @@
         <field name="perm_write" eval="1" />
         <field name="perm_unlink" eval="1" />
     </record>
+    <record model="ir.model.access" id="access_edi_exchange_record_create_user">
+        <field name="name">access_edi_exchange_record user</field>
```
simahawk, line 76: what is this? Users can alread create records w/ the rule above.

edi_oca/__manifest__.py
```diff
@@ -8,11 +8,12 @@
...
     """,
-    "version": "13.0.1.17.0",
+    "version": "14.0.1.0.0",
+    "website": "https://github.com/OCA/edi",
     "development_status": "Alpha",
     "license": "LGPL-3",
-    "author": "ACSONE,Odoo Community Association (OCA)",
-    "maintainers": ["simahawk"],
+    "author": "ACSONE,Creu Blanca,Odoo Community Association (OCA)",
+    "maintainers": ["simahawk", "etobella"],
```
simahawk, line 16: :partying_face:

simahawk (changes requested):
strange rule

simahawk:
worth to mention in the README that the module has been renamed maybe?

Also, nothing to do for the migration?

Base PR: merged. Connector PR: asked something,

edi_oca/security/ir_model_access.xml
etobella: This rule is for the wizard, on 14.0 we need to add permissions for the wizards too :wink:

simahawk: ahh sorry, I was looking only at the diff of the mig commit and I missed that the wiz name is `edi.exchange.record.create`. 
Not a blocker, but IMO it should be renamed to `edi.exchange.record.create.wiz`. What do you think?

etobella: sure!

etobella:
@simahawk All comments attended

simahawk:
Nothing to do for the migration?

etobella:
Not really, it was an easy migration. We will need to add the change of name on OpenUpgrade, but it was easy :wink:

simahawk (approved):

albariera (approved):

PR #372, opened 2021-05-06
[14.0] [mig] edi_storage_oca

fshah-initos:

PR #378, opened 2021-05-10
[FIX] edi_oca: Missing changes of name

etobella:
Some errors where found during the migration of related modules. For this reason #372 is failing

@fshah-initos


etobella:
I have tested the tests on #372 and worked properly, also a JS error was raised, now it is fixed (the widget would not appear...)

simahawk (approved):
seems more than legit :smile: 
Any other find/replace loop missing? :stuck_out_tongue_winking_eye:

etobella:
I hope there isn't more :laughing:

PR #379, opened 2021-05-11
[13.0][FIX] order_import_order_ubl_http fail if exists

TDu:
Raise an error if the same order is being imported a 2nd time.
This will make the job fail and will be noticed.
Otherwise the job is done but the order not updated.

### DIFF
--- a/sale_order_import_ubl_http/tests/test_sale_order_import_endpoint.py
+++ b/sale_order_import_ubl_http/tests/test_sale_order_import_endpoint.py
@@ -6,6 +6,7 @@ import os
 from werkzeug.exceptions import BadRequest, Unauthorized
 
 from odoo import tools
+from odoo.exceptions import UserError
 from odoo.tests.common import SingleTransactionCase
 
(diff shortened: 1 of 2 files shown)


mmequignon (approved):

sale_order_import_ubl_http/tests/test_sale_order_import_endpoint.py
```diff
@@ -42,6 +43,15 @@ def test_import_so_ubl(self):
         order_ref = res.split(" ")[2]
         new_order = self.env["sale.order"].search([("name", "=", order_ref)])
         self.assertEqual(new_order.state, "draft")
+        # Importing a 2nd time raises an error
+        with self.assertRaises(UserError):
```
simahawk:
```suggestion
        with self.assertRaisesRegex(UserError, "Sales order has already been imported before"):
```

TDu: Done

simahawk (approved):

PR #390, opened 2021-06-01
[14.0] [mig] account_invoice_import_ubl

fshah-initos:


fshah-initos:
Superseding #334 by @alexis-via

rconjour (approved):

account_invoice_import_ubl/tests/files/UBLKetentest_Referentiefactuur_20150100.xml
```diff
@@ -0,0 +1,242 @@
+<?xml version="1.0" encoding="UTF-8" ?>
```
PieterPaulussen, line 1: Is there not a more generic file name that can be used? `invoice_1.xml` for instance?

account_invoice_import_ubl/tests/test_ubl.py
```diff
@@ -0,0 +1,87 @@
...
+            wiz.import_invoice()
+            invoices = amo.search(
+                [
+                    ("state", "=", "draft"),
+                    ("move_type", "in", ("in_invoice", "in_refund")),
+                    ("ref", "=", res_dict["invoice_number"]),
+                ]
+            )
+            self.assertEqual(len(invoices), 1)
+            inv = invoices[0]
```
PieterPaulussen, line 61: Why not just do 
```
inv = amo.search(
    ...
)
self.assertEqual(len(inv), 1)
```
I find this shorter and equally clear code.

account_invoice_import_ubl/tests/test_ubl.py
```diff
@@ -0,0 +1,87 @@
...
+            self.assertEqual(inv.move_type, res_dict.get("move_type", "in_invoice"))
```
PieterPaulussen, line 62: `res_dict` does not contain a `move_type` key in the above defined dictionaries. Could this be shortened to 
```
self.assertEqual(inv.move_type, "in_invoice")
```
then?

account_invoice_import_ubl/tests/test_ubl.py
```diff
@@ -0,0 +1,87 @@
...
+            )
+            self.assertEqual(
+                float_compare(
+                    inv.amount_total,
+                    res_dict["amount_total"],
+                    precision_rounding=precision_rounding,
+                ),
+                0,
+            )
+            invoices.unlink()
```
PieterPaulussen, line 87: Why unlink since data is cleared after the test transaction has completed? Is each imported invoice possibly impacted by a previous one? Perhaps a refactor of this test to a `SavepointCase` would prove useful?

PieterPaulussen (changes requested):
I notice that most of the methods in in the wizard are not documented. Perhaps you could add them in this migration?

account_invoice_import_ubl/demo/demo_data.xml
```diff
@@ -1,82 +1,88 @@
...
-        <field name="vat">NL009276749B01</field>
-    </record>
-    <record id="multi_tank_import_config" model="account.invoice.import.config">
-        <field name="name">Multi Tank</field>
-        <field name="partner_id" ref="multi_tank" />
-        <field name="invoice_line_method">nline_no_product</field>
-        <field name="account_id" ref="charge_account" />
-    </record>
+
+<record id="charge_account" model="account.account">
```
StefanRijnhart, line 4: Can I ask you why the indentation changed? I'm surprised pre-commit allows this. Or did pre-commit do this itself?

PieterPaulussen:
@fshah-initos You can disregard my comments since these are not really in scope of the migration of this module.

PieterPaulussen (approved):

StefanRijnhart:
@fshah-initos what do you think of this suggestion https://github.com/OCA/edi/pull/334#issuecomment-805550722?

account_invoice_import_ubl/demo/demo_data.xml
StefanRijnhart: I see that the indentation was modified in the 13.0 port, and then restored in the 14.0 port. I don't mind having the indentation as it is now, but I think the refactoring should be squashed.
The original PR, which is not based on the 13.0 port does not have this problem BTW.

StefanRijnhart:
Because this is based on the 13.0 port, XML linting is applied to the UBL XML files already. Not sure if we should keep the original indentation. The original PR does not have the UBL XML linting applied (and that is why Travis currently fails for it).

fshah-initos:
Currently we should keep it as it is changes we can do it in another PR.

mylbco:
Can this be merged then @StefanRijnhart as suggested by @fshah-initos ?

StefanRijnhart:
I don't like the indentation roundtrip (if anything, the indentation in the 13.0 port is more compliant and IMHO should not be reverted here), and I think the UBL files should not be formatted and be ignored in `.prettierignore` instead.

fshah-initos:
Yes but the pre-commit allowed it and sholud not be a problem.

fshah-initos:
can we merge this PR?

StefanRijnhart:
I'll abstain from approving or merging but my comments should not be considered blocking.

fshah-initos:
Can we merge this?

PR #392, opened 2021-06-01
[FIX]bad requirements version

flotho:
prevent 
image

0.3.2 succeed

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,5 @@
-invoice2data>=0.2.74,<=0.3.2
+wheel
+invoice2data>=0.2.74
 factur-x>=0.3
 phonenumbers
 weboob
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-invoice2data>=0.2.74
+invoice2data>=0.2.74,<=0.3.2
 factur-x>=0.3
 phonenumbers
 weboob


flotho:
pop @alexis-via @legalsylvain , any validation heure would be appreciated

legalsylvain:
sorry, I'm not very aware of python lib trouble. i prefer to abstain on that point.

StefanRijnhart:
I don't understand how this is a problem. The Travis job of the 10.0 branch of this repo is green. Can you clarify when the error occurs?

Note that `wheel` is installed depending on some variable in the Travis scripts: https://github.com/OCA/maintainer-quality-tools/blob/master/travis/travis_install_nightly#L104-L122. If `wheel` needs to be available, it might be better to adapt those scripts rather than individual branches in the addons repos.

flotho:
Hi stefan, 
the problem occurs when making locally a pip install requirements.
the latest version of invoice2data > 0.3.2 and after this version the installation is requiring wheel.
So either we limit the version of invoice2data or we add the wheel package.
Am I doing a bad proposition ?

StefanRijnhart:
Well, as https://github.com/OCA/maintainer-quality-tools/blob/master/travis/travis_install_nightly#L104-L122 implies, whether or not to install wheel seems to be a responsibility of your local bootstrapping, and not a dependency of this repository. On a related note, I don't seem to have any problems installing invoice2data without wheel:

```
(venv_test2.7) odoo@odoo:~$ pip install invoice2data
Using legacy 'setup.py install' for invoice2data, since package 'wheel' is not installed.
Using legacy 'setup.py install' for regex, since package 'wheel' is not installed.
Installing collected packages: chardet, pillow, pyyaml, pytz, tzlocal, regex, six, python-dateutil, dateparser, unidecode, invoice2data
    Running setup.py install for regex ... done
    Running setup.py install for invoice2data ... done
Successfully installed chardet-4.0.0 dateparser-0.7.6 invoice2data-0.3.5 pillow-6.2.2 python-dateutil-2.8.1 pytz-2021.1 pyyaml-5.4.1 regex-2021.4.4 six-1.16.0 tzlocal-2.1 unidecode-1.2.0
WARNING: You are using pip version 20.2.1; however, version 20.3.4 is available.
You should consider upgrading via the '/home/odoo/tmp/venv_test2.7/bin/python -m pip install --upgrade pip' command.
```

simahawk:
I agree it should be handled on your own installation.

PR #394, opened 2021-06-07
[14.0] several fixes on account_invoice_import

alexis-via:
This PR include 2 fixes:
- bad analytic account field name
- bad default value for partner_id on account.invoice.import.config when the partner is created via the invoice import wizard.


rconjour (approved):

etobella (approved):
LGTM :+1:

PR #401, opened 2021-06-21
[13.0][FIX] webservice: server.env.mixin needs to be inherited

LoisRForgeFlow:
The funtionality is used by the module as the method
`_server_env_fields` is defined, but the inherit of the mixin
using it is missing.

@ForgeFlow

cc @etobella

### DIFF
--- a/webservice/models/webservice_backend.py
+++ b/webservice/models/webservice_backend.py
@@ -7,7 +7,7 @@ from odoo import fields, models
 class WebserviceBackend(models.Model):
 
     _name = "webservice.backend"
-    _inherit = ["collection.base", "server.env.techname.mixin"]
+    _inherit = ["collection.base", "server.env.techname.mixin", "server.env.mixin"]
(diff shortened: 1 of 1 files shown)


simahawk (approved):
seems legit :)

etobella:
Maybe it has more sense to add the inherit on the tech mixin, WDYT @simahawk ?

simahawk:
> Maybe it has more sense to add the inherit on the tech mixin, WDYT @simahawk ?

Well, I thought about it then I kept it this way for these reasons: keep them decoupled, make it explicit.
Now, if people prefer to make it inherit by default... fine.

BTW as I find relying on `name` fragile I'd even move the tech field stuff to the main mixin.

etobella (approved):

dreispt:
Travis is red.

LoisRForgeFlow:
I didn't notice it. Thanks, travis fixed!

PR #402, opened 2021-06-22
[13.0][IMP] edi_storage: implement basis of the input process.

LoisRForgeFlow:
* A cron job check related storage on edi backends to create the needed exchange files.
* Add basic receive component.

@ForgeFlow

### DIFF
--- a/edi_storage/components/base.py
+++ b/edi_storage/components/base.py
@@ -27,7 +27,7 @@ class EDIStorageComponentMixin(AbstractComponent):
     def storage(self):
         return self.backend.storage_id
 
-    def _dir_by_state(self, direction, state, absolute=False):
+    def _dir_by_state(self, direction, state):
         """Return remote directory path by direction and state.
 
         :param direction: string stating direction of the exchange
@@ -36,11 +36,11 @@ class EDIStorageComponentMixin(AbstractComponent):
         """
         assert direction in ("input", "output")
         assert state in ("pending", "done", "error")
-        if absolute:
-            return PurePath(self.backend[direction + "_dir_" + state] or "")
-        return PurePath((self.backend[direction + "_dir_" + state] or "").strip(" /"))
+        return PurePath(
+            (self.backend[direction + "_dir_" + state] or "").strip().rstrip("/")
+        )
 
-    def _remote_file_path(self, direction, state, filename, absolute=False):
+    def _remote_file_path(self, direction, state, filename):
         """Return remote file path by direction and state for give filename.
 
         :param direction: string stating direction of the exchange
@@ -48,7 +48,7 @@ class EDIStorageComponentMixin(AbstractComponent):
         :param filename: string for file name
         :return: PurePath object
         """
-        return self._dir_by_state(direction, state, absolute) / filename.strip("/ ")
+        return self._dir_by_state(direction, state) / filename.strip("/ ")
 
     def _get_remote_file(self, state, filename=None, binary=False):
         """Get file for current exchange_record in the given destination state.
--- a/edi_storage/components/base.py
+++ b/edi_storage/components/base.py
@@ -27,7 +27,7 @@ class EDIStorageComponentMixin(AbstractComponent):
     def storage(self):
         return self.backend.storage_id
 
-    def _dir_by_state(self, direction, state):
+    def _dir_by_state(self, direction, state, absolute=False):
         """Return remote directory path by direction and state.
 
         :param direction: string stating direction of the exchange
@@ -36,9 +36,11 @@ class EDIStorageComponentMixin(AbstractComponent):
         """
         assert direction in ("input", "output")
         assert state in ("pending", "done", "error")
+        if absolute:
+            return PurePath(self.backend[direction + "_dir_" + state] or "")
         return PurePath((self.backend[direction + "_dir_" + state] or "").strip(" /"))
 
-    def _remote_file_path(self, direction, state, filename):
+    def _remote_file_path(self, direction, state, filename, absolute=False):
         """Return remote file path by direction and state for give filename.
 
         :param direction: string stating direction of the exchange
@@ -46,7 +48,7 @@ class EDIStorageComponentMixin(AbstractComponent):
         :param filename: string for file name
         :return: PurePath object
         """
-        return self._dir_by_state(direction, state) / filename.strip("/ ")
+        return self._dir_by_state(direction, state, absolute) / filename.strip("/ ")
 
     def _get_remote_file(self, state, filename=None, binary=False):
         """Get file for current exchange_record in the given destination state.
--- a/edi_storage/components/receive.py
+++ b/edi_storage/components/receive.py
@@ -1,0 +1,27 @@
+# Copyright 2021 ForgeFlow S.L. (https://www.forgeflow.com)
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+from odoo.addons.component.core import Component
+
+
+class EDIStorageReceiveComponent(Component):
+
+    _name = "edi.storage.component.receive"
+    _inherit = [
+        "edi.component.receive.mixin",
+        "edi.storage.component.mixin",
+    ]
+    _usage = "storage.receive"
+
+    def receive(self):
+        checker = self.component(usage="storage.check")
+        result = checker.check()
+        if not result:
+            # all good here
+            return True
+
+        direction = self.exchange_record.direction
+        filename = self.exchange_record.exchange_filename
+        path = self._remote_file_path(direction, "pending", filename, absolute=True)
+        filedata = self.storage.get(path.as_posix(), binary=True)
+        return filedata
--- a/edi_storage/components/receive.py
+++ b/edi_storage/components/receive.py
@@ -22,6 +22,6 @@ class EDIStorageReceiveComponent(Component):
 
         direction = self.exchange_record.direction
         filename = self.exchange_record.exchange_filename
-        path = self._remote_file_path(direction, "pending", filename, absolute=True)
-        filedata = self.storage.get(path.as_posix(), binary=True)
+        path = self._remote_file_path(direction, "pending", filename)
+        filedata = self.storage.get(path.as_posix())
         return filedata
--- a/edi_storage/models/edi_backend.py
+++ b/edi_storage/models/edi_backend.py
@@ -125,6 +125,9 @@ class EDIBackend(models.Model):
         return [
             ("backend_type_id", "=", self.backend_type_id.id),
             ("direction", "=", "input"),
+            "|",
+            ("backend_id", "=", False),
+            ("backend_id", "=", self.id),
         ]
 
     def _storage_create_record_if_missing(self, exchange_type, remote_file_name):
--- a/edi_storage/models/edi_backend.py
+++ b/edi_storage/models/edi_backend.py
@@ -1,8 +1,13 @@
 # Copyright 2020 ACSONE SA
 # @author Simone Orsi <simahawk@gmail.com>
+# Copyright 2021 ForgeFlow S.L. (https://www.forgeflow.com)
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
 
+import logging
+
 from odoo import fields, models
+
+_logger = logging.getLogger(__name__)
 
 
 class EDIBackend(models.Model):
@@ -76,3 +81,47 @@ class EDIBackend(models.Model):
         return (
             1 if getattr(component_class, "_storage_backend_type", False) else 0,
         ) + res
+
+    def _cron_check_storage_pending_input(self, **kw):
+        for backend in self.filtered(lambda b: b.storage_id):
+            backend._check_storage_pending_input(**kw)
+
+    def _check_storage_pending_input(self, **kw):
+        self.ensure_one()
+        if not self.storage_id or not self.input_dir_pending:
+            _logger.info(
+                "%s ignored. No storage and/or input directory specified.", self.name
+            )
+            return False
+
+        pending_files = self.storage_id.list_files(self.input_dir_pending)
+        exchange_type = self.env["edi.exchange.type"].search(
+            self._domain_exchange_type_storage_pending_input()
+        )
+        if not len(exchange_type) == 1:
+            _logger.info("%s ignored. More than one exchange type found.", self.name)
+            return False
+        for file_name in pending_files:
+            existing = self.env["edi.exchange.record"].search(
+                [
+                    ("backend_id", "=", self.id),
+                    ("type_id", "=", exchange_type.id),
+                    ("exchange_filename", "=", file_name),
+                ]
+            )
+            if existing:
+                continue
+            self.create_record(
+                exchange_type.code, self._get_exchange_record_vals(file_name)
+            )
+            _logger.debug("%s: new exchange record generated.", self.name)
+        return True
+
+    def _domain_exchange_type_storage_pending_input(self):
+        return [
+            ("backend_type_id", "=", self.backend_type_id.id),
+            ("direction", "=", "input"),
+        ]
+
+    def _get_exchange_record_vals(self, file_name):
+        return {"exchange_filename": file_name, "edi_exchange_state": "input_pending"}
--- a/edi_storage/models/edi_backend.py
+++ b/edi_storage/models/edi_backend.py
@@ -148,12 +148,13 @@ class EDIBackend(models.Model):
             # If there is not patter, return everything
             return self.storage_id.list_files(self.input_dir_pending)
 
-        # TODO: validate pattern usage.
         bits = [exchange_type.exchange_filename_pattern]
         if exchange_type.exchange_file_ext:
-            bits.append("*." + exchange_type.exchange_file_ext)
+            bits.append(r"\." + exchange_type.exchange_file_ext)
         pattern = "".join(bits)
-        return self.storage_id.find_files(pattern, self.input_dir_pending)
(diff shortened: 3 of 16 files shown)


edi_storage/components/base.py
```diff
@@ -36,17 +36,19 @@ def _dir_by_state(self, direction, state):
         """
         assert direction in ("input", "output")
         assert state in ("pending", "done", "error")
+        if absolute:
+            return PurePath(self.backend[direction + "_dir_" + state] or "")
         return PurePath((self.backend[direction + "_dir_" + state] or "").strip(" /"))
```
LoisRForgeFlow: @simahawk I did not know exactly why you were stripping the path here, I use absolute paths in the edi backend configuration (input_dir_pending, input_dir_done...). Any clarification or recommendation?

etobella:
Might be possible to make a test? It looks great at first sight :wink:

edi_storage/components/base.py
simahawk: maybe just strip space from both sides and `/` only from the right?

edi_storage/models/edi_backend.py
```diff
@@ -76,3 +81,47 @@ def _component_sort_key(self, component_class):
         return (
             1 if getattr(component_class, "_storage_backend_type", False) else 0,
         ) + res
+
+    def _cron_check_storage_pending_input(self, **kw):
+        for backend in self.filtered(lambda b: b.storage_id):
```
simahawk: this way you loop twice: filter inside the loop

edi_storage/components/receive.py
```diff
@@ -0,0 +1,27 @@
...
+    def receive(self):
+        checker = self.component(usage="storage.check")
+        result = checker.check()
+        if not result:
+            # all good here
+            return True
+
+        direction = self.exchange_record.direction
+        filename = self.exchange_record.exchange_filename
+        path = self._remote_file_path(direction, "pending", filename, absolute=True)
```
simahawk: if you keep absolute flag it should come from the backend.

edi_storage/components/receive.py
```diff
@@ -0,0 +1,27 @@
...
+        filedata = self.storage.get(path.as_posix(), binary=True)
```
simahawk: binary is the default

edi_storage/models/edi_backend.py
```diff
@@ -76,3 +81,47 @@ def _component_sort_key(self, component_class):
...
+            backend._check_storage_pending_input(**kw)
+
+    def _check_storage_pending_input(self, **kw):
+        self.ensure_one()
+        if not self.storage_id or not self.input_dir_pending:
```
simahawk, line 97: I would filter this above

edi_storage/models/edi_backend.py
```diff
@@ -76,3 +81,47 @@ def _component_sort_key(self, component_class):
...
+            _logger.info(
+                "%s ignored. No storage and/or input directory specified.", self.name
+            )
+            return False
+
+        pending_files = self.storage_id.list_files(self.input_dir_pending)
+        exchange_type = self.env["edi.exchange.type"].search(
```
simahawk:
```suggestion
        exchange_types = self.env["edi.exchange.type"].search(
```

edi_storage/models/edi_backend.py
```diff
@@ -76,3 +81,47 @@ def _component_sort_key(self, component_class):
...
+            self._domain_exchange_type_storage_pending_input()
+        )
+        if not len(exchange_type) == 1:
```
simahawk: why? You could have multiple exchange types and you should check them all, no?

edi_storage/models/edi_backend.py
```diff
@@ -76,3 +81,47 @@ def _component_sort_key(self, component_class):
...
+            _logger.info("%s ignored. More than one exchange type found.", self.name)
+            return False
+        for file_name in pending_files:
+            existing = self.env["edi.exchange.record"].search(
```
simahawk: if you don't need the record:
```suggestion
            existing = self.env["edi.exchange.record"].search_count(
```

edi_storage/models/edi_backend.py
```diff
@@ -76,3 +81,47 @@ def _component_sort_key(self, component_class):
...
+                    ("exchange_filename", "=", file_name),
+                ]
+            )
+            if existing:
+                continue
+            self.create_record(
+                exchange_type.code, self._get_exchange_record_vals(file_name)
+            )
+            _logger.debug("%s: new exchange record generated.", self.name)
+        return True
```
simahawk, line 120: I'm not sure all this logic should stay here :thinking: 
I would be more in favor to use a specific component for this machinery.
Here we should only have the cron hook looking for exchange types to work with.
Also, consider that a specific backend type or exchange type could require a different logic.
Using a component would make this easier.

simahawk (changes requested):

edi_storage/models/edi_backend.py
LoisRForgeFlow: If I keep this small filter here I can add this method in a server action or button.

edi_storage/models/edi_backend.py
LoisRForgeFlow: As I expect a singleton below (and there is a check for that), it seems a more descriptible variable name if kept as singular.

edi_storage/models/edi_backend.py
LoisRForgeFlow: Because of this: https://github.com/OCA/edi/blob/13.0/edi_storage/models/edi_backend.py#L34. It is not currently supported to manage more than one exchange type per edi backend, when that case is covered the logic to find the exchange type could be extended, in the meanwhile I don't see any reason to make it more complex.

simahawk: if the issue is how to find files for a specific type only, we should use the filename pattern to filter the file list from the storage.
I've added this method https://github.com/OCA/storage/blob/13.0/storage_backend/models/storage_backend.py#L114 back in the days having this kind of feature in mind.

simahawk: > Because of this: https://github.com/OCA/edi/blob/13.0/edi_storage/models/edi_backend.py#L34. It is not currently supported to manage more than one exchange type per edi backend, when that case is covered the logic to find the exchange type could be extended, in the meanwhile I don't see any reason to make it more complex.

See my comment above :wink:

edi_storage/models/edi_backend.py
LoisRForgeFlow: you have to define at least a basic process. The same way it is done here: https://github.com/OCA/edi/blob/13.0/edi/models/edi_backend.py#L314. What can be different in this process for different backends other than the storage used?, which is already abstracted in the specific modules (I only use the high level method `list_files`).

If you still think this need to be moved to a component, can you help me with naming/structure? can you propose me names for these components?

edi_storage/models/edi_backend.py
LoisRForgeFlow: True, I removed the filter here and relied on what it is in `_check_storage_pending_input`.

edi_storage/components/receive.py
LoisRForgeFlow: Done!

edi_storage/models/edi_backend.py
LoisRForgeFlow: Done!

edi_storage/models/edi_backend.py
LoisRForgeFlow: But when you receive files you cannot assure that the name of the file will be always consistent with a structure defined in Odoo. Imagine you receive bills from your vendors, they will put whatever name to the file (most likely the bill number) and nothing that refers to a exchange type classification pattern. Don't you think?

simahawk: Why not? I mean, most of the time the bill number will have the same format likely w/ a prefix.
Or you can ask them to send files to you w/ a specific filename format.

On the type you can configure the [pattern](https://github.com/OCA/edi/blob/13.0/edi/models/edi_exchange_type.py#L42)  and the pattern can be even a path. Eg: `*/vendor_bills/*.xml`.
For output that field is used to generate the resulting filename but for input it can be used to find files.

Am I missing anything?

edi_storage/models/edi_backend.py
simahawk: fine, let's keep it simple. I opened this https://github.com/ForgeFlow/edi/pull/1 with some improvements (use file pattern, prefix all methods w/ `_storage`, use jobs, well... you'll see).
Let me know what you think!

edi_storage/components/base.py
LoisRForgeFlow: Done. no need to keep absolute now, thanks!

edi_storage/components/receive.py
LoisRForgeFlow: Removed

edi_storage/models/edi_backend.py
LoisRForgeFlow: Will review now, I just pushed some changes, hope we don't conflict :crossed_fingers:

LoisRForgeFlow:
@etobella @simahawk I added some tests just now.

edi_storage/models/edi_backend.py
LoisRForgeFlow: The bill number could have the same format for a given vendor, but it will not be consistent across vendors. Of course I can ask them to do that, but it is better if my software works without needing to bother my suppliers with that and to depend on their development teams to do the task. I prefer to not depend on the vendor's dev to deliver my project to the customer.

Do you think we can support both functionalities, so we cover wider use cases?

simahawk: @LoisRForgeFlow but I would never handle files from different suppliers w/ the same EDI backend. And if you use one backend per source then you can configure different paths per vendor.
Do you see any problem with this?

edi_storage/models/edi_backend.py
```diff
@@ -76,3 +82,79 @@ def _component_sort_key(self, component_class):
...
+            # We should probably run this code in a separate job per exchange type.
+            file_names = self._storage_get_input_filenames(exchange_type)
+            _logger.info(
+                "Processing exchange type '%s': found %s files to process",
+                exchange_type.display_name,
+                len(file_names),
+            )
+            for file_name in file_names:
+                # TODO: add config for job
+                self.with_delay()._storage_create_record_if_missing(
```
simahawk, line 117: can you add a job config for this?

edi_storage/models/edi_backend.py
```diff
@@ -76,3 +82,79 @@ def _component_sort_key(self, component_class):
...
+            return
+        record = self.create_record(
+            exchange_type.code, self._storage_new_exchange_record_vals(file_name)
+        )
+        _logger.debug("%s: new exchange record generated.", self.name)
+        return record.identifier
+
+    def _storage_get_input_filenames(self, exchange_type):
+        if not exchange_type.exchange_filename_pattern:
+            # If there is not patter, return everything
```
simahawk, line 151: sounds like a good compromise :)

edi_storage/models/edi_backend.py
LoisRForgeFlow: My initial idea is to have one backend shared for several vendors if they use the same data structure (say UBL), and some other backends vendor specific. Now that you highlight it, it is true that I would need a folder per vendor in the shared backends... so seems that your proposal makes more sense.

I have holidays until next week. I will check in detail your proposal using pattern next week.

Thanks for the review and help :wink:

simahawk: Same data structure: this is why we have backend types :wink: 
You can have the same exchange types reusable by backend type.
And you can tie exchange types to specific backend only if you need.

simahawk: >Thanks for the review and help wink

Thanks for your proposal and for accepting my suggestions :wink:

edi_storage/models/edi_backend.py
```diff
@@ -76,3 +82,79 @@ def _component_sort_key(self, component_class):
...
+            return self.storage_id.list_files(self.input_dir_pending)
+
+        # TODO: validate pattern usage.
+        bits = [exchange_type.exchange_filename_pattern]
+        if exchange_type.exchange_file_ext:
+            bits.append("*." + exchange_type.exchange_file_ext)
+        pattern = "".join(bits)
+        return self.storage_id.find_files(pattern, self.input_dir_pending)
```
LoisRForgeFlow: @simahawk  I think, we cannot really use the pattern here. To convert the pattern we need a exchange record (see https://github.com/OCA/edi/blob/13.0/edi/models/edi_exchange_type.py#L134), but here were about to create it so we don't have one.

The maximun we could do is something like this:

EDIT: See last commit (https://github.com/OCA/edi/commit/4cda6390d9cbf36d05506b1ba3cd8857b8843967)

 What do you think? is there anything I'm missing?

edi_storage/models/edi_backend.py
LoisRForgeFlow: What do you expect here? there is no config in all the other queue jobs on edi module.

edi_storage/models/edi_backend.py
LoisRForgeFlow: This is my proposal: https://github.com/OCA/edi/pull/402/commits/4cda6390d9cbf36d05506b1ba3cd8857b8843967. Let me know if you see possible improvements :)

edi_storage/models/edi_backend.py
LoisRForgeFlow: Great! :raised_hands:

edi_storage/models/edi_backend.py
simahawk: > @simahawk I think, we cannot really use the pattern here. To convert the pattern we need a exchange record (see https://github.com/OCA/edi/blob/13.0/edi/models/edi_exchange_type.py#L134), but here were about to create it so we don't have one.
> 
> The maximun we could do is something like this:

I don't think that's completely correct. The pattern is in your control. You decide if you need the exchange record or not.

And in this case, you don't care about `exchange_filename` name because you are going to set it when creating the record (it's computed but write-able).

So, what you put in the pattern does not have to satisfy `_make_exchange_filename` because that method is called only if the filename is not set.

I'd rollback this change, unless I'm missing something else.

edi_storage/models/edi_backend.py
simahawk: that's the point: I've simply drafted the delayed call but we need a `queue.job.function` record to use it, which should be added to `edi_storage` :wink: 
Sorry for not making it clear at first.

edi_storage/models/edi_backend.py
LoisRForgeFlow: So, then what do you expect users to enter in the pattern field? a regex? 

To make things more clear and explicit, confirm if the following is correct: your idea is to use the `exchange_filename_pattern` in two different ways:

- to generate the exchange filename in and output type. pattern is to use available variables between brackets, e.g: `{exchange_record}-{type}-{dt}`.
- to find/filter files to be imported from an storage for the input exchange type. In this case using a regex (because that's what expect `find_files` method, e.g: `some-type.*\.xml`

Please, correct me if I'm misunderstanding your proposal.

edi_storage/models/edi_backend.py
LoisRForgeFlow: Got it. Now it is done, thanks!

edi_storage/models/edi_backend.py
simahawk: This! :smiley: 
We could add an help text in the form maybe when edi_storage is installed

LoisRForgeFlow: Awesome, done in https://github.com/OCA/edi/pull/402/commits/90e00436b1dc57682f3617904d30084f4fd3b59b.

LoisRForgeFlow:
@simahawk I think this is ready now, do you miss anything?

simahawk:
@LoisRForgeFlow seems good. Thanks for your work!

Just one detail: you could/should skip version from the commit message as when you fwd/backport such commits "13.0" becomes irrelevant and sometimes misleading too.

Also try to keep commit message subject short and move extra info to long description. :wink: 

Refs: https://www.odoo.com/documentation/13.0/developer/reference/guidelines.html#commit-message-full-description and
https://chris.beams.io/posts/git-commit/

Not blockers, only good to have!

simahawk (approved):
@etobella would you mind to review?

edi_storage/models/edi_exchange_type.py
```diff
@@ -0,0 +1,20 @@
...
+    # Extend help to explain new usage.
+    exchange_filename_pattern = fields.Char(
+        help="For output exchange types this should be a formatting string "
+        "with the following variables available (to be used between "
+        "brackets, `{}`): `exchange_record`, `record_name`, `type` and "
+        "`dt`. For instance, a valid string would be "
+        "{record_name}-{type.code}-{dt}\n"
+        "For input exchange types it should be a regex expresion to filter "
+        "the files to be fetched from the pending directory in the related "
+        "storage. E.g: `.*my-type-[0-9]*.\\.csv`"
```
simahawk:
```suggestion
        "For input exchange types related to storage backends "
        "it should be a regex expression to filter "
        "the files to be fetched from the pending directory in the related "
        "storage. E.g: `.*my-type-[0-9]*.\\.csv`"
```

LoisRForgeFlow:
@simahawk Sure, I tend to prefer to know from which version the initial change came, but you are right that is better to follow the guidelines. I will remove the version number when forward porting the changes.

Thanks for the time spent reviewing!

etobella (approved):

PR #403, opened 2021-06-22
TEST branch 12

bealdav:


bealdav:
It seems the fix is there https://github.com/OCA/edi/pull/415 thanks to @hhgabelgaard

hhgabelgaard:
@bealdav Yes, I believe the solution is to use the previous version.
But my attempts to get 0.3.5 in #415 fails....

bealdav:
Problably, here's my fail "odoo.exceptions.UserError: ('Unable to install module "account_invoice_import_invoice2data" because an external dependency is not met: invalid syntax', '')"

hhgabelgaard:
The root cause is here:
File "/home/travis/virtualenv/python3.5.10/lib/python3.5/site-packages/invoice2data/extract/plugins/tables.py", line 39
    logger.debug(f"Regex pattern = {table['body']}")

That line is in v0.3.6 
In requirements.txt I try to get 0.3.5 which is without this line - But apparently I still get the newest version

PR #405, opened 2021-06-28
[FIX] edi_voxel_stock_picking: Fix tests after change date format

carlosdauden:
Adapt test after https://github.com/OCA/edi/commit/68f2357a4f10226ecc264de74932f1040d1b750d
@Tecnativa


pedrobaeza:
It seems it's failing the contrary, isn't it? I remember to pass tests before committing. Do you know what is happening? I'm going to investigate.

pedrobaeza:
Found the problem, as the `secondary_unit` extension was also putting such data. If it's green, I'm fast-tracking.

PR #349, opened 2021-03-12
[13.0][IMP] Add sale order ubl parse latest delivery

TDu:
Parse the latest delivery date and time from the delivery node of an UBL
sales order.
And assign the value to the `commitment_date` of the sales order.

PR #316, opened 2021-01-09
[MIG] sale_order_import_ubl: Migration to 14.0

phuctranfxvn:
- Standard migration.
- Requires modules:
  + `sale_order_import` migrated in https://github.com/OCA/edi/pull/302
  + `base_ubl` migrated in https://github.com/OCA/edi/pull/303

PR #406, opened 2021-07-05
[FW][14.0][IMP] Add sale order ubl parse latest delivery

hailangvn:
This is the forward update from PR https://github.com/OCA/edi/pull/349.

Parse the latest delivery date and time from the delivery node of an UBL
sales order.
And assign the value to the `commitment_date` of the sales order.

### DIFF
--- a/base_ubl/models/ubl.py
+++ b/base_ubl/models/ubl.py
@@ -724,6 +724,18 @@ class BaseUbl(models.AbstractModel):
                 )
         return partner_dict
 
+    @api.model
+    def ubl_parse_delivery_details(self, delivery_node, ns):
+        delivery_dict = {}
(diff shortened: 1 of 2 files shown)


simahawk:
I missed this, thanks bot :)

simahawk (commented):
@hailangvn pls do not change the commit msg. It shouldn't been merged w/ the `13.0` in it but rewriting it will just be a source of issues when fwd/back porting.

hailangvn:
Thanks @simahawk for your note. I was new to migration work at the time. I have just found out that the commit has been merged with current title by [PR#316](https://github.com/OCA/edi/pull/316). So this PR is not needed anymore.

PR #408, opened 2021-07-05
[13.0][IMP] edi_storage: add listener to move files to done/error dirs

LoisRForgeFlow:
When a file is processed the listener will move it to the done directory or the error directory based on the result of processing the file.

cc @simahawk 

@ForgeFlow

### DIFF
--- a/edi_storage/components/listener.py
+++ b/edi_storage/components/listener.py
@@ -1,0 +1,60 @@
+# Copyright 2021 ForgeFlow S.L. (https://www.forgeflow.com)
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+import functools
+from pathlib import PurePath
+
+from odoo.addons.component.core import Component
+
+
+class EdiStorageListener(Component):
+    _name = "edi.storage.component.listener"
+    _inherit = "base.event.listener"
+
+    def _move_file(self, storage, from_dir_str, to_dir_str, filename):
+        from_dir = PurePath(from_dir_str)
+        to_dir = PurePath(to_dir_str)
+        if filename not in storage.list_files(from_dir):
+            # The file might have been moved after a previous error.
+            return False
+        self._add_after_commit_hook(
+            storage._move_files, [(from_dir / filename).as_posix()], to_dir.as_posix()
+        )
+        return True
+
+    def _add_after_commit_hook(self, move_func, sftp_filepath, sftp_destination_path):
+        """Add hook after commit to move the file when transaction is over.
+        """
+        self.env.cr.after(
+            "commit",
+            functools.partial(move_func, sftp_filepath, sftp_destination_path),
+        )
+
+    def on_edi_exchange_done(self, record):
+        storage = record.backend_id.storage_id
+        res = False
+        if record.direction == "input" and storage:
+            file = record.exchange_filename
+            pending_dir = record.backend_id.input_dir_pending
+            done_dir = record.backend_id.input_dir_done
+            error_dir = record.backend_id.input_dir_error
+            if not done_dir:
+                return res
+            res = self._move_file(storage, pending_dir, done_dir, file)
+            if not res:
+                # If a file previously failed it should have been previously
+                # moved to the error dir, therefore it is not present in the
+                # pending dir and we need to retry from error dir.
+                res = self._move_file(storage, error_dir, done_dir, file)
+        return res
+
+    def on_edi_exchange_error(self, record):
+        storage = record.backend_id.storage_id
+        res = False
+        if record.direction == "input" and storage:
+            file = record.exchange_filename
+            pending_dir = record.backend_id.input_dir_pending
+            error_dir = record.backend_id.input_dir_error
+            if error_dir:
+                res = self._move_file(storage, pending_dir, error_dir, file)
+        return res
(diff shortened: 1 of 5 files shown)


edi_storage/components/listener.py
```diff
@@ -0,0 +1,46 @@
...
+
+class EdiStorageListener(Component):
+    _name = "edi.storage.component.listener"
+    _inherit = "base.event.listener"
+
+    def _move_file(self, storage, from_dir_str, to_dir_str, filename):
+        from_dir = PurePath(from_dir_str)
+        to_dir = PurePath(to_dir_str)
+        try:
+            storage.move_files([(from_dir / filename).as_posix()], to_dir.as_posix())
```
simahawk: it would better to do this after the transaction is done. 
Please find an example here https://github.com/OCA/connector-interfaces/pull/74/files#diff-60319609653ac5e93d1758d594b36b9274e778933247e17ef51799b7a28a341bR61 (BTW if you have time to drop a review there it would be nice :wink: )

edi_storage/components/listener.py
```diff
@@ -0,0 +1,46 @@
...
+        storage = record.backend_id.storage_id
+        res = False
+        if record.direction == "input" and storage:
+            file = record.exchange_filename
+            pending_dir = record.backend_id.input_dir_pending
+            done_dir = record.backend_id.input_dir_done
+            error_dir = record.backend_id.input_dir_error
+            res = self._move_file(storage, pending_dir, done_dir, file)
+            if not res:
+                # Retry from error dir.
```
simahawk: why?

simahawk (changes requested):
pls add test

edi_storage/components/listener.py
LoisRForgeFlow: Just extended the comment to explain.

edi_storage/components/listener.py
LoisRForgeFlow: Interesting :thinking: I will need to check if this is compatible with trying twice to move files, first from pending and then from error. 

BTW, tests added.

LoisRForgeFlow: Done, the changes are here: https://github.com/OCA/edi/pull/408/commits/efc487e8d5bd39f87589bd0e89ad2705c8bc0d59.

edi_storage/components/listener.py
```diff
@@ -0,0 +1,58 @@
...
+        storage = record.backend_id.storage_id
+        res = False
+        if record.direction == "input" and storage:
+            file = record.exchange_filename
+            pending_dir = record.backend_id.input_dir_pending
+            done_dir = record.backend_id.input_dir_done
+            error_dir = record.backend_id.input_dir_error
+            res = self._move_file(storage, pending_dir, done_dir, file)
+            if not res:
+                # If a file previously failed it should have been previously
```
simahawk, line 45: Question: if it landed in the error dir it means that corrections were needed. Hence, I would expect to be fixed and move to pending, no?

simahawk (approved):
LG, thanks for accepting my suggestions once again :)

edi_storage/components/listener.py
simahawk: could you drop a review there when you time? :pray:

edi_storage/components/listener.py
LoisRForgeFlow: But the file is already fetched and in the file content of exchange record. Therefore it can happen that the processing initially failed due to a misconfiguration in any place in Odoo (a tax wrongly encoded for saying something), and after fixing the configuration in Odoo [the next time that exchange record is processed](https://github.com/OCA/edi/blob/13.0/edi/models/edi_backend.py#L550) it will succeed because the file itself did not need any correction. In that situation the file is in the error folder and need to be moved to the done folder.

edi_storage/components/listener.py
LoisRForgeFlow: It is a big PR but I will try to review it this week! :smile:

edi_storage/components/listener.py
```diff
@@ -0,0 +1,58 @@
...
+        )
+
+    def on_edi_exchange_done(self, record):
+        storage = record.backend_id.storage_id
+        res = False
+        if record.direction == "input" and storage:
+            file = record.exchange_filename
+            pending_dir = record.backend_id.input_dir_pending
+            done_dir = record.backend_id.input_dir_done
+            error_dir = record.backend_id.input_dir_error
```
etobella, line 40: I miss a 
```python
if not done_dir: 
    return
```

etobella (commented):
Just a small comment

edi_storage/components/listener.py
simahawk: I see. Thanks for the explanation :+1:

simahawk (approved):

edi_storage/components/listener.py
simahawk: actually part of it is from another PR that should be reviewed first. BTW that PR adds kind of the same concept we have here in EDI module to allow advanced configuration for components.

edi_storage/components/listener.py
LoisRForgeFlow: Done!

LoisRForgeFlow:
Rebased and autosquashed!

PR #409, opened 2021-07-08
[13.0][FIX] account_invoice_ubl: do not force report rendering

LoisRForgeFlow:
There is no need for this because there exists a conext in
standard (`force_report_rendering`) for that purpose and that
is correctly set in the tests of this module.

Before this change it was forced for any of pdf report
related to this module (basically invoice reports). This
could make the tests of other addons installed in the database
and not depending on this module fail.

@ForgeFlow

### DIFF
--- a/account_invoice_ubl/models/ir_actions_report.py
+++ b/account_invoice_ubl/models/ir_actions_report.py
@@ -28,16 +28,6 @@ class IrActionsReport(models.Model):
                     pdf_content = invoice.embed_ubl_xml_in_pdf(pdf_content)
         return pdf_content
 
-    def render_qweb_pdf(self, res_ids=None, data=None):
-        """This is only necessary when tests are enabled.
-        It forces the creation of pdf instead of html."""
-        if isinstance(res_ids, int):
-            res_ids = [res_ids]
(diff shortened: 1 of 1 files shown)


account_invoice_ubl/models/ir_actions_report.py
```diff
@@ -29,13 +29,8 @@ def _post_pdf(self, save_in_attachment, pdf_content=None, res_ids=None):
         return pdf_content
 
     def render_qweb_pdf(self, res_ids=None, data=None):
```
etobella: With this change, you need this function?

LoisRForgeFlow: It is an standard function, used to print any qweb report

LoisRForgeFlow: Actually, you are right, we can get rid of all the function here.

JordiBForgeFlow (approved):

AaronHForgeFlow (approved):

etobella (approved):
Thanks, this should be better :wink:

AdriaGForgeFlow (approved):

PR #411, opened 2021-07-09
[13.0] Add edi_gs1 + edi_gs1_stock

acsonefho:


agent-z28:
NICE

agent-z28:
@acsonefho i think the edi_gs1 module should only extend the parner models etc, and be morge generic. 
If we only want edi_gs1_edifact integration all the XML stuff will be unwanted. but the gln is needed, and so on

simahawk:
I think we should merge it as it is and refactor things out later if needed.

simahawk:
of course, we need reviews to merge it :wink:

agent-z28:
@simahawk it's part of my review, that i implement another edi_gs1 standard, and the edi_gs1 here is not compatible. so i cannot use it like this. the module should only contain edi_gs1 related things. the xml implementatin should be edi_gs1_xmlstandardname while i do edi_gs1_edifact_*.

simahawk:
what I'm saying is that if what we have here works it would be nice to merge it and then you can propose any change you think it would make it better. 
So, could you please drop a review for what it is today and just ask to add a line on the roadmap to split out the xml part from edi_gs1? It will be easier for everybody and for us especially to review the changes.
Could it work for you?

agent-z28:
@simahawk i understand that it would be easier, but afterwards, you need migration scripts and it'll create much more trouble as splitting it directly. i can add pull request with edi_gs1 modul on its own?

simahawk:
> @simahawk i understand that it would be easier, but afterwards, you need migration scripts and it'll create much more trouble as splitting it directly. i can add pull request with edi_gs1 modul on its own?

sure, feel free :)

etobella:
@agenterp Any news on your proposal? If it is not possible to have the new changes, we can start with this version and then propose the improvements.

agent-z28:
@etobella i'm still working on this issue internally. will be done in the middle of next week. But starting with a module name, where we later need migrations, and we already know it, seems not be a good approach. another general question, why has the edi module be renamed to edi_oca? i thought company or organisation names doesn't belong to module names.

etobella:
`edi` was taken by odoo and was not published on odoo apps (but there was no module...). At the end, we decided to change the name in order to erase problems. (renaming is done by OpenUpgrade, so not a problem on migration)

About migrations. it is not really a problem for me, at the end, a lot of migrations are made on version change every year or even in the same version, so I don't see the issue there

agent-z28:
@etobella @simahawk  okay if you dont see a problem in migration later, i will not block it, i just wanted to mention, that i already have the problem. but if it's easy to fix it later lets go

bealdav:
Hi sorry for my noob question but I would like to know if you plan or need a future plan to interact with these features
https://github.com/OCA/stock-logistics-barcode/pull/387/files 
or maybe be not because I'm really noob on this matter :-)
Maybe at least base_gs1_barcode ?
Sorry for my confuse question

simahawk:
@bealdav on my side no plan but I'm not working on this subject since a while. Would be nice to do it tho. Could you create an issue for this?
FTR regarding this PR, I still have to get back to it and clean it up.
I started it here https://github.com/OCA/edi/pull/449 but got no time to move it fwd.

bosd:
Has  this been  included in a different pr?

acsonefho:
> Has this been included in a different pr?

Unfortunately not :/

simahawk:
I'll try to have a look at this by the end of the year, even if is a bit hard because I'm not working on v13 since a while.
Plus, there are a lot of changes that should be backported from 14 that might be tricky to integrate.

bosd:
Any update on this :shy:?

acsonefho:
> Any update on this :shy:?

Probably (and certainly) some improvements must be done.
Already in production for almost 2 years and working perfectly.
So no update because "it's working" :muscle:

bosd:
You're still running some install on V13?
I have some v14 in production and feel like it's time to migrate..

PR #415, opened 2021-07-11
[FiX][12.0] Need to pin invoice2data to 0.3.5

hhgabelgaard:
Newer versions use f-formats which fails on python 3.5

### DIFF
--- a/account_invoice_import_invoice2data/__manifest__.py
+++ b/account_invoice_import_invoice2data/__manifest__.py
@@ -11,7 +11,7 @@
     'author': 'Akretion,Odoo Community Association (OCA)',
     'website': 'https://github.com/OCA/edi',
     'depends': ['account_invoice_import'],
-    'external_dependencies': {'python': ['invoice2data']},
+    'external_dependencies': {'python': ['invoice2data==0.3.5']},
     'data': ['wizard/account_invoice_import_view.xml'],
     'demo': ['demo/demo_data.xml'],
     'installable': True,
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
 PyPDF2
 phonenumbers
-invoice2data
+invoice2data==0.3.5
 factur-x
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
 PyPDF2
 phonenumbers
-git+https://github.com/invoice-x/invoice2data@v0.3.5
+invoice2data==0.3.5
 factur-x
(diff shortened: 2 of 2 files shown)


gdgellatly:
Why not just `invoice2data==0.3.5` - that's what we did in prod. Seemed to work.

hhgabelgaard:
@gdgellatly Yes, so did we
But https://github.com/OCA/edi/pull/415/commits/29524ab2effda1e106378bfe453708c67415a809 failed on travis
Let's give it another try

hhgabelgaard:
This fails with:

`File "/home/travis/virtualenv/python3.5.10/lib/python3.5/site-packages/invoice2data/extract/plugins/tables.py", line 39
    logger.debug(f"Regex pattern = {table['body']}")`

But version 0.3.5 of invoice2data contains this in line 39:
`            if not line.strip('').strip('\n') or not line:
                continue`
From https://github.com/invoice-x/invoice2data/blob/v0.3.5/src/invoice2data/extract/plugins/tables.py#L39

So a newer invoice2data must be installed somewhere else....

etobella:
Put the change in https://github.com/OCA/edi/blob/12.0/account_invoice_import_invoice2data/__manifest__.py#L14

It is using pip on installation

etobella:
Travis did not launch. Can you squash the commits?

account_invoice_import_invoice2data/__manifest__.py
```diff
@@ -11,7 +11,7 @@
     'author': 'Akretion,Odoo Community Association (OCA)',
     'website': 'https://github.com/OCA/edi',
     'depends': ['account_invoice_import'],
-    'external_dependencies': {'python': ['invoice2data']},
+    'external_dependencies': {'python': ['invoice2data==0.3.5']},
```
etobella, line 14: Runbot is not liking this, can you set it to invoice2data and keep the change only on requirements?

etobella: the problem seems that runbot and travis are working differently :dizzy_face:

hhgabelgaard: > the problem seems that runbot and travis are working differently

Yes, that is exactly the problem.
If we have ` 'external_dependencies': {'python': ['invoice2data']},` runbot works, but travis install the newest invoice2data, and then it fails on python 3.6 f-strings....

The only fix is probaly to get runbot and travis (MQT) to have the same understanding of external_dependencies and requirements

i-vyshnevska (approved):

PR #416, opened 2021-07-13
[13.0][IMP] base_ubl: add _ubl_get_seller_code_from_product method

FerranCforgeFlow:
Added __ubl_get_seller_code_from_product_ method to be inherited in case a custom product code is used instead of the Odoo product.default_code in the UBL cac:SellersItemIdentification field.

PR #417, opened 2021-07-15
[14.0][IMP] base_ubl: add _ubl_get_seller_code_from_product

FerranCforgeFlow:
Forward port https://github.com/OCA/edi/pull/416 @bealdav


LoisRForgeFlow (approved):
Thanks

etobella (approved):
Just as a comment, you don't need to include the version on the commit name, as it is usually forwarded and it looses the sense :wink:

PR #421, opened 2021-07-26
[13.0][IMP] account_invoice_ubl & base_ubl: add hooks

FerranCforgeFlow:
- Added the possibility to add a cac:BuyersItemIdentification ubl field if a customer product code exists.
- Added __ubl_get_customer_product_code_ method to be inherited in case a custom customer product code is used in the cac:BuyersItemIdentification ubl field.
- Added __ubl_get_invoice_type_code_ method to be inherited in case a custom invoice code is used instead of the 380/381 default ones.
- Added __ubl_get_customer_assigned_id_ method to be inherited in case a custom customer reference is used in the cbc:CustomerAssignedAccountID ubl field.
- Added __ubl_get_order_reference_ method to be inherited to fill the cbc:ID of the cac:OrderReference ubl field.

PR #423, opened 2021-07-30
[13.0][IMP] base_ubl: add customer as an argument in _ubl_add_item method

FerranCforgeFlow:
- Add _customer_ as an argument in the __ubl_add_item_ method the same way as the seller, as default set to false
- Add also _customer_ as an argument in the __ubl_get_customer_product_code_ so the product code can be taken either from the customer or from the product
- Fixed: the current version of the _ubl_add_supplier_party and _ubl_add_customer_party was not allowing to add the cbc:CustomerAssignedID and the cbc:SupplierAssignedID respectively, if the partner for which we want to build the party block was our company.The if not company condition has been removed, and a call to the _ubl_get_customer_assigned_id() method is first added, so if it does not return False, the UBL field is added.

### DIFF
--- a/base_ubl/models/ubl.py
+++ b/base_ubl/models/ubl.py
@@ -197,6 +197,9 @@ class BaseUbl(models.AbstractModel):
             )
         self._ubl_add_contact(partner, party, ns, version=version)
 
+    def _ubl_get_customer_assigned_id(self, partner):
+        return partner.commercial_partner_id.ref
+
     @api.model
     def _ubl_add_customer_party(
         self, partner, company, node_name, parent_node, ns, version="2.1"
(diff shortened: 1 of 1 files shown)


LoisRForgeFlow (approved):

FerranCforgeFlow:
@bealdav could you check?

bealdav:
Hi thanks @FerranCforgeFlow , it seems there are already merged code here ?
like in https://github.com/OCA/edi/pull/421/files

Maybe a rebase could only let real updates. Or I missed something ? Thanks

FerranCforgeFlow:
In this PR I am fixing the supplier and customer party logic which is not in the #421. However, I am adding one more argument to the _ubl_get_customer_product_code method I added in the #421, so I am not adding the whole method again. I am also changing the location of  the _ubl_get_customer_assigned_id method that I added in #421 since in my view it makes more sense to be defined before two methods that use it instead of between them.

bealdav:
OK I'll dive into. Thanks a lot

FerranCforgeFlow:
What's your view @bealdav ?

bealdav (approved):

PR #424, opened 2021-07-30
[14.0][FIX] webservice: server.env.mixin needs to be inherited

LoisRForgeFlow:
The funtionality is used by the module as the method
`_server_env_fields` is defined, but the inherit of the mixin
using it is missing.

Forward port of https://github.com/OCA/edi/pull/401.

@ForgeFlow

### DIFF
--- a/webservice/models/webservice_backend.py
+++ b/webservice/models/webservice_backend.py
@@ -7,7 +7,7 @@ from odoo import fields, models
 class WebserviceBackend(models.Model):
 
     _name = "webservice.backend"
-    _inherit = ["collection.base", "server.env.techname.mixin"]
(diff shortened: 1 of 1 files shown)


etobella (approved):

webservice/models/webservice_backend.py
```diff
@@ -7,7 +7,7 @@
 class WebserviceBackend(models.Model):
 
     _name = "webservice.backend"
-    _inherit = ["collection.base", "server.env.techname.mixin"]
+    _inherit = ["collection.base", "server.env.techname.mixin", "server.env.mixin"]
```
simahawk, line 10: FTR we are going to change this need ;)

simahawk (approved):

PR #425, opened 2021-08-02
[13.0][IMP] edi: be able to link a backend to a partner

LoisRForgeFlow:
This can be usuful in some situation in which a backend is
restricted to only a given partner.

@ForgeFlow

cc @simahawk @etobella


simahawk (changes requested):
are we sure it should stay in core module and not in `edi_backend_partner`?

LoisRForgeFlow:
@simahawk I thought that too, but I wanted to see your opinion. I will move to a new module soon!

LoisRForgeFlow:
Closed in favor of https://github.com/OCA/edi/pull/428.

PR #426, opened 2021-08-04
[13.0][FIX] edi_storage: remove duplicate backend_type_id field in edi backend form

JordiMForgeFlow:
The field backend_type_id is already added in the base edi_backend_view_form inside the config group, so adding it here duplicates it in the view.


LoisRForgeFlow (approved):
Finally! :laughing:

AaronHForgeFlow (approved):

PR #428, opened 2021-08-06
[13.0][ADD] edi_backend_partner

LoisRForgeFlow:
This module add the a "Partner" field in EDI backend, this can be useful in
some situation in which a backend is restricted to only a given partner or
to have another criteria to search edi backends.

@ForgeFlow


edi_backend_partner/models/res_partner.py
```diff
@@ -0,0 +1,29 @@
...
+class Partner(models.Model):
+    _inherit = "res.partner"
+
+    edi_backend_ids = fields.One2many(
+        comodel_name="edi.backend", inverse_name="partner_id",
+    )
+    edi_backend_count = fields.Integer(compute="_compute_edi_backend_count")
+
+    def _compute_edi_backend_count(self):
+        for rec in self:
```
simahawk, line 16: if you expect this to run a large recordset it would be better to use read_group

simahawk (approved):

edi_backend_partner/models/res_partner.py
LoisRForgeFlow: actually this field is only used in the form view, so it will be computed one by one.

JordiMForgeFlow (approved):
Code + functional review 👍🏼

LoisRForgeFlow:
@simahawk Good to merge?

PR #429, opened 2021-08-10
[13.0][IMP] base_ubl: hook to add the AdditionalAccountID ubl field

FerranCforgeFlow:
Added a _ubl_get_additional_reference method to be able to add the AdditionalAccountID ubl field in the _ubl_add_supplier_party

### DIFF
--- a/base_ubl/models/ubl.py
+++ b/base_ubl/models/ubl.py
@@ -232,6 +232,7 @@ class BaseUbl(models.AbstractModel):
                 node_name="AccountingContact",
                 version=version,
             )
+        return customer_party_root
 
     @api.model
     def _ubl_add_supplier_party(
@@ -269,6 +270,7 @@ class BaseUbl(models.AbstractModel):
         self._ubl_add_party(
             partner, company, "Party", supplier_party_root, ns, version=version
         )
+        return supplier_party_root
 
     @api.model
     def _ubl_add_delivery(self, delivery_partner, parent_node, ns, version="2.1"):


FerranCforgeFlow:
@bealdav could you check?

RaulOForgeFlow (approved):
:+1:

FerranCforgeFlow:
@bealdav could you please check?

etobella (approved):

albariera (approved):
LGTM :+1:

bealdav (approved):
Ok, thanks
I ask myself if there is no other way to add new misc nodes dynamically (method with parameter).
It seems there many conditional additions in the code now.
But if we need to do smth, it could be in an other PR I suppose

etobella:
Wouldn't be easier to return the node `customer_party_root` and anyone can do its specific changes if needed?? (For example for the partner reference or the additional Account Id)

FerranCforgeFlow:
What you are saying is actually a more solid option. I opted for the alternative of adding generic methods that people can inherit if wanted.

FerranCforgeFlow:
Could it be merged?

etobella:
Might be possible to apply the proposed change? This way it is easier and all this kind of hooks are unnecessary.

FerranCforgeFlow:
Taking a look at the code, it is better to return the nodes as you said to allow customization. I have already changed it in both, customer and supplier methods. Thanks for the advice!

etobella (approved):

LoisRForgeFlow:
@etobella could you try again the merge?

PR #432, opened 2021-08-14
[12.0] account_invoice_import_simple_pdf (invoice2data alternative)

alexis-via:
This is a backport from v14 to v12 of the PR https://github.com/OCA/edi/pull/420

PR #433, opened 2021-08-25
[ADD] edi_pdf2data

etobella:
This module allows to define templates for importing documents using the library invoice2data. It can be configured in order to import invoices and other kinds of documents


edi_pdf2data/components/edi_input_pdf2data_process.py
```diff
@@ -0,0 +1,29 @@
+# Copyright 2021 Creu Blanca
+# @author: Enric Tobella
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo.addons.component.core import Component
+
+
+class EdiInputPd2DataProcess(Component):
```
albariera: Typo error. The f of PDF is left :wink:

pedrobaeza:
Isn't this the same in features as #432 ?

etobella:
#432 intends to remove the invoice2data dependancy, and it is only related to invoices, this can be used in a generic way for all kind of documents, for example, we are trying to import quality reports from a laboratory.

pedrobaeza:
OK, but do you need the dependency to invoice2data or can you do something similar as the other module?

etobella:
We could check how to remove the dependency, let me think about that.

However, why do you want to remove the dependency?

pedrobaeza:
For not depending on external libraries and possible frictions with versions. And I also remember a not good pipeline flow with the library.

etobella:
Ok, thanks, I will review which is the best solution!! Thanks :smile:

FernandoRomera:
> For not depending on external libraries and possible frictions with versions. And I also remember a not good pipeline flow with the library.

#432 has three externs libraries, https://github.com/OCA/edi/blob/26892422e3dd19e784ae5718f9831d5823d75678/account_invoice_import_simple_pdf/readme/INSTALL.rst

gdgellatly:
@pedrobaeza personally I think invoice2data is the less harmful dependency than the dependencies of #432. I have spent quite some time with invoice2data (2+ years), pymupdf (again 2+years) and #432 (in test and eval) and this is much more in line with my ideal than the existing invoice2data_pdf module, the new simple_pdf proposal and actually my own work I am porting to v14. 

Multiline support for a start is a key differentiator. OCR processing too although more fringe. My main issue with the 2 existing ones coming from invoice import is the ignorance of purchase orders in a stock environment. 

The main benefits of simple_pdf proposal is the fields interface and I think in future that idea could be generalised for the basic fields. Also for non-Europeans, the OCA dependencies are annoying for simple_pdf. We have to install all sorts of Euro specific data files. This is nice and clean.

But honestly, PyMuPDF is great when it works, and admittedly it is being used in a rather simple use case here (EDIT: being #432, not actually here), but it is a dependency you'd live without if you could IMO. But things like dynamic form filling on printed pdf forms it is pretty much all there is that works.

pedrobaeza:
OK, it was just a suggestion as it seems working better according to @alexis-via

gdgellatly:
@pedrobaeza Yes of course, until it doesn't. In practical use, simple_pdf is most suited to regular bills, coded to 1 line. If you use operating-unit, or you use purchase orders and stock, forget it, if you need to fill the origin field, forget it. For me they are completely different use cases for different kinds of businesses/supplier invoices.

As for pipeline issues, yes there were some problematic commits and release tags on invoice2data. A typo and then a py 3.6 f-string. In fact I was just fixing a v12 server for precisely that.

alexis-via:
@gdgellatly What do you mean by "Also for non-Europeans, the OCA dependencies are annoying for simple_pdf." ? Which  dependencies are annoying for you? I don't understand that part of your comment.

gdgellatly:
@alexis-via UNECE stuff of base business document import.

```
"account_tax_unece",
        "uom_unece",
```

edi_pdf2data/models/pdf2data_template.py
```diff
@@ -0,0 +1,130 @@
...
+    pdf2data_template_yml = fields.Text()
+    pdf_file = fields.Binary(attachment=True)
+    pdf_filename = fields.Char()
+    file_result = fields.Char(readonly=True)
+    file_processed_result = fields.Html(readonly=True)
+    sequence = fields.Integer(default=20)
+    extracted_text = fields.Text(readonly=True)
+
+    def _parse_pdf(self, data):
+        data_file = NamedTemporaryFile(
```
LoisRForgeFlow, line 63: Wouldn't it make sense to move this to the specific component of each template?

etobella: Not really, as the parsing depends directly on the template you are using, so, IMO, it seems legit to me.

edi_pdf2data/components/edi_input_pdf2data_process.py
```diff
@@ -0,0 +1,29 @@
...
+        extracted_text, data, template = (
+            self.env["pdf2data.template"]
+            .search([])
+            ._parse_pdf(self.exchange_record.exchange_file)
+        )
+        if not template:
+            return
+        component = self.component(
+            usage="process_data",
+            backend_type=self.exchange_record.backend_id.backend_type_id.code,
```
simahawk: these keys should be already in the work ctx

edi_pdf2data/components/edi_input_pdf2data_process.py
```diff
@@ -0,0 +1,29 @@
...
+    _name = "edi.input.process.pdf2data.import_data"
+    _inherit = "edi.component.input.mixin"
+    _usage = "input.process"
+    _backend_type = "import_data"
+    _exchange_type = "pdf2data"
+
+    def process(self):
+        extracted_text, data, template = (
+            self.env["pdf2data.template"]
+            .search([])
```
simahawk: can't we search for a specific thing?

edi_pdf2data/components/edi_process_data.py
```diff
@@ -0,0 +1,25 @@
...
+from odoo.addons.component.core import AbstractComponent
+
+
+class EDIBackendInputComponentMixin(AbstractComponent):
+    """Generate input content.
+    """
+
+    _name = "edi.component.process_data.mixin"
+    _inherit = "edi.component.mixin"
+    _usage = "process_data"
```
simahawk, line 13: `process_data` seems too wide and not explicit. I'd call it `pdf2data` simply.

edi_pdf2data/models/pdf2data_template.py
```diff
@@ -0,0 +1,130 @@
...
+            if type(tpl["keywords"]) is not list:
+                tpl["keywords"] = [tpl["keywords"]]
+            # Define excluded_keywords as empty list if not provided
+            # Convert to list if only one provided
+            if "exclude_keywords" not in tpl.keys():
+                tpl["exclude_keywords"] = []
+            elif type(tpl["exclude_keywords"]) is not list:
+                tpl["exclude_keywords"] = [tpl["exclude_keywords"]]
+
+            t = InvoiceTemplate(tpl)
```
simahawk: so this is only for invoices? Seems not to be generic to stay here.

edi_pdf2data/models/pdf2data_template.py
```diff
@@ -0,0 +1,130 @@
...
+                {
+                    "file_result": False,
+                    "file_processed_result": "Data cannot be processed",
+                    "extracted_text": "No text extracted",
+                }
+            )
+            return
+        self.extracted_text = extracted_text
+        self.file_result = data
+        backend = self.env.ref("edi_pdf2data.pdf2data_backend")
```
simahawk: hardcoded reference to a backend?

edi_pdf2data/models/pdf2data_template.py
```diff
@@ -0,0 +1,130 @@
...
+        if component and hasattr(component, "preview_data"):
+            self.file_processed_result = component.preview_data(data, self)
+        else:
+            self.file_processed_result = self._preview_data(data)
+
+    def _preview_data(self, data):
+        result = ""
+        for key in data:
+            value = data[key]
+            if isinstance(value, list):
```
simahawk: better to use a qweb template

edi_pdf2data/tests/com.amazon.aws.yml
```diff
@@ -0,0 +1,26 @@
+# -*- coding: utf-8 -*-
```
simahawk, line 1: pls move these files into `fixtures` folder

edi_pdf2data/wizard/pdf2data_import.py
```diff
@@ -0,0 +1,32 @@
...
+
+class Pdf2dataImport(models.TransientModel):
+    _name = "pdf2data.import"
+    _description = "Wizard to import pdf and extract data"
+
+    pdf_file = fields.Binary(string="PDF", required=True)
+    pdf_file_name = fields.Char()
+
+    def import_pdf(self):
+        backend = self.env.ref("edi_pdf2data.pdf2data_backend")
```
simahawk: again... this should be selectable in this case

edi_pdf2data/__manifest__.py
```diff
@@ -0,0 +1,21 @@
...
+{
+    "name": "Pdf2data Import",
+    "summary": """
+        Module that allows to import data from a pdf""",
+    "version": "13.0.1.0.0",
+    "license": "AGPL-3",
+    "author": "Creu Blanca,Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/edi",
+    "depends": ["edi"],
+    "external_dependencies": {"python": ["invoice2data", "PyMuPDF"]},
```
simahawk: We should not depend on these libs here IMO.
Here we should have the core machinery to be able to define parsers for PDF data.

simahawk (changes requested):
@etobella I'm available for a call if you want to talk about the overall design ;)

edi_pdf2data/__manifest__.py
etobella: You are right, now is the moment to make it new. Let me check it :smile:

Should we reuse the idea from invoice2data or should we make it new?

bosd: IMO invoice2data cen be used. 
Currently we are developing some interesmsting featuresfor next release:
Improved lines parser
Improved tesseract support (auto detect language)
Pdfplumber input support

Currently working on implementing Excalibur/Camelot for advanced table parsing. It is a lot of work, probably not finished for next release.


However Pymupdf is not yet supported as a input parser.


The biggest drawback of the module is to write/store the templates.

etobella: Well, the name of the library is confusing too... It is called invoice2data, but could be used for other kind of documents. Something like pdf2data would have been better IMO.

We have already used it to import docs and generate specific records.

bosd: I think the name of the library has been discussed before, but kept unchanged.
It can always be used to parse pdf's with other content.
Currently I am working on integrating camelot and invoice2data. Very handy tool for importing table data from pdf's.

etobella: Also, the problem is that it is focused on invoices, so you need to disallow some fields if you want to import other kind of data. With the new changes, it is easier for the user to make it, as it is done entirely on web interface. Not using yaml anymore (but you can import invoice2data yml :smile: )

PR #439, opened 2021-09-06
[13.0][FIX] edi_storage: incorrect call to output checker in receive component.

LoisRForgeFlow:
Checker is only relevant for output direction.

@ForgeFlow

### DIFF
--- a/account_invoice_export_server_env/README.rst
+++ b/account_invoice_export_server_env/README.rst
@@ -25,9 +25,6 @@ Server environment for Account Invoice Export
 
 |badge1| |badge2| |badge3| |badge4| |badge5| 
 
-Server environment for Account Invoice Export
-=============================================
-
 This module is based on the `server_environment` module to use files for
 configuration. So we can have a different configuration for each
 environment (dev, test, integration, prod).  This module define the config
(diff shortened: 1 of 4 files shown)


etobella (approved):
I understand your approach. However, just to keep to be safe, a comment from @simahawk would be great to know if there was a reason for it

LoisRForgeFlow:
> I understand your approach. However, just to keep to be safe, a comment from @simahawk would be great to know if there was a reason for it

I copied it from the output component, so it was my fault when creating this input component, Simone wouldn't have done such a mistake in the first place :stuck_out_tongue:

JordiMForgeFlow (approved):

etobella:
https://github.com/OCA/edi/pull/402/files#diff-afd8f746cd5b16528b2bcead53330d7aa99b668b30305f51826902834179b3ddR17

I see, in that case it seem right to me to proceed with merge. Can you review that this error has not been added on 14?

/ocabot merge patch

simahawk:
Hola! Seems I'm arriving late to the party :smile: 

> Simone wouldn't have done such a mistake in the first place stuck_out_tongue

Naaa it passed under my eyes in the 1st place... :wink: 

Indeed, doesn't make sense here, **today**.

TBH `check` is not really clear at all because it could be used for both input and output.
In the future we should maybe rename it to something more clear or pass the direction or something else.

LoisRForgeFlow:
@etobella @simahawk travis issue fixed! could you merge?

PR #442, opened 2021-09-09
[14.0][IMP] base_ubl: add hooks to allow customization

FerranCforgeFlow:
Return the _item_ node in the __ubl_add_item_ method of the base_ubl

### DIFF
--- a/base_ubl/models/ubl.py
+++ b/base_ubl/models/ubl.py
@@ -385,6 +385,7 @@ class BaseUbl(models.AbstractModel):
         description.text = name
         name_node = etree.SubElement(item, ns["cbc"] + "Name")
         name_node.text = product_name or name.split("\n")[0]
+
         if seller_code:
             seller_identification = etree.SubElement(
                 item, ns["cac"] + "SellersItemIdentification"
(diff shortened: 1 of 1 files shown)


LoisRForgeFlow (approved):
Thanks!

bealdav (approved):

etobella:
Wouldn't be easier to apply the same changes from here:
https://github.com/OCA/edi/pull/429/files

This way, this hooks are no longer necessary

LoisRForgeFlow:
@etobella Probably you are right and that will be the best solution going forward. However, since these hooks are already in v13, I find some benefit in just including them in v14 as they are now and keeping versions aligned. Less migration headaches! :smile:

etobella:
I know it is easier, but it is the best place to do this on migrations, so, I would prefer to add a return and remove the need of this hooks

LoisRForgeFlow:
@etobella OK, let's do that then. 

@FerranCforgeFlow Could you attend Enric's suggestion?

FerranCforgeFlow:
Done! All the changes related with the supplier party have been removed since in #429 the node is returned and can be edited, and the item node of the __ubl_add_item_ method is now also returned.

etobella (approved):

Issue #445, opened 2021-09-21
invoice2data most recent version crash odoo


flotho:
Here are the logs with `invoice2data 0.3.6`

```
pipenv/lib/python2.7/site-packages/invoice2data/extract/invoice_template.py", line 69
    def prepare_input(self, extracted_str: str) -> str:
                                         ^
SyntaxError: invalid syntax

```
I tested with 0.3.4 and it works like a charm

bosd:
This one can be closed now :)

PR #446, opened 2021-09-21
FIX too recent invoice2data versions causes Odoo crash

flotho:
FIX https://github.com/OCA/edi/issues/445

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-invoice2data>=0.2.74
+invoice2data>=0.2.74,<=0.3.4
 factur-x>=0.3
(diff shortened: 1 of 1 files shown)


flotho:
ping @alexis-via

alexis-via:
@flotho is it because invoice2data dropped py2 support ?

flotho:
> @flotho is it because invoice2data dropped py2 support ?

I didn't check but probably yes.
Regards

etobella (approved):
Is no longer compatible with python2.7, so it seems logic

flotho:
thanks @etobella

PR #448, opened 2021-09-23
[14.0] edi framework misc improvements

simahawk:

### DIFF
--- a/edi_oca/models/edi_backend.py
+++ b/edi_oca/models/edi_backend.py
@@ -16,6 +16,10 @@ from ..exceptions import EDIValidationError
 _logger = logging.getLogger(__name__)
 
 
+def _get_exception_msg(exc):
+    return exc.name if hasattr(exc, "name") else repr(exc)
+
+
 class EDIBackend(models.Model):
     """Generic backend to control EDI exchanges.
 
@@ -196,7 +200,7 @@ class EDIBackend(models.Model):
             try:
                 self._validate_data(exchange_record, output)
             except EDIValidationError as err:
-                error = repr(err)
+                error = _get_exception_msg(err)
                 state = "validate_error"
                 message = exchange_record._exchange_status_message("validate_ko")
                 exchange_record.update(
@@ -261,7 +265,7 @@ class EDIBackend(models.Model):
         except self._swallable_exceptions() as err:
             if self.env.context.get("_edi_send_break_on_error"):
                 raise
-            error = repr(err)
+            error = _get_exception_msg(err)
             state = "output_error_on_send"
             message = exchange_record._exchange_status_message("send_ko")
             res = False
@@ -411,7 +415,7 @@ class EDIBackend(models.Model):
         except self._swallable_exceptions() as err:
             if self.env.context.get("_edi_receive_break_on_error"):
                 raise
-            error = repr(err)
+            error = _get_exception_msg(err)
             state = "input_processed_error"
             res = False
         else:
@@ -458,14 +462,14 @@ class EDIBackend(models.Model):
                 exchange_record._set_file_content(content)
                 self._validate_data(exchange_record)
         except EDIValidationError as err:
-            error = repr(err)
+            error = _get_exception_msg(err)
             state = "validate_error"
             message = exchange_record._exchange_status_message("validate_ko")
             res = False
         except self._swallable_exceptions() as err:
             if self.env.context.get("_edi_receive_break_on_error"):
                 raise
-            error = repr(err)
+            error = _get_exception_msg(err)
             state = "input_receive_error"
             message = exchange_record._exchange_status_message("receive_ko")
             res = False
--- a/edi_oca/models/edi_backend.py
+++ b/edi_oca/models/edi_backend.py
@@ -21,9 +21,10 @@ class EDIBackend(models.Model):
 
     Backends can be organized with types.
 
-    The backend should be responsible for generating export records.
+    The backend should be responsible for managing records.
     For each record it can generate or parse their values
-    depending on their direction (incoming, outgoing).
+    depending on their direction (incoming, outgoing)
+    and send or receive them automatically depending on their state.
     """
 
     _name = "edi.backend"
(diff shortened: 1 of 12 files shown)


simahawk:
@etobella would you have time for a review?

edi_oca/models/edi_backend.py
```diff
@@ -330,7 +337,7 @@ def _check_output_exchange_sync(self, skip_send=False):
             len(new_records),
         )
         for rec in new_records:
-            self.with_delay().exchange_generate(rec)
+            self.with_delay(self._get_job_delay_params(rec)).exchange_generate(rec)
```
etobella: self.with_delay(**self._get_job_delay_params(rec)).exchange_generate(rec)

edi_oca/models/edi_backend.py
```diff
@@ -343,7 +350,7 @@ def _check_output_exchange_sync(self, skip_send=False):
         )
         for rec in pending_records:
             if rec.edi_exchange_state == "output_pending":
-                self.with_delay().exchange_send(rec)
+                self.with_delay(self._get_job_delay_params(rec)).exchange_send(rec)
```
etobella: self.with_delay(**self._get_job_delay_params(rec)).exchange_send(rec)

edi_oca/models/edi_backend.py
```diff
@@ -521,7 +528,7 @@ def _check_input_exchange_sync(self, **kw):
             len(pending_records),
         )
         for rec in pending_records:
-            self.with_delay().exchange_receive(rec)
+            self.with_delay(self._get_job_delay_params(rec)).exchange_receive(rec)
```
etobella: self.with_delay(**self._get_job_delay_params(rec)).exchange_receive(rec)

edi_oca/models/edi_backend.py
```diff
@@ -559,7 +566,9 @@ def _exchange_check_ack_needed(self, pending_records):
             len(ack_pending_records),
         )
         for rec in ack_pending_records:
-            self.with_delay().exchange_create_ack_record(rec)
+            self.with_delay(self._get_job_delay_params(rec)).exchange_create_ack_record(
```
etobella: self.with_delay(**self._get_job_delay_params(rec)).exchange_create_ack_record(rec)

etobella (commented):
I think there is some errors with the queue_job can you check?

simahawk:
> I think there is some errors with the queue_job can you check?

ehm, I was planning to add tests... I finished too late to add them :sweat_smile: 
I'll add tests anyway later.
Good catch mate, thanks :)

simahawk:
meanwhile, could you have a look at this if you have time https://github.com/OCA/edi/pull/447 ?

simahawk:
@etobella fixed and improved tests

simahawk:
@LoisRForgeFlow would you mind having a look here?

etobella (approved):

edi_webservice_oca/views/edi_backend.xml
```diff
@@ -6,11 +6,13 @@
         <field name="model">edi.backend</field>
         <field name="inherit_id" ref="edi_oca.edi_backend_view_form" />
         <field name="arch" type="xml">
-            <group name="config" position="after">
-                <group name="webservice_config">
-                    <field name="webservice_backend_id" />
-                </group>
-            </group>
+            <notebook position="inside">
+                <page name="webservice_config">
```
simahawk:
```suggestion
                <page name="webservice_config" string="Webservice">
```

edi_oca/models/edi_exchange_record.py
```diff
@@ -244,6 +260,18 @@ def action_exchange_process(self):
...
+        if not self.retryable:
+            return False
+        new_state = self._rollback_state_mapping[self.edi_exchange_state]
+        fname = "edi_exchange_state"
+        self[fname] = new_state
+        display_state = self._fields[fname].convert_to_export(self[fname], self)
+        self.message_post(
+            body=_("Action retry: state moved back to '%s'") % display_state
+        )
+        return True
```
LoisRForgeFlow: With this new action, I think we should not process failed files by default here: https://github.com/OCA/edi/blob/14.0/edi_oca/models/edi_backend.py#L543.

Also, it would be good if it accepts recordsets, to be potentially used in batch with a server action.

LoisRForgeFlow (commented):
Now that you are doing some improvements, let me bring up one topic: what do you think about accepting multiple records linked to a single edi exchange record?

simahawk:
> Now that you are doing some improvements, let me bring up one topic: what do you think about accepting multiple records linked to a single edi exchange record?

@LoisRForgeFlow ahh... just read your comment after our conversation at OCA booth :smile: 

We can write an RFC based on what we said (having not file content and lookup on the parent).

edi_oca/models/edi_exchange_record.py
simahawk: > With this new action, I think we should not process failed files by default here: https://github.com/OCA/edi/blob/14.0/edi_oca/models/edi_backend.py#L543.

I don't get this, can you elaborate? From my POV if the file failed either the system or the user can try it again. What is probably desiderable is that that system stops after N attempts to try retry them?

> Also, it would be good if it accepts recordsets, to be potentially used in batch with a server actio

Sure, I'll change it.

edi_oca/models/edi_exchange_record.py
```diff
@@ -237,12 +260,35 @@ def action_exchange_process(self):
...
```
etobella, line 278: That will not work if exchange_file_auto_generate is not checked. As this record will not be processed by the cron. So, at the end, we need to add similar button that I proposed (with less logic, obviously) or add the logic here (I think it is better with the buttons)

etobella: Not blocking, we can proceed with the merge and then check how to improve this :wink:

edi_oca/models/edi_exchange_record.py
simahawk: added loop for recordsets.

For the other issue I think we can discuss it on a 2nd round.

PR #449, opened 2021-09-24
[13.0] edi + edi_exchange_template misc fix/imp

simahawk:
Extrapolated from https://github.com/OCA/edi/pull/411


etobella (approved):
Looking commit by commit seems fine. It might be possible to squash some commits? I mean, some of them are just a fix of the previous one

PR #450, opened 2021-09-24
[14.0] Add endpoint + endpoint_auth_api_key + edi_endpoint_oca

simahawk:
I suggest reviewing commit by commit (eg: first the `endpoint` module).
I could split int 3 PRs but for now I prefer to keep the 3 modules together.
I'll open soon an extra PR for integrating UBL w/ this.

Depends on:

- [x] #448 

@etobella you might be interested in this ;)


etobella (commented):
Mmmh, interesting approach, you are trying to make a specific system to receive documents. Isn't it? Like the inverse of webservice :smile: 

On a first sight, it seem right, but I would prefer to wait for travis to be green :wink:

simahawk:
closing in favour of https://github.com/OCA/edi/pull/469

PR #317, opened 2021-01-09
[MIG] sale_order_import_ubl_http: Migration to 14.0

phuctranfxvn:
- Migration from https://github.com/OCA/edi/pull/241
- Modules required:
  + `auth_api_key` (https://github.com/OCA/server-auth/tree/14.0)
  + `queue_job` (https://github.com/OCA/queue/tree/14.0)
  + `sale_order_import_ubl` migrated in https://github.com/OCA/edi/pull/316

PR #451, opened 2021-09-24
[14.0] Plug sale_order_import w/ EDI framework

simahawk:
Depends on / includes:

- [x] #470 
- [x] https://github.com/OCA/edi/pull/483

`edi_sale_order_import_ubl_endpoint` is meant to replace `sale_order_import_ubl_http` https://github.com/OCA/edi/pull/317
The module itself it's not required IMO but it will be an example on how to setup and endpoint for EDI exchanges.


simahawk:
@sbidoul any idea of what's going wrong here on pre-commit run?
```
     ModuleNotFoundError: No module named 'setuptools'
```
https://github.com/OCA/edi/runs/4650728150?check_suite_focus=true#step:4:163

I've tried updating version as per https://github.com/OCA/oca-addons-repo-template/pull/100 but has no effect (as expected).

FTR I had to do the change manually because `copier update` does not pull latest changes here :man_shrugging:

sbidoul:
@simahawk https://github.com/OCA/oca-addons-repo-template/pull/101

simahawk:
trying w/out the pre-commit fix as suggested on OCA/oca-addons-repo-template#101

sebalix (approved):

edi_ubl_oca/__manifest__.py
```diff
@@ -0,0 +1,22 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    "name": "EDI UBL",
+    "summary": """Define EDI backend type for UBL.""",
+    "version": "14.0.1.0.0",
+    "development_status": "Alpha",
+    "license": "AGPL-3",
+    "website": "https://github.com/OCA/edi",
+    "author": "Camptcamp,Odoo Community Association (OCA)",
```
jbaudoux:
```suggestion
    "author": "Camptocamp,Odoo Community Association (OCA)",
```
:)

jbaudoux: @simahawk There are many typo like this, please review them all

jbaudoux (approved):
LGTM, no tests

edi_ubl_oca/__manifest__.py
simahawk: good catch, I fixed in other modules in this repo https://github.com/OCA/edi/pull/514 I'll do it here too. Tnx!

PR #458, opened 2021-10-06
[IMP] edi: Add buttons for not automatically managed and EDI errors

etobella:
Some buttons are added to manage errors on Odoo or, even, not processed EDI exchange records (last commit)
+ backward port of https://github.com/OCA/edi/pull/448

@simahawk

### DIFF
--- a/edi/views/edi_exchange_record_views.xml
+++ b/edi/views/edi_exchange_record_views.xml
@@ -27,12 +27,18 @@
                         name="action_open_related_record"
                         type="object"
                         string="Related record"
+                        attrs="{'invisible': [('res_id', '=', False)]}"
                     />
+                    <!-- FIXME: this `invisible` domain does not work -->
                     <button
                         name="action_open_related_exchanges"
                         type="object"
                         string="Related exchanges"
+                        attrs="{'invisible': [('related_exchange_ids', '=', False)]}"
                     />
+                    <!-- FIXME: colors are ignored...
+                    and can't find how they are supposed to work.
+                    Probably not supported at all. -->
                     <field
                         attrs="{'invisible': [('direction', '=', 'input')]}"
                         name="edi_exchange_state"
@@ -103,6 +109,7 @@
                             name="related_odoo_record"
                             string="Related record"
                             groups="base.group_no_one"
+                            attrs="{'invisible': [('res_id', '=', 0)]}"
                         >
                           <field name="res_id" />
                           <field name="model" />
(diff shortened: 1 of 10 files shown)


simahawk:
@etobella I'm not sure we want to always expose all those buttons. Also, this looks pretty similar to what I've done here w/ just one button that says "Retry" and is going to retry the right action but async... Do you really need sync action?

etobella:
No need of sync, I was just trying to manage that. I will check your commits and push it here :wink:

simahawk:
I missed the link for "here" https://github.com/OCA/edi/pull/448 :rofl:

etobella:
I applied your changes, but added the buttons on a last commit. Just one comment, on 13, you cannot have two statusbar over the same field. So, I removed it (maybe it is the problem you find with the colors?)

This also happened on 12 and it seems to happen here too, I don't know if it happens on 14

etobella:
@simahawk Can you review this once as your original PR as been merged :smiley: ?

edi/views/edi_exchange_record_views.xml
```diff
@@ -23,28 +23,65 @@
...
+                <field name="exchange_file_auto_generate" invisible="1" />
+                <header />
                 <sheet>
+                    <div name="button_box" class="oe_button_box">
+<button
+                            name="action_exchange_send"
+                            type="object"
+                            class="oe_stat_button"
+                            icon="fa-send"
+                            attrs="{'invisible': ['|', ('edi_exchange_state', '!=', 'output_pending'), ('exchange_file_auto_generate', '=', True)]}"
```
simahawk: the send should be independent from `exchange_file_auto_generate`. This flag is only for generating the output.

edi/views/edi_exchange_record_views.xml
```diff
@@ -23,28 +23,65 @@
...
+                        />
+                        <button
+                            name="action_exchange_generate"
+                            type="object"
+                            class="oe_stat_button"
+                            icon="fa-cubes"
+                            attrs="{'invisible': ['|', ('edi_exchange_state', '!=', 'new'), ('exchange_file_auto_generate', '=', True)]}"
+                            string="Generate"
+                        />
+                        <button
```
simahawk, line 38: same all around. The only btn that should use `exchange_file_auto_generate` is the generate one.
I mean, when something fails you should rely on the automatic machinery to do the job. And for this - apart from the file generation - the "Retry" btn is enough AFAIS.

simahawk (changes requested):

simahawk:
can you rebase?

edi/views/edi_exchange_record_views.xml
etobella: I see, you are right. Fixing it

simahawk:
@etobella better if you reset hard and cherry-pick your commits becase I cleaned up the history on my PR

etobella:
@simahawk All comments have been attended, shall we proceed with merge?

etobella:
I passed one commit to #465 as a conflict was raised between branches @simahawk It was independent to all this, so it makes sense :smile:

simahawk (approved):

simahawk:
@etobella would you mind fwd porting the last commit when merged?

PR #461, opened 2021-10-11
8.0 FIX pin invoice2data to 0.3.4

bealdav:
MAKE 8.0 GREEN AGAIN

With Invoice2data 0.3.6 we have syntax incompatible with a python 2.7 like here

https://github.com/invoice-x/invoice2data/blob/1adb49cb74b17e1dd278886ee4c432ed2dcaf443/src/invoice2data/extract/invoice_template.py#L69


bealdav:
see https://github.com/OCA/edi/pull/446

etobella (approved):

PR #465, opened 2021-10-15
[IMP] edi: manage the case when no check is defined

etobella:
When no check is defined, we could assume that it was processed properly. WDYT?

@simahawk

### DIFF
--- a/edi/models/edi_backend.py
+++ b/edi/models/edi_backend.py
@@ -37,6 +37,12 @@ class EDIBackend(models.Model):
         required=True,
         ondelete="restrict",
     )
+    output_sent_processed_auto = fields.Boolean(
+        help="""
+    Automatically set the record as processed after sending.
+    Usecase: the web service you send the file to processes it on the fly.
+    """
+    )
 
     def _get_component(self, exchange_record, key):
         candidates = self._get_component_usage_candidates(exchange_record, key)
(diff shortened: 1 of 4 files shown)


simahawk:
while it might be good not to raise an error, IDT is a good idea to change the state.
You should handle the transition in your own implementation probably (eg: automated action).
Can you elaborate a bit on the use case?

etobella:
Well, with your comment, I have to implement every time a check that everything worked fine, but it is not already correct. For example, if you use a Webservice, you know at the end that everything went right, but you cannot change the information on the component, as the state will be written later. Without the state change we cannot expect to use queue in this cases. Also, why using a new automated action if we already have this one?

simahawk:
Sorry, I still don't get the point. Seems you are mixing up "webservice" case and "storage" case. We should probably have a chat around this :)

etobella:
@simahawk Changes applied as talked

edi/models/edi_backend.py
```diff
@@ -37,6 +37,7 @@ class EDIBackend(models.Model):
         required=True,
         ondelete="restrict",
     )
+    output_sent_processed_auto = fields.Boolean()
```
simahawk:
```suggestion
    output_sent_processed_auto = fields.Boolean(help=""""
    Automatically set the record as processed after sending.
    Usecase: the web service you send the file to processes it on the fly.
    """)
```

simahawk (approved):

olgamarcocb (approved):
:+1:

albariera (approved):

PR #467, opened 2021-10-22
[IMP] edi: manage the case when no check is defined

olgamarcocb:
cherry-pick from #465


etobella (approved):
Thanks for the cherry-pick :+1:

simahawk (approved):

PR #469, opened 2021-10-26
[14.0] Add endpoint_route_handler + endpoint modules

simahawk:
* `endpoint_route_handler` allows to create routes on the fly both as a tool and as a mixin
* `endpoint` leverages the route handler to provide an handy UI and exec machinery to manage custom routes from the UI

Replaces #450 (compared to that, it moves out edi_endpoint and endpoint_auth_api_key and refactor base route handing out).


gurneyalex:
Just starting to look at the module, first question that comes up is: why is this in edi and not in rest-framework?

simahawk:
> Just starting to look at the module, first question that comes up is: why is this in edi and not in rest-framework?

Because it has nothing to deal w/ REST framework. 
In this repo we have already `webservice` module which is already integrated here and we'll have `edi_endpoint_oca` from https://github.com/OCA/edi/pull/470, thus I decided to keep them together for now.

IMO we need a new repo like `OCA/webservices` which can host any thing related to this no matter if related to REST stuff or not. Then maybe we can make rest-framework converge there (seems like @lmignon is not satisfied with that name anymore).

lmignon:
@simahawk IMO we'll keep the rest-framework repo. This repo already includes addons related to other kind of services (graphql) . Thinking about the future of base_rest,  I do have doubts about the relevance of the name 'base_rest' in relation to the features covered. If these last days I had the feeling that this name was not the most appropriate, my last readings and researches finally confirm me in this initial choice.

simahawk:
> @simahawk IMO we'll keep the rest-framework repo. This repo already includes addons related to other kind of services (graphql) . Thinking about the future of base_rest, I do have doubts about the relevance of the name 'base_rest' in relation to the features covered. If these last days I had the feeling that this name was not the most appropriate, my last readings and researches finally confirm me in this initial choice.

Good to know, thanks for the update :)
Any opinion regarding having a new repo for "web services" modules?

lmignon:
> Good to know, thanks for the update :) Any opinion regarding having a new repo for "web services" modules?

"web services" ? What's the difference between "web services" and a "REST api"?

simahawk:
"web services" includes everything that expose something internally or deals w/ an external service no matter if it's REST or not. An example is the `webservice` generic module that stays in this repo already: you call an external service, you don't create one. Also, the concept of REST - strictly speaking - may not apply to every "endpoint" you expose.

lmignon:
strictly speaking, `base_rest` doesn't enforce the REST principles... At this stage `base_rest` allows you to define a set of endpoint (aka `@restapi.method` ) for a given resource name( `_usage='sales'`).  All these resources are grouped to create an application (`_collection_name='my_eshop'`). The entry point for such an application is setup by implementing a specific RestController where we  declare the root path of the application and others contextual info such as the default authentication to use. At this stage, there is nothing specific to REST. :smirk: 

The main difference here is that your addon can be used to create simple endpoints directly from the UI (or xml records). It become therefore easy to expose a resource without the "application" and "resource" levels introduced by `base_rest`. I don't really have any arguments to say if it would be better to have a dedicated repo for this module or to put it in res-framework next to graphql_base.

simahawk:
@lmignon that's right. And because of this, I feel even more that "rest" inside that repo name is reductive. But you know what? That's not important ATM. It's just my POV. Interesting discussion, we can maybe move it fwd at OCA days ;)

That being said, anyone willing to drop a review? :grin: :pray:

endpoint/controllers/main.py
```diff
@@ -0,0 +1,59 @@
...
+        return self._handle_endpoint(env, endpoint_route, **params)
+
+
+class EndpointController(http.Controller, EndpointControllerMixin):
+    pass
+
+    # @http.route(["/demo/user"], auth="user")
+    # def test_endpoint(self, **params):
+    #     request.env
+    #     return "ok"
```
lmignon: @simahawk Do we need these commented lines?

endpoint/models/endpoint_mixin.py
```diff
@@ -0,0 +1,178 @@
...
+        required=True,
+    )
+    code_snippet = fields.Text()
+    code_snippet_docs = fields.Text(
+        compute="_compute_code_snippet_docs",
+        default=lambda self: self._default_code_snippet_docs(),
+    )
+    exec_as_user_id = fields.Many2one(comodel_name="res.users")
+
+    def _selection_exec_mode(self):
```
lmignon, line 33: @simahawk What's the benefit to use a callable to get the available choices of you selection field in place of using `selection_add` if an addon need to extend this list?

lmignon (approved):
LGTM (Code review only)

endpoint/controllers/main.py
simahawk: no, garbage from early tests indeed, I'll drop it.

endpoint/models/endpoint_mixin.py
simahawk: Just a matter of taste I guess. I prefer to have a real method to override instead of using the fields.Selection's magic.
Any concrete pro in favor of it?

lmignon: Just a matter of taste :smirk: I prefer `selection_add` It seems to me more contextual since the command is declared in the field definition and not into a linked method. But the result is the same.

endpoint_route_handler/__manifest__.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2021 Camptcamp SA
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+{
+    "name": "Endpoint",
```
TDu:
```suggestion
    "name": "Endpoint Route Handler",
```

endpoint_route_handler/__manifest__.py
```diff
@@ -0,0 +1,16 @@
...
+    "summary": """Provide custom endpoint machinery.""",
```
TDu: You probably want to change this as well

endpoint_route_handler/models/endpoint_route_handler.py
```diff
@@ -0,0 +1,278 @@
...
+        compute="_compute_endpoint_hash", help="Identify the route with its main params"
+    )
+
+    csrf = fields.Boolean(default=False)
+
+    _sql_constraints = [
+        (
+            "endpoint_route_unique",
+            "unique(route)",
+            "You can register an endpoint route only once.",
```
TDu, line 59: So you could not register multiple `request_method` for the same route ?

endpoint_route_handler/models/endpoint_route_handler.py
```diff
@@ -0,0 +1,278 @@
...
+    def _compute_route(self):
+        for rec in self:
+            rec.route = rec._clean_route()
+
+    def _inverse_route(self):
+        for rec in self:
+            rec.route = rec._clean_route()
+
+    _endpoint_route_prefix = ""
+    """Prefix for all routes, includes slashes.
```
TDu: Should this not be a comment ?

endpoint_route_handler/models/endpoint_route_handler.py
simahawk: good point. Well, we should handle this on the same endpoint by allowing to define more than 1 method as routing allows to. I'll add it to the roadmap!

endpoint_route_handler/models/endpoint_route_handler.py
simahawk: nope, this is how you can document attributes especially if you generate docs using sphinx.
Now, the comment that I would add is another: trash this attribute and use something else (eg: a computed field).
Not sure about this TBH.

simahawk: done

endpoint_route_handler/models/endpoint_route_handler.py
simahawk: done

simahawk:
@TDu tnx for your review. All comments attended.

TDu (approved):

PR #472, opened 2021-10-29
[14.0] edi: fix wrong status bar display new exc record

simahawk:
Without this check you get 2 overlapping bars till you select an exc type.


simahawk:
@etobella would you mind?

etobella (approved):

PR #473, opened 2021-11-02
[IMP] edi: External identifier must be unique on a type and backend

etobella:
It might be possible to have the same external_identifier for different backends and types.

@simahawk


simahawk (approved):
I'm not sure this always the case but probably the most common.
@etobella pls keep commit msg short and move the details into their description :wink: 

Eg:

```
edi: ext identifier unique by type and backend

It might be possible to have the same external_identifier for different backends and types.
```

LoisRForgeFlow (approved):
Nothing against

PR #480, opened 2021-11-08
[IMP] edi: fix external identifier

olgamarcocb:
cherry-pick from #473


simahawk (changes requested):
you should cherry-pick @etobella commit (or use format-patch)

etobella (approved):

simahawk (approved):

PR #487, opened 2021-11-29
[14.0] edi_account: forward port commits from 13.0

etobella:


MiquelRForgeFlow (approved):

pedrobaeza (approved):

edi_account_oca/views/account_move.xml
```diff
@@ -7,6 +7,11 @@
         <field name="model">account.move</field>
         <field name="inherit_id" ref="account.view_move_form" />
         <field name="arch" type="xml">
+            <group name="payments_info_group" position="after">
```
MiquelRForgeFlow:
```suggestion
            <group name="sale_info_group" position="after">
```
that group doesn't exist anymore. see https://github.com/odoo/odoo/commit/def8817d727a392d76f18b654ffa01be99135ace

edi_account_oca/views/account_move.xml
```diff
@@ -7,6 +7,11 @@
         <field name="model">account.move</field>
         <field name="inherit_id" ref="account.view_move_form" />
         <field name="arch" type="xml">
+            <group name="sale_info_group" position="after">
```
MiquelRForgeFlow: Sorry, in fact it should be better the following
```suggestion
            <group name="accounting_info_group" position="after">
```
but this is not blocking haha

FernandoRomera (approved):

PR #494, opened 2021-12-14
[14.0] Fix account_invoice_export wrong assignment

TDu:
This `invoice_send` field does not exist (anymore ?)

### DIFF
--- a/account_invoice_export/models/account_move.py
+++ b/account_invoice_export/models/account_move.py
@@ -53,7 +53,6 @@ class AccountMove(models.Model):
                     # The chatter of the invoice need to be updated, when the job fails
                     self.with_env(new_env).log_error_sending_invoice(values)
             raise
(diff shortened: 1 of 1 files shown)


account_invoice_export/models/account_move.py
```diff
@@ -53,7 +53,6 @@ def _job_export_invoice(self, resend_invoice=False):
                     # The chatter of the invoice need to be updated, when the job fails
                     self.with_env(new_env).log_error_sending_invoice(values)
             raise
-        self.invoice_send = True
```
simahawk, line 56: time to test this? :)

TDu: I do think to add a test for it would be good, yes

simahawk (approved):
on prod since a while, tests can be added later

PR #497, opened 2021-12-17
[10.0][IMP] base_business_document_import: search for Bank Account without partner set on it

yankinmax:
Bank Accounts can exist without partners set on them, so at the moment if the search is performed we don't get the expected result => existing Bank Account without Partner
Constraint is applied and we get a noisy warning message though there is actually existing Bank Account.
```
Odoo Warning - Validation Error
Account Number must be unique
```
To fix this we need to include in search also results without partners. Similar condition is applied for `company_id` Checked on real database.

### DIFF
--- a/base_business_document_import/models/business_document_import.py
+++ b/base_business_document_import/models/business_document_import.py
@@ -281,6 +281,7 @@ class BusinessDocumentImport(models.AbstractModel):
             '|', ('company_id', '=', False),
             ('company_id', '=', company_id),
             ('sanitized_acc_number', '=', iban),
+            '|', ('partner_id', '=', False),
             ('partner_id', '=', partner.id),
             ])
         if bankaccounts:
--- a/base_business_document_import/models/business_document_import.py
+++ b/base_business_document_import/models/business_document_import.py
@@ -285,6 +285,8 @@ class BusinessDocumentImport(models.AbstractModel):
             ('partner_id', '=', partner.id),
             ])
         if bankaccounts:
+            if not bankaccounts[0].partner_id:
+                bankaccounts[0].partner_id = partner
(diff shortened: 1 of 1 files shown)


ajaniszewska-dev (approved):

i-vyshnevska (changes requested):
partner should be set

base_business_document_import/models/business_document_import.py
```diff
@@ -281,9 +281,12 @@ def _match_partner_bank(
             '|', ('company_id', '=', False),
             ('company_id', '=', company_id),
             ('sanitized_acc_number', '=', iban),
+            '|', ('partner_id', '=', False),
             ('partner_id', '=', partner.id),
             ])
         if bankaccounts:
+            if not bankaccounts[0].partner_id:
+                bankaccounts[0].partner_id = partner
```
yvaucher, line 289: May I suggest to raise an error in this case. This is clearly something fishy with the data and changing it automatically forces us to make assumptions that might be wrong and the user won't be aware of the change.

As the acc_number is unique, we can most likely assume that bankaccounts variable contains 0 or 1 result and we could remove the partner_id part for the search. Because we are in trouble either the partner_id is null or if it is set on an other partner. (This last case could happen in case of duplicates)

We cannot indeed try to create a new bank account in case of bad partner_id, so raising proper exceptions could be a life saver here and let the user figure out how to fixup his data.

yvaucher: @alexis-via any opinion on this? What do you prefer?

Better error messages -> instead of a unique number error raise an error to explain the partner setting is wrong on the bank account with IBAN ...
Or
Automatically fixup the partner_id

?

PR #506, opened 2022-01-10
[14.0] Add module edi_bank_statement_import

TeoGoddet:
This module plugs edi framework into account_statement_import from OCA/bank-statement-import
It allow to parse data retrieved trough the framework as bank statement.

It's heavily based on https://github.com/Noviat/account_ebics/blob/14.0/account_ebics/models/ebics_file.py and #451

### DIFF
--- a/edi_bank_statement_import/components/process.py
+++ b/edi_bank_statement_import/components/process.py
@@ -1,0 +1,92 @@
+# Copyright 2021 AGEPoly - Téo Goddet
+# @author: Téo Goddet <teo.goddet@gmail.>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+import base64
+
+from odoo import _, api
+
+from odoo.addons.component.core import Component
+
+
+# TODO: add tests
+class EDIExchangeBankStatementInput(Component):
+    """Process bank statements."""
+
+    _name = "edi.input.bank.statement.process"
+    _inherit = "edi.component.input.mixin"
+    _usage = "input.process.bank.statement"
+
+    def __init__(self, work_context):
+        super().__init__(work_context)
+        self.settings = self.type_settings.get("bank_statement", {})
+
+    def process(self):
+        wiz = self._setup_wizard()
+        result = {
+            "statement_ids": [],
+            "notifications": [],
+        }
+
+        file_data = base64.b64decode(wiz.statement_file)
+        wiz.import_single_file(file_data, result)
+
+        return self._handle_result(result["statement_ids"], wiz.statement_filename)
+
+    def _handle_result(self, statement_ids, filename):
+        if not statement_ids:
+            self.exchange_record.sudo().message_post(
+                body=_(
+                    "You have already imported this file (%s) or this file "
+                    "only contains already imported transactions."
+                )
+                % filename
+            )
+            return True
+
+        if len(statement_ids) == 1:
+            self._handle_single_result(self.exchange_record, statement_ids[0])
+            return True
+        else:
+            for statement_id in statement_ids:
+
+                child_exchange = self.exchange_record.create(
+                    {
+                        "identifier": "%s_%s"
+                        % (self.exchange_record.identifier, statement_id),
+                        "type_id": self.exchange_record.type_id.id,
+                        "backend_id": self.exchange_record.backend_id.id,
+                        "edi_exchange_state": "input_processed",
+                        "parent_id": self.exchange_record.id,
+                    }
+                )
+                self._handle_single_result(child_exchange, statement_id)
+                self.exchange_record.sudo().message_post(
+                    body=_("Multiple Bank Statements created (See child exchanges)")
+                )
+            return True
+
+    @api.model
+    def _handle_single_result(self, exchange_record, statement_id):
+        statement = self.env["account.bank.statement"].browse(statement_id)
+        if self._statement_should_be_confirmed():
+            statement.button_post()
+        exchange_record.sudo().message_post(
+            body=_("Bank Statement %s created") % statement.name
+        )
+        exchange_record.sudo()._set_related_record(statement)
+
+    def _setup_wizard(self):
+        """Init a `account.statement.import` instance for current record."""
+        wiz_ctx = self.settings.get("wiz_ctx", {})
+
+        wiz_model = "account.statement.import"
+        wiz_vals = {
+            "statement_filename": self.exchange_record.exchange_filename,
+            "statement_file": self.exchange_record._get_file_content(binary=False),
+        }
+        wiz = self.env[wiz_model].with_context(wiz_ctx).sudo().create(wiz_vals)
+
+        return wiz
+
+    def _statement_should_be_confirmed(self):
+        return self.settings.get("auto_post", False)
--- a/edi_bank_statement_import/data/data.xml
+++ b/edi_bank_statement_import/data/data.xml
@@ -1,0 +1,24 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="edi_exchange_backend" model="edi.backend.type">
+        <field name="name">Bank Statement Backend Type</field>
+        <field name="code">bank_statement_backend_type</field>
+    </record>
+    <record id="edi_exchange_type" model="edi.exchange.type">
+        <field name="name">Bank Statement Import</field>
+        <field name="code">bank_statement_import</field>
+        <field name="backend_type_id" ref="edi_exchange_backend" />
+        <field name="direction">input</field>
+        <field name="exchange_file_ext">xml</field>
+        <field name="advanced_settings_edit">
+components:
+    process:
+        usage: input.process.bank.statement
+bank_statement:
+    auto_post: false
+    wiz_ctx:
+        foo: "bar"
+        default_sheet_mapping_id: 1
+        </field>
+    </record>
+</odoo>
--- a/edi_bank_statement_import/readme/DESCRIPTION.rst
+++ b/edi_bank_statement_import/readme/DESCRIPTION.rst
@@ -1,0 +1,7 @@
+Plug bank_statement_import into EDI machinery.
+
+
+This module allows to use account_bank_statement_import to parse edi exchange (input).
+This allow you ro retrieve bank stazement from a any edi exchange backend (webservice, sftp, local filesystem, ...)
+
+TODO: shall we add an exchange type example as demo?
--- a/edi_bank_statement_import/README.rst
+++ b/edi_bank_statement_import/README.rst
@@ -1,0 +1,1 @@
+wait for the bot ;)
--- a/edi_bank_statement_import/__init__.py
+++ b/edi_bank_statement_import/__init__.py
@@ -1,0 +1,1 @@
+from . import components
--- a/edi_bank_statement_import/__manifest__.py
+++ b/edi_bank_statement_import/__manifest__.py
@@ -1,0 +1,17 @@
+# Copyright 2021 AGEPoly - Téo Goddet
+# @author: Téo Goddet <teo.goddet@gmail.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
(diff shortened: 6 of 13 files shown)


edi_bank_statement_import/components/process.py
```diff
@@ -0,0 +1,74 @@
...
+            self.exchange_record.sudo()._set_related_record(statement)
+            return _("Bank Statement %s created") % statement.name
+        else:
+            for statement_id in result['statement_ids']:
+                statement = self.env["account.bank.statement"].browse(statement_id)
+                if self._statement_should_be_confirmed():
+                    statement.button_post()
+                names.append(statement.name)
+    
+                # TODO : I don't know what to do when there is multiple statements,
```
TeoGoddet: Hi !
Here it's possible to have mutiple statements in a single xml file, that mean multiple records for one exchange.
I suppose that it's possible to create child exchange but i could not find infos/example on how it should be done.

simahawk: @TeoGoddet we've discussed this recently. IMO a good approach could be:

1. make exchange_record.exchange_file computed and write-able
2. if the record is a child and there's no value look at the parent

If you take action on this, please do it in a separated PR.

edi_bank_statement_import/components/process.py
```diff
@@ -0,0 +1,81 @@
...
+        if not result['statement_ids']:
+            self.exchange_record.sudo().message_post(
+                body=_(
+                    "You have already imported this file (%s) or this file "
+                    "only contains already imported transactions."
+                ) % wiz.statement_filename
+            )
+            return True
+
+        if len(result['statement_ids']) == 1:
```
simahawk: pls split this `process` method into smaller methods

edi_bank_statement_import/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,7 @@
+Plug bank_statement_import into EDI machinery.
+
+
+This module allows to use account_bank_statement_import to parse edi exchange (input).
+This allow you ro retrieve bank stazement from a any edi exchange backend (webservice, sftp, local filesystem, ...)
+
+TODO: shall we add an exchange type example as demo?
```
simahawk, line 7: yes pls :)

edi_bank_statement_import/tests/edi_bank_statement_import_test.py
```diff
@@ -0,0 +1,161 @@
...
+            "remote_account": "NL69ABNA0522123643",
+            "transferred_amount": 1405.31,
+            "value_date": date(year=2014, month=1, day=5),
+            "ref": "115",
+        },
+    ]
+
+    def setUp(self):
+        super(TestImport, self).setUp()
+        bank = self.env["res.partner.bank"].create(
```
simahawk: this code runs for every test. Please use `SavepointCase` + `setUpClass` and if you don't rely on tracking also use this line:
```
```
You'll see this approach in edi_oca too.

edi_bank_statement_import/tests/edi_bank_statement_import_test.py
```diff
@@ -0,0 +1,161 @@
...
+                "name": "Bank Journal - (test camt)",
+                "code": "TBNKCAMT",
+                "type": "bank",
+                "bank_account_id": bank.id,
+                "currency_id": self.env.ref("base.EUR").id,
+            }
+        )
+
+    def test_statement_import(self):
+        """Test correct creation of single statement."""
```
simahawk: here you are testing the base wizard, this should be done in the base module not here unless you are testing a special behavior provided by this new module.

edi_bank_statement_import/tests/edi_bank_statement_import_test.py
```diff
@@ -0,0 +1,161 @@
...
+                    )
+                    # TODO and bank_account_id was removed from line
+                    # and line.bank_account_id.acc_number
+                    # == self.transactions[0]["account_number"]
+                    for line in statement_lines
+                )
+            )
+
+    def test_zip_import(self):
+        """Test import of multiple statements from zip file."""
```
simahawk: same.

edi_bank_statement_import/tests/edi_bank_statement_import_test.py
```diff
@@ -0,0 +1,161 @@
...
+            camt_file = base64.b64encode(datafile.read())
+            self.env["account.statement.import"].create(
+                {"statement_filename": "test import", "statement_file": camt_file}
+            ).import_file_button()
+            bank_st_record = self.env["account.bank.statement"].search(
+                [("name", "in", ["1234Test/2", "1234Test/3"])]
+            )
+
+        self.assertTrue(all([st.line_ids for st in bank_st_record]))
+        self.assertEqual(bank_st_record[0].line_ids.mapped("sequence"), [1, 2, 3])
```
simahawk: pls have a look at how tests are done in edi_oca and test the component.

simahawk (changes requested):

edi_bank_statement_import/components/process.py
TeoGoddet: done

edi_bank_statement_import/components/process.py
TeoGoddet: #542

edi_bank_statement_import/tests/edi_bank_statement_import_test.py
TeoGoddet: removed the tests

edi_bank_statement_import/tests/edi_bank_statement_import_test.py
TeoGoddet: removed the tests

edi_bank_statement_import/tests/edi_bank_statement_import_test.py
TeoGoddet: removed the tests

edi_bank_statement_import/tests/edi_bank_statement_import_test.py
TeoGoddet: removed the tests

edi_bank_statement_import/readme/DESCRIPTION.rst
TeoGoddet: I'm not really able to, but let the message (inspired from #451)

TeoGoddet:
Hello @OCA/edi-maintainers, 
I'm sorry to bother you, but I don't understand why the test fails on another module. Does someone have an idea ?

edi_bank_statement_import/readme/DESCRIPTION.rst
TeoGoddet: Done, please check it

TeoGoddet:
Hello @simahawk, 
I've added a demo and some tests (based on your work for sale order [75b81519f61aa385c3b412be4049aeeaf60736b1](https://github.com/OCA/edi/commit/75b81519f61aa385c3b412be4049aeeaf60736b1))

etobella (approved):
Makes sense to me.

etobella:
@simahawk As original reviewer, can you give us your opinions ? :pray:

edi_bank_statement_import/components/process.py
```diff
@@ -0,0 +1,92 @@
...
+                        "backend_id": self.exchange_record.backend_id.id,
+                        "edi_exchange_state": "input_processed",
+                        "parent_id": self.exchange_record.id,
+                    }
+                )
+                self._handle_single_result(child_exchange, statement_id)
+                self.exchange_record.sudo().message_post(
+                    body=_("Multiple Bank Statements created (See child exchanges)")
+                )
+            return True
```
simahawk, line 66: Not a blocker, but nice to have:

```suggestion
        file_data = base64.b64decode(wiz.statement_file)
        wiz.import_single_file(file_data, result)
        
        if not statement_ids:
            return self._handle_already_imported()
        if len(statement_ids) == 1:
            return self._handle_single_result(self.exchange_record, statement_ids[0])
        return self._handle_result(result["statement_ids"], wiz.statement_filename)
   
    def _handle_already_imported(self, wiz):
            self.exchange_record.sudo().message_post(
                body=_(
                    "You have already imported this file (%s) or this file "
                    "only contains already imported transactions."
                )
                % wiz.statement_filename
            )
            return False
            
    def _handle_result(self, statement_ids, filename):
            for statement_id in statement_ids:

                child_exchange = self.exchange_record.create(
                    {
                        "identifier": "%s_%s"
                        % (self.exchange_record.identifier, statement_id),
                        "type_id": self.exchange_record.type_id.id,
                        "backend_id": self.exchange_record.backend_id.id,
                        "edi_exchange_state": "input_processed",
                        "parent_id": self.exchange_record.id,
                    }
                )
                self._handle_single_result(child_exchange, statement_id)
                self.exchange_record.sudo().message_post(
                    body=_("Multiple Bank Statements created (See child exchanges)")
                )
            return True
```

simahawk: Well, the formatting is a bit messed up, but I think you get my point ;)

edi_bank_statement_import/components/process.py
```diff
@@ -0,0 +1,92 @@
...
+    def _setup_wizard(self):
+        """Init a `account.statement.import` instance for current record."""
+        wiz_ctx = self.settings.get("wiz_ctx", {})
+
+        wiz_model = "account.statement.import"
+        wiz_vals = {
+            "statement_filename": self.exchange_record.exchange_filename,
+            "statement_file": self.exchange_record._get_file_content(binary=False),
+        }
+        wiz = self.env[wiz_model].with_context(wiz_ctx).sudo().create(wiz_vals)
```
simahawk, line 87: Recommendation for all the cases where you use `with_context`: never override the full ctx.
Always explode the keys to not override the rest:
```suggestion
        wiz = self.env[wiz_model].with_context(**wiz_ctx).sudo().create(wiz_vals)
```

edi_bank_statement_import/data/data.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
```
simahawk, line 2: Do you really think we need a specific backend type for this?
We should probably have a generic one into a new module `edi_bank`?
In any case, if you keep it, please move it to its own file and make it noupdate.

edi_bank_statement_import/data/data.xml
```diff
@@ -0,0 +1,24 @@
...
+    <record id="edi_exchange_backend" model="edi.backend.type">
+        <field name="name">Bank Statement Backend Type</field>
+        <field name="code">bank_statement_backend_type</field>
+    </record>
+    <record id="edi_exchange_type" model="edi.exchange.type">
```
simahawk, line 7: Creating an exchange type by default w/out noupdate is dangerous.

BTW personally I'm thinking about removing default exchange types from the modules I maintain because it can be dangerous too due to their high level of configuration.

I tend to say that you should leave this to the specific setup of the instance and move this to demo data.
Just a recommendation tho. I'd like to hear your feedback and @etobella 's too.

edi_bank_statement_import/readme/DESCRIPTION.rst
simahawk: See my comment above

edi_bank_statement_import/data/data.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="edi_exchange_backend" model="edi.backend.type">
+        <field name="name">Bank Statement Backend Type</field>
+        <field name="code">bank_statement_backend_type</field>
```
simahawk, line 5:
```suggestion
        <field name="name">Bank Statement</field>
        <field name="code">bank_statement</field>
```

simahawk (changes requested):
Thanks for your work, I left some more remarks.

PR #507, opened 2022-01-11
edits requirements.txt

Lencof:


etobella (changes requested):
Requirements is autogenerated by precommit. I don't get the neex of this

PR #522, opened 2022-02-08
[14.0] account_invoice_import_simple_pdf: support several tools for text extraction

alexis-via:
Add extract_rule position_min/position_max
Add list of fields on search_read() to improve perf

These fixes are also available in the v12 PR

### DIFF
--- a/.copier-answers.yml
+++ b/.copier-answers.yml
@@ -12,4 +12,5 @@ repo_name: edi
 repo_slug: edi
 travis_apt_packages:
 - poppler-utils
+- python3-fitz
 travis_apt_sources: []
--- a/.copier-answers.yml
+++ b/.copier-answers.yml
@@ -12,5 +12,7 @@ repo_name: edi
 repo_slug: edi
 travis_apt_packages:
 - poppler-utils
-- python3-fitz
+- mupdf-tools
+- mupdf
+- libmupdf-dev
 travis_apt_sources: []
--- a/account_invoice_import/views/res_partner.xml
+++ b/account_invoice_import/views/res_partner.xml
@@ -8,16 +8,16 @@
     <record id="view_partner_property_form" model="ir.ui.view">
         <field name="model">res.partner</field>
         <field name="inherit_id" ref="account.view_partner_property_form" />
+        <field
+            name="groups_id"
+            eval="[(4, ref('account.group_account_invoice')), (4, ref('account.group_account_readonly'))]"
+        />
         <field name="arch" type="xml">
             <button
                 name="%(base.action_res_partner_bank_account_form)d"
                 position="after"
             >
-                <div
-                    name="invoice_import_configs"
-                    groups="account.group_account_invoice,account.group_account_readonly"
-                    colspan="2"
-                >
+                <div name="invoice_import_configs" colspan="2">
                     <button
                         type="object"
                         class="btn-link"
--- a/account_invoice_import_simple_pdf/__manifest__.py
+++ b/account_invoice_import_simple_pdf/__manifest__.py
@@ -13,7 +13,10 @@
     "website": "https://github.com/OCA/edi",
     "depends": ["account_invoice_import"],
     # "excludes": ["account_invoice_import_invoice2data"],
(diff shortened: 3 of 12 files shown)


alexis-via:
I'm thinking about replacing pymupdf by pdfminer or pdfminer.six, to avoid the headaches of installing pymupdf, which is a problem for many users. pdfminer is pure-python, so no headaches to install it. It's a big decision, because all the existing users of account_invoice_import_simple_pdf will certainly have to update the setup for many vendors because the text will not be extracted in the same manner/order.
Reading this again : https://dida.do/blog/how-to-extract-text-from-pdf
they say PyPDF2 is not good for text extraction, but pdfminer is fine. pdfminer is already used by Odoo itself, cf https://github.com/odoo/odoo/blob/15.0/addons/attachment_indexation/models/ir_attachment.py#L13

@gdgellatly What do you think?
If I do the switch, should I used pdfminer or pdfminer.six ? pdfminer.six is a fork of pdfminer, but I don't know the reason.

gdgellatly:
Hi @alexis-via I'm OK either way. pymupdf via mupdf does a whole lot more (form filling for example) and easier than other python libs. But the original mupdf is pretty much dead, replaced by mupdf in java. That in itself can cause issues because you are setting JVM vars now in your deployments and you need to be careful with worker memory settings. And then just last month a new release of Pymupdf was breaking on python 3.7-3.10 so now I have to pin at 1.19.4 , but overall it is very well maintained. But I only use it for automated form filling post report processing.

Text extraction is generally via invoice2data through pdfminer and for scanned images like packing slips etc we use GVision API.

But if there are alternatives for your usecase that are already in use and work then it makes sense. It's not so much pymupdf, it is the java.

If you switch you want pdfminer.six - originally I think it was to be py2/3 compatible but original pdfminer is no longer maintained.

alexis-via:
Thanks @gdgellatly  for your detailed answer.
I think I'll add an ir.config.parameter where you can choose the extraction tool that you want to use. That way, users that have already installed the tool using pymupdf will be able to continue to use it and new users may want to go with pdfminer.six if it's difficult for them to install pymupdf. Maybe I can have a third option using pdftotext, which was the default tool in invoice2data (uses poppler-utils).

alexis-via:
I started to implement several extraction backends in addition to PyMuPDF:
- pdfminer.six
- pdfplumbing
- pypdf2
- pdftotext
So I can now compare the extraction result from different engines on the same PDF invoice.
The 2 best are definitely pdftotext and PyMuPDF. As explained in this article https://dida.do/blog/how-to-extract-text-from-pdf having an underlying PDF viewer is a big quality booster on text extraction. Reminder : invoice2data was using pdftotext.
With PyPDF2, the results are very poor.
With pdfminer.six, the main problem is with tables which are badly extracted, as explained in https://stackoverflow.com/questions/46222918/how-to-extract-tables-from-a-pdf-with-pdfminer
With pdfplumber (pure python, based on pdfminer.six) when using pdf_page.extract_text(layout=True) the result on tables is qood. But I can't make detect_vertical=True work, so it doesn't extract correctly the vertical lines that the document may contain.

Conclusion: adding support for pdftotext may be interesting, but it's not pure-python: it requires poppler. pdfplumber could be an option if I manage to get  detect_vertical=True to work. The others would not do a good job.

alexis-via:
I managed to find a good configuration for pdfplumber, so we now have a pure-python option. It's not as good as pymupdf nor pdftotext, but it works. I didn't implement pypdf2, because the results are too bad. So we now have 3 main options: pymupdf, pdftotext (cmd or lib) and pdfplumber.

alexis-via:
The implementation of the multi-text extraction method is now ready to be merged. I updated the README accordingly. The test solution on the partner form shows which method has been used. It is possible to force a particular method via an ir.config_parameter. Everything is explained in the README.

I plan to merge this PR soon because OCA/edi v14 branch is red due to a problem in the install of the latest version of pymupdf. This PR will make the branch green again.
@gdgellatly what do you think about the implementation?

TDu (approved):

alexis-via:
All the OCA projects that include OCA/edi are blocked because of fitz, so I merge it rapidly.

PR #523, opened 2022-02-08
[13.0][FIX] edi: remove span inside anchor to fix exchange record navigation

JordiMForgeFlow:
By adding the span element, the navigation is not performed correctly and the anchor just redirects to '#'.

### DIFF
--- a/edi/templates/exchange_chatter_msg.xml
+++ b/edi/templates/exchange_chatter_msg.xml
@@ -13,7 +13,7 @@
                     t-att-data-oe-model="exchange_record._name"
                     t-att-data-oe-id="exchange_record.id"
                 >
-                    <span t-field="exchange_record.name" />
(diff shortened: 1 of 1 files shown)


JordiMForgeFlow:
CC @LoisRForgeFlow @simahawk

LoisRForgeFlow (approved):
Thanks for tracking it down :raised_hands:

etobella (approved):

simahawk:
@JordiMForgeFlow tnx for the fix. Next time, pls, keep the msg short and move details to the description of the commit. 
Also, pls, do not include the odoo version. Eg

```
edi_oca: fix chatter msg anchor

By adding the span element, the navigation is not performed correctly and the anchor just redirects to '#'.
Just remove it.
```

PR #526, opened 2022-02-14
[13.0][ADD] edi_purchase_oca

LoisRForgeFlow:
This module intends to create a base to be extended by local edi rules
for purchase.

In order to add a new integration, you need to create a listener:

.. code-block:: python

    class MyEventListener(Component):
        _name = "purchase.order.event.listener.demo"
        _inherit = "base.event.listener"
        _apply_on = ["purchase.order"]

        def on_button_confirm_purchase_order(self, move):
            """Add your code here"""

@ForgeFlow


edi_purchase/__manifest__.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2022 ForgeFlow S.L. (https://www.forgeflow.com)
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+{
+    "name": "Edi Purchase",
+    "summary": """
+        Define EDI Configuration for Purchase Orders""",
+    "version": "13.0.1.1.0",
```
JordiMForgeFlow:
```suggestion
    "version": "13.0.1.0.0",
```

JordiMForgeFlow (commented):
Apart from this all LGTM 👍🏼 (code + functional review)

edi_purchase/__manifest__.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2022 ForgeFlow S.L. (https://www.forgeflow.com)
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+{
+    "name": "Edi Purchase",
```
simahawk: EDI

edi_purchase/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,14 @@
+This module intends to create a base to be extended by local edi rules
+for purchase.
+
+In order to add a new integration, you need to create a listener:
```
simahawk: This is not required in theory. Can you provide a use case?

edi_purchase/views/edi_exchange_record_views.xml
```diff
@@ -0,0 +1,30 @@
...
+        name="Exchange records"
+        parent="purchase.menu_purchase_root"
+        sequence="20"
+    />
+    <menuitem
+        id="menu_purchase_edi_exchange_record"
+        name="Moves"
+        action="act_open_edi_exchange_record_purchase_order_view"
+        parent="menu_purchase_edi_root"
+        sequence="20"
```
simahawk: the icon is not needed apparently. Also, if you need it, you can reuse the one from base_edi

simahawk (changes requested):

edi_purchase/readme/DESCRIPTION.rst
LoisRForgeFlow: Mmm why not? how do you consume the listener otherwise. It is the same idea than edi_account that is consumed [here](https://github.com/OCA/l10n-spain/blob/14.0/l10n_es_facturae/components/account_move_l10n_es_facturae_listener.py#L22) for instance.

edi_purchase/views/edi_exchange_record_views.xml
LoisRForgeFlow: The icon is for the app in the kanban view of modules, not for the menus. Am I missing something?

edi_purchase/__manifest__.py
LoisRForgeFlow: Done!

etobella (approved):
It seems right for me, it follows the strategy we have on `edi_account_oca`

Can you rename it as `edi_purchase_oca`? Just to keep the name on the 14 migration :rofl:

simahawk (approved):
+1 for changing the name :cry:

LoisRForgeFlow:
@etobella @simahawk Rename done :angel:

PR #529, opened 2022-02-16
[14.0] Add sale_order_packaging_import

TDu:
This module extends the `sale_order_import` module to include on the sale order line
the import of the product packaging and uses the quantity imported has the
the packaging quantity.

Also two function hooks have been added on the depending modules

### DIFF
--- a/base_business_document_import/models/business_document_import.py
+++ b/base_business_document_import/models/business_document_import.py
@@ -985,25 +985,9 @@ class BusinessDocumentImport(models.AbstractModel):
                 # used for to_remove
                 existing_lines_dict[product]["import"] = True
                 oline = existing_lines_dict[product]["line"]
-                res["to_update"][oline] = {}
-                if float_compare(
-                    iline["qty"],
-                    existing_lines_dict[product]["qty"],
-                    precision_digits=qty_precision,
-                ):
-                    res["to_update"][oline]["qty"] = [
-                        existing_lines_dict[product]["qty"],
-                        iline["qty"],
-                    ]
-                if "price_unit" in iline and float_compare(
-                    iline["price_unit"],
-                    existing_lines_dict[product]["price_unit"],
-                    precision_digits=price_precision,
-                ):
-                    res["to_update"][oline]["price_unit"] = [
-                        existing_lines_dict[product]["price_unit"],
-                        iline["price_unit"],
-                    ]
+                res["to_update"][oline] = self._prepare_order_line_update_values(
+                    existing_lines_dict[product], iline, qty_precision, price_precision
+                )
             else:
                 res["to_add"].append(
                     {"product": product, "uom": uom, "import_line": iline}
@@ -1015,6 +999,27 @@ class BusinessDocumentImport(models.AbstractModel):
                 else:
                     res["to_remove"] = exiting_dict["line"]
         return res
+
+    def _prepare_order_line_update_values(
+        self, existing_line, iline, qty_precision, price_precision
+    ):
+        values = {}
+        if float_compare(
+            iline["qty"],
+            existing_line["qty"],
+            precision_digits=qty_precision,
+        ):
+            values["qty"] = [existing_line["qty"], iline["qty"]]
+        if "price_unit" in iline and float_compare(
+            iline["price_unit"],
+            existing_line["price_unit"],
+            precision_digits=price_precision,
+        ):
+            values["price_unit"] = [
+                existing_line["price_unit"],
+                iline["price_unit"],
+            ]
+        return values
 
     def _prepare_account_speed_dict(self):
         company_id = self._context.get("force_company") or self.env.company.id
--- a/sale_order_packaging_import/models/business_document_import.py
+++ b/sale_order_packaging_import/models/business_document_import.py
@@ -1,0 +1,34 @@
+# Copyright 2022 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+from odoo import models
+
+
+class BusinessDocumentImport(models.AbstractModel):
+    _inherit = "business.document.import"
+
+    def _prepare_order_line_update_values(
+        self, existing_line, iline, qty_precision, price_precision
+    ):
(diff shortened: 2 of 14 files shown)


base_business_document_import/models/business_document_import.py
```diff
@@ -1016,6 +1000,27 @@ def compare_lines(
...
+        self, existing_line, iline, qty_precision, price_precision
+    ):
+        values = {}
+        if float_compare(
+            iline["qty"],
+            existing_line["qty"],
+            precision_digits=qty_precision,
+        ):
+            values["qty"] = [existing_line["qty"], iline["qty"]]
+            if "price_unit" in iline and float_compare(
```
simahawk: this IF was not nested before, correct?

sale_order_packaging_import/models/business_document_import.py
```diff
@@ -0,0 +1,34 @@
...
+            if packaging:
+                qty = packaging.qty * iline["qty"]
+                if qty != oline.product_uom_qty:
+                    values["qty"] = [oline.product_uom_qty, qty]
+                elif "qty" in values.keys():
+                    values.pop("qty")
+        if packaging and packaging != oline.product_packaging:
+            values["packaging"] = [oline.product_packaging.id, packaging.id]
+        elif not packaging and oline.product_packaging:
+            values["packaging"] = [oline.product_packaging.id, False]
```
simahawk, line 33: can't you collect these as dict? Would more clear IMO. Also, what's the point of keeping old and new value?

sale_order_packaging_import/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,3 @@
+This module extends the `sale_order_import` module to include on the sale orde line
```
simahawk: worth to mention that you assume that the GTIN provided is the one of the packaging instead of the product one

simahawk (changes requested):

simahawk:
Unrelated error:
```
2022-02-16 13:58:32,957 996 ERROR odoo odoo.addons.account_invoice_import.tests.test_invoice_import: ERROR: TestInvoiceImport.test_email_gateway
  File "/__w/edi/edi/setup/account_invoice_import/odoo/addons/account_invoice_import/tests/test_invoice_import.py", line 161, in test_email_gateway
    ).message_process("account.invoice.import", mail)
  File "/opt/odoo/addons/mail/models/mail_thread.py", line 1145, in message_process
    thread_id = self._message_route_process(message, msg_dict, routes)
  File "/opt/odoo/addons/mail/models/mail_thread.py", line 1056, in _message_route_process
    thread = ModelCtx.message_new(message_dict, custom_values)
  File "/__w/edi/edi/setup/account_invoice_import/odoo/addons/account_invoice_import/wizard/account_invoice_import.py", line 1390, in message_new
    msg_dict["email_to"],
KeyError: 'email_to'
```

Also unrelated... but important... can't we silent these logs? 

https://github.com/OCA/edi/runs/5217188937?check_suite_focus=true#step:8:1144
https://github.com/OCA/edi/runs/5217188937?check_suite_focus=true#step:8:1035
CC @alexis-via

base_business_document_import/models/business_document_import.py
TDu: Indeed, thank you.

sale_order_packaging_import/models/business_document_import.py
TDu: About old/new values this is how it is done in the other module, I would think mainly to inform the user accordingly in the chatter.

sale_order_packaging_import/readme/DESCRIPTION.rst
TDu: I updated the description, I did not mention gtin but gave a better explanation of how it works

sale_order_packaging_import/models/business_document_import.py
mmequignon: Also, if a recordset is empty, recordset.id is `False`.
```python
>>> self.env["product.product"].id
False
```
So, couldn't we merge those ifs?
```python
if oline.product_packaging:
    values["packaging"] = [oline.product_packaging.id, packaging.id]
```

TDu: We could but we may loose in readability IMO

simahawk:
> Unrelated error:
> 
> ```
> 2022-02-16 13:58:32,957 996 ERROR odoo odoo.addons.account_invoice_import.tests.test_invoice_import: ERROR: TestInvoiceImport.test_email_gateway
> Traceback (most recent call last):
>   File "/__w/edi/edi/setup/account_invoice_import/odoo/addons/account_invoice_import/tests/test_invoice_import.py", line 161, in test_email_gateway
>     ).message_process("account.invoice.import", mail)
>   File "/opt/odoo/addons/mail/models/mail_thread.py", line 1145, in message_process
>     thread_id = self._message_route_process(message, msg_dict, routes)
>   File "/opt/odoo/addons/mail/models/mail_thread.py", line 1056, in _message_route_process
>     thread = ModelCtx.message_new(message_dict, custom_values)
>   File "/__w/edi/edi/setup/account_invoice_import/odoo/addons/account_invoice_import/wizard/account_invoice_import.py", line 1390, in message_new
>     msg_dict["email_to"],
> KeyError: 'email_to'
> ```
> 
> Also unrelated... but important... can't we silent these logs?
> 
> https://github.com/OCA/edi/runs/5217188937?check_suite_focus=true#step:8:1144 https://github.com/OCA/edi/runs/5217188937?check_suite_focus=true#step:8:1035 CC @alexis-via

@alexis-via ping :)

simahawk:
The reason why tests are failing here https://github.com/OCA/edi/blob/14.0/account_invoice_import/wizard/account_invoice_import.py#L1370 is because that method seems to be quite fragile and not fully tested.

This new module adds a dependency on `sale_stock` which pulls `stock` which creates a new company.
I'm working on a fix.

simahawk:
fix here https://github.com/OCA/edi/pull/598

simahawk (approved):

PR #490, opened 2021-12-05
[MIG][15.0] edi_oca

i-vyshnevska:
Depends on

- [x] https://github.com/OCA/connector/pull/425

PR #530, opened 2022-02-18
[MIG] edi_oca: Migration to 15.0

etobella:
Superseeds #490. Once it is green, it can be merged (only small changes from original migration to fix travis and runboat)


etobella:
I will proceed with the merge, as all the approvals exists on the original PR
/ocabot merge nobump

PR #532, opened 2022-02-21
[14.0] endpoint: block all RPC calls

simahawk:
Depends on 
- [x] https://github.com/OCA/server-tools/pull/2293

### DIFF
--- a/endpoint/migrations/14.0.1.1.0/pre-migrate.py
+++ b/endpoint/migrations/14.0.1.1.0/pre-migrate.py
@@ -1,0 +1,25 @@
+# Copyright 2022 Camptocamp SA (http://www.camptocamp.com)
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+import logging
+
+from odoo import SUPERUSER_ID, api
+
+_logger = logging.getLogger(__name__)
+
+
+def migrate(cr, version):
+    if not version:
+        return
+
+    env = api.Environment(cr, SUPERUSER_ID, {})
+    module = env["ir.module.module"].search(
+        [
+            ("name", "=", "rpc_helper"),
+            ("state", "=", "uninstalled"),
+        ]
+    )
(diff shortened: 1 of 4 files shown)


mmequignon (approved):

endpoint/migrations/14.0.1.1.0/pre-migrate.py
```diff
@@ -0,0 +1,25 @@
...
+    env = api.Environment(cr, SUPERUSER_ID, {})
+    module = env["ir.module.module"].search(
+        [
+            ("name", "=", "rpc_helper"),
+            ("state", "=", "uninstalled"),
+        ]
+    )
+    if module:
+        _logger.info("Install module rpc_helper")
+        module.write({"state": "to install"})
```
etobella, line 24: Isn't this done automatically?

simahawk: yes :)

etobella: So, you don't need the migration script, isn't it?

simahawk: I got your question wrong then. Without this you won't get the module installed automatically because it's a new dependency. AFAIK Odoo won't install new dependencies automatically.

etobella: I usually add more dependencies and if i load as -u it works as expected. If you make -i it will not install it

flotho: Hi @etobella , is it a change request?

etobella: Not really, code will never be executed because odoo will mark it as to install before this is executed. It is not blocking

etobella (approved):

PR #542, opened 2022-03-15
[14.0][IMP] oca_edi: child exchange compute file from parent

TeoGoddet:
As discussed here : https://github.com/OCA/edi/pull/506#discussion_r793324325
When a child exchange has not an explicit file set, we compute it from the parents, it allow to handle cases where a single document exchange need to create multiple objects. (Such as multiple bank statement in a single xml file)

### DIFF
--- a/edi_oca/migrations/14.0.1.9.0/pre-migration.py
+++ b/edi_oca/migrations/14.0.1.9.0/pre-migration.py
@@ -1,0 +1,15 @@
+# Copyright 2022 Téo Goddet
+
+from openupgradelib import openupgrade
+
+_column_renames = {
+    "edi_exchange_record": [("exchange_file", "real_exchange_file")],
+}
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    if not openupgrade.column_exists(
+        env.cr, "edi_exchange_record", "real_exchange_file"
+    ):
+        openupgrade.rename_columns(env.cr, _column_renames)
--- a/edi_oca/models/edi_exchange_record.py
+++ b/edi_oca/models/edi_exchange_record.py
@@ -39,9 +39,21 @@ class EDIExchangeRecord(models.Model):
         model_field="model",
     )
     related_name = fields.Char(compute="_compute_related_name", compute_sudo=True)
-    exchange_file = fields.Binary(attachment=True)
+    # here we store exchange file when needed (when no parent hold the file)
+    real_exchange_file = fields.Binary(attachment=True)
+    # here we give the exchange file (from parent exchange or self)
+    exchange_file = fields.Binary(
+        store=False,
+        compute="_compute_exchange_file",
+        inverse="_inverse_exchange_file",
+    )
+    has_exchange_file = fields.Boolean(
+        store=True,
+        compute="_compute_exchange_file",
+        help="True if the exchange_file is set on the record or on any parent record",
+    )
     exchange_filename = fields.Char(
-        compute="_compute_exchange_filename", readonly=False, store=True
+        store=True, compute="_compute_exchange_filename", readonly=False
     )
     exchanged_on = fields.Datetime(
         string="Exchanged on",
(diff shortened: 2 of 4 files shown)


edi_oca/models/edi_exchange_record.py
```diff
@@ -38,7 +38,9 @@ class EDIExchangeRecord(models.Model):
         readonly=True,
         model_field="model",
     )
-    exchange_file = fields.Binary(attachment=True)
+    exchange_file = fields.Binary(
+        store=True, compute="_compute_exchange_file", attachment=True
+    )
```
simahawk: These changes break normal behavior. Also, in this way you store 2 or N times the same file for all records in the hierarchy.
We should probably have another field which is not stored which computes the value from the current record or from the parent or not touch the field and rely only on `_get_file_content`. In that case, it should be checked every place where we use the `.exchange_file` directly.

simahawk (changes requested):

edi_oca/models/edi_exchange_record.py
TeoGoddet: Hi ! I don't really see why it broke the normal behavior ? 
Also for the duplication of the file i'm aware, but the other way are way more complex, so i'm not sure what is the best

TeoGoddet:
Hi ! @simahawk, i'm struggling with this as the exchange_file field is expected to be searcheable
https://github.com/OCA/edi/blob/b2fcc1c07f428125b6f0d21a94f416df4d55f5fe/edi_oca/models/edi_backend.py#L386

I see two solutions : 
 - store the attachement on each record
 - store a boolean true if the exchange file is present on the record or on any parent

I'm trying solution 2

edi_oca/models/edi_exchange_record.py
TeoGoddet: Ok, see next comments

TeoGoddet:
Finally have so something working,
Please tell me what do you think of this proposal

simahawk:
Hi there, sorry for the late reply but I got COVID and then I got holidays (luckily :smile: ). 
I'll get back to this in the next few days as I have some EDI-related work to do.
Anyway, at first sight looks better, thanks :wink:

edi_oca/migrations/14.0.1.9.0/pre-migration.py
```diff
@@ -0,0 +1,15 @@
...
+
+_column_renames = {
+    "edi_exchange_record": [("exchange_file", "real_exchange_file")],
+}
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    if not openupgrade.column_exists(
+        env.cr, "openupgrade.rename_columns", "real_exchange_file"
```
etobella: Why do you need to rename if it is an attachment field ???

edi_oca/models/edi_exchange_record.py
```diff
@@ -39,9 +39,22 @@ class EDIExchangeRecord(models.Model):
...
     related_name = fields.Char(compute="_compute_related_name", compute_sudo=True)
-    exchange_file = fields.Binary(attachment=True)
+    # here we store exchange file when needed (when no parent hold the file)
+    real_exchange_file = fields.Binary(attachment=True)
+    # here we give the exchange file (from parent exchange or self)
+    exchange_file = fields.Binary(
+        store=False,
+        compute="_compute_exchange_file",
+        inverse="_inverse_exchange_file",
+        attachment=True,
```
etobella: No need of attachment as it has compute and inverse...
store false is not necessary

edi_oca/models/edi_exchange_record.py
```diff
@@ -116,12 +129,30 @@ def _compute_related_name(self):
...
+            rec.has_exchange_file = True
+
+    @api.depends("model", "type_id", "parent_id")
     def _compute_exchange_filename(self):
         for rec in self:
-            if not rec.type_id:
+            if rec.exchange_filename:
                 continue
-            if not rec.exchange_filename:
+            if not rec.real_exchange_file and rec.parent_id.exchange_file:
```
etobella, line 152: IMO, this should only happen in some kind of backends or backend types, so I would add a condition there.

etobella (changes requested):
I don't understand why do you need to make the real_exchange_file and exchange_file fields

edi_oca/models/edi_exchange_record.py
simahawk: TBH I don't know what's best here: shall we store the filename on all children? Probably not.

edi_oca/migrations/14.0.1.9.0/pre-migration.py
simahawk: if you make the current field not stored you need another field and migrate previous values, no?

etobella: Well, in any case, this script has no sense openupgade.rename_columns is not a table...

Also, I don't understand why he is creating this new field, some context might clarify things

edi_oca/models/edi_exchange_record.py
TeoGoddet: This is pretty advanced orm knowledge I don't have, thanks

edi_oca/migrations/14.0.1.9.0/pre-migration.py
TeoGoddet: My first migration, sorry for the bad quality, 
I still think that this is needed

the new field is created as requested by simahawk, it allow the exchange_file to be computed from the parent or expliciteky set on the record. If you see a better way to do this, tell me :)

edi_oca/models/edi_exchange_record.py
TeoGoddet: I dont know, I can have the same thing that we have for the exchange file but it has constraints (harder to search on it, ...) and the db footprint is reasonable (to me, i may be wrong)

TeoGoddet: @etobella maybe the processor should be responsible for this ? Or applying it only to in exchange ?

etobella:
@TeoGoddet Can you rebase it please?

PR #546, opened 2022-03-18
[ADD] edi_stock_oca

albariera:
This module intends to create a base to be extended by local edi rules
for stock.


etobella (approved):

edi_stock_oca/models/stock_picking.py
```diff
@@ -0,0 +1,26 @@
...
+
+from odoo import fields, models
+
+
+class StockPicking(models.Model):
+    _name = "stock.picking"
+
+    _inherit = ["stock.picking", "edi.exchange.consumer.mixin"]
+
+    disable_edi_auto = fields.Boolean(
```
simahawk: Question: would be interesting to have this on the consumer mixin?
Also, matter of taste maybe but I'd prefer to prefix such fields w/ `edi`. In this case I vote for `edi_auto_disabled`.

edi_stock_oca/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,14 @@
...
+In order to add a new integration for an stock picking, you need to create a listener:
+
+.. code-block:: python
+
+    class MyEventListener(Component):
+        _name = "stock.picking.event.listener.demo"
+        _inherit = "base.event.listener"
+        _apply_on = ["stock.picking"]
+
+        def on_validate_stock_picking(self, picking):
```
simahawk:
```suggestion
        def on_validate(self, picking):
```
Seems redundant otherwise.

edi_stock_oca/views/stock_picking.xml
```diff
@@ -0,0 +1,37 @@
...
+<!-- Copyright 2022 OCA
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
+<odoo>
+
+    <record model="ir.ui.view" id="stock_picking_form_view">
+        <field name="name">stock.picking.form (in edi_stock_oca)</field>
+        <field name="model">stock.picking</field>
+        <field name="inherit_id" ref="stock.view_picking_form" />
+        <field name="arch" type="xml">
+            <group name="other_infos" position="after">
```
simahawk: Wouldn't be better to have a tab for EDI?

simahawk (changes requested):
@SilvioC2C you might be interested in this

edi_stock_oca/models/stock_picking.py
etobella: Well, the field can be added on edi.exchange.mixin from edi_oca, but, at the end, we will allways need to inherit on the models in order to set the states and so on :thinking: 

I agree with the change of name and move to `edi`, but it is already being used from l10n-spain facturae (at least...) could we add a comment to rename it on 16?

simahawk: IMO as this module is new this is the best time to change its name before is adopted by a lot of projects.

I propose this: 

1. change it here
2. for Spain localization, add `disable_edi_auto = fields.Boolean(related="edi_auto_disabled", readonly=False)`

Then you get rid of it when you want on Spain loc. Would it work for you?

If not, I won't keep it till v16. We are on v13 here, you can change this on v14, no?

etobella: :laughing: I thought the same, but I didn't want to make it too hard :laughing:

Let's do this then

edi_account/models/account_move.py
```diff
@@ -8,12 +8,13 @@ class AccountMove(models.Model):
...
 
-    disable_edi_auto = fields.Boolean(
-        help="When marked, EDI could be avoided",
-        readonly=True,
+    edi_auto_disabled = fields.Boolean(
         states={"draft": [("readonly", False)]},
     )
 
+    disable_edi_auto = fields.Boolean(related="edi_auto_disabled")
+    # TODO: remove this field at version 16.0
```
simahawk, line 14: why not 14.0?

edi_stock_oca/tests/test_edi.py
```diff
@@ -0,0 +1,55 @@
+# Copyright 2022 Creu Blanca
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+import logging
+
+from odoo.addons.component.core import Component
+from odoo.addons.component.tests.common import SavepointComponentRegistryCase
+from odoo.addons.stock.tests.common import TestStockCommon
+
+_logger = logging.getLogger(__name__)
```
simahawk: logger for?

edi_stock_oca/views/stock_picking.xml
simahawk: ping :)

albariera: Yes! Sorry, I thought I had already changed it. Done :)

edi_stock_oca/tests/test_edi.py
albariera: Removed 👍

edi_account/models/account_move.py
albariera: Because edi_account is already migrated to 14.0 and 15.0. I was leaving it just for security reasons, maybe someone has already custom modules based on this module at those versions. Do you think it is better to remove it at 14.0?

simahawk: IMO: yes. For existing installations you probably need a small migration script (the did here is ok).
What you can do: define a post install hook that can be used later for upgrades too (you can import an hook from the `migrations` folder but you cannot import vice versa).  Up to you... I don't have a full background on how and where you are using this already... you guys know better ;)

simahawk (approved):

simahawk:
@albariera pre-commit to be fixed. Also, try to keep commit msg shorter (move a long description to the commit long desc :wink: )

etobella:
The precommit error is solved by updating copier, but several issues were found with some other modules #562

olgamarcocb (approved):
LGTM :+1:

edi_stock_oca/models/stock_picking.py
```diff
@@ -0,0 +1,24 @@
...
+            "waiting": [("readonly", False)],
+            "confirmed": [("readonly", False)],
+            "assigned": [("readonly", False)],
+        },
+    )
+
+    def button_validate(self):
+        res = super().button_validate()
+        self._event("on_validate").notify(self)
+        return res
```
LoisRForgeFlow: Wouldn't it be better to listen at `action_done`?, in button_validate there are several events that can pop before actually doing the picking (backorder wizard, no quantity done wizard, etc.).

albariera: You are right. Changed! Thanks :)

etobella: @LoisRForgeFlow Can you recheck it please :smile: ?

LoisRForgeFlow (approved):
LGTM, thanks

PR #540, opened 2022-03-10
[14.0] edi: misc fix/imp

simahawk:
Please, see atomic commits for details.
Most important changes:

* fixed the ordering which was just random due to the override of `_search` CC @etobella 
* added many handy filters and additional menu items
* moved q.job relation from backend to exchange record (see motivation in its commit)

PR #551, opened 2022-03-23
[FIX] edi_oca: Restore removed functions

etobella:
We want to keep functionality on the same version. It is marked to remove for 16

#540

### DIFF
--- a/edi_oca/models/edi_backend.py
+++ b/edi_oca/models/edi_backend.py
@@ -179,6 +179,14 @@ class EDIBackend(models.Model):
             ("backend_id", "=", self.id),
         ]
 
+    def _delay_action(self, rec):
+        # TODO: Remove this on 16.0
+        _logger.warning(
+            "This function has been replaced by rec.with_delay(). "
+            "It will be removed on 16.0."
+        )
+        return self.with_delay(**rec._job_delay_params())
+
     def exchange_generate(self, exchange_record, store=True, force=False, **kw):
         """Generate output content for given exchange record.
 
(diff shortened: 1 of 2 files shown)


edi_oca/models/edi_backend.py
```diff
@@ -179,6 +179,21 @@ def _get_exchange_type_domain(self, code):
...
 
+    def _get_job_delay_params(self, exchange_record):
+        params = {}
+        channel = exchange_record.type_id.job_channel_id
+        if channel:
+            params["channel"] = channel.complete_name
+        return params
+
+    def _delay_action(self, rec):
+        # TODO: Remove this on 16.0
```
simahawk, line 183: Can you please log a deprecation warning?
People should switch to the new approach as soon/far as possible.

edi_oca/models/edi_backend.py
```diff
@@ -179,6 +179,21 @@ def _get_exchange_type_domain(self, code):
             ("backend_id", "=", self.id),
         ]
 
+    def _get_job_delay_params(self, exchange_record):
+        params = {}
+        channel = exchange_record.type_id.job_channel_id
+        if channel:
+            params["channel"] = channel.complete_name
```
simahawk: here you can take values from the new exchange record method instead

simahawk (changes requested):

etobella:
Comments attended

etobella:
Comments attended

simahawk (approved):
thanks!

PR #541, opened 2022-03-15
[15.0][PORT] edi: misc fix/imp

ivantodorovich:
Port of this (still open) PR:
- https://github.com/OCA/edi/pull/540

Please review that one first 🙏🏻 

@simahawk @etobella

PR #552, opened 2022-03-23
[15.0] edi: misc fix/imp

etobella:
All commits from #540 #541

### DIFF
--- a/edi_oca/models/edi_exchange_record.py
+++ b/edi_oca/models/edi_exchange_record.py
@@ -19,7 +19,6 @@ class EDIExchangeRecord(models.Model):
     _order = "exchanged_on desc"
     _rec_name = "identifier"
 
-    name = fields.Char(compute="_compute_name")
     identifier = fields.Char(required=True, index=True, readonly=True)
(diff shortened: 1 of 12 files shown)


edi_oca/models/edi_exchange_record.py
```diff
@@ -108,12 +109,11 @@ class EDIExchangeRecord(models.Model):
         ),
     ]
 
-    @api.depends("type_id.code", "model", "res_id")
-    def _compute_name(self):
+    @api.depends("model", "res_id")
```
i-vyshnevska, line 112:
```suggestion
    @api.depends("record.display_name")
```
Afais previously instead of an empty name we set to default "Unrelated" value, but now when it is empty we can use directly related field

simahawk: you cannot depend on `record`. It's a property. We should have a Reference field but I don't like to store the same info twice. Also, I removed "Unrelated" because: it made poor sense overall and no sense at all when the exchange does not touch any real record -> no relation needed.
When you have a relation, you already see that on the form.

simahawk (approved):

PR #556, opened 2022-04-05
edi_storage_oca: manage listeners for outgoing exchanges

SilvioC2C:


edi_storage_oca/models/edi_exchange_record.py
```diff
@@ -0,0 +1,17 @@
+# Copyright 2022 Camptocamp SA
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+from odoo import models
+
+
+class EDIExchangeRecord(models.Model):
+    _inherit = "edi.exchange.record"
+
+    def _get_dirs(self):
```
simahawk: please:
1. move it to edi.backend
2. rename it to `_storage_get_dirs`
3. pass the direction to it

simahawk (changes requested):

SilvioC2C:
@simahawk tests added but I ain't 100% sure about them: please feel free to review and leave suggestions

simahawk (commented):
LG, pls squash

simahawk (approved):
@LoisRForgeFlow mind to review?

ivantodorovich:
I seem to be missing something 😅 
I thought output directories were meant for the external service to find/process/move records. (opposite than input)
How come we need to move files around there? If there's an error generating the file, we simply not upload it in the first place, no? And it's the external service the one that needs to move the file to either Done or Error after it processes it 🤔  ?

simahawk:
agree

PR #561, opened 2022-04-12
[15.0] [IMP] edi_storage_oca: add a common directory_path per backend

ivantodorovich:
And make it configurable through server env.

This gives more flexibility to configure paths through env.

Previous to this commit, the only way to do that was by configuring the `storage.backend.directory_path`, but then it means we can't reuse the `storage.backend` for multiple `edi.backend` records if their relative paths don't share the same root.. 

It's usage is optional.

### DIFF
--- a/edi_storage_oca/components/listener.py
+++ b/edi_storage_oca/components/listener.py
@@ -30,30 +30,37 @@ class EdiStorageListener(Component):
         )
 
     def on_edi_exchange_done(self, record):
-        storage = record.backend_id.storage_id
         res = False
-        if record.direction == "input" and storage:
-            file = record.exchange_filename
-            pending_dir = record.backend_id.input_dir_pending
-            done_dir = record.backend_id.input_dir_done
-            error_dir = record.backend_id.input_dir_error
-            if not done_dir:
-                return res
-            res = self._move_file(storage, pending_dir, done_dir, file)
+        backend = record.backend_id
+        storage = backend.storage_id
+        if record.direction == "input" and storage and backend.input_dir_done:
+            res = self._move_file(
+                storage,
+                backend._storage_full_path(backend.input_dir_pending),
+                backend._storage_full_path(backend.input_dir_done),
+                record.exchange_filename,
+            )
             if not res:
                 # If a file previously failed it should have been previously
                 # moved to the error dir, therefore it is not present in the
                 # pending dir and we need to retry from error dir.
-                res = self._move_file(storage, error_dir, done_dir, file)
+                res = self._move_file(
+                    storage,
+                    backend._storage_full_path(backend.input_dir_error),
+                    backend._storage_full_path(backend.input_dir_done),
+                    record.exchange_filename,
+                )
         return res
 
     def on_edi_exchange_error(self, record):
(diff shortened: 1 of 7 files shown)


edi_storage_oca/components/listener.py
```diff
@@ -30,30 +30,37 @@ def _add_after_commit_hook(self, move_func, sftp_filepath, sftp_destination_path
...
         res = False
-        if record.direction == "input" and storage:
-            file = record.exchange_filename
-            pending_dir = record.backend_id.input_dir_pending
-            done_dir = record.backend_id.input_dir_done
-            error_dir = record.backend_id.input_dir_error
-            if not done_dir:
-                return res
-            res = self._move_file(storage, pending_dir, done_dir, file)
+        backend = record.backend_id
```
simahawk, line 34: can you please review this so that we can merge it before?
https://github.com/OCA/edi/pull/556
It contains a small refactoring of this part.

edi_storage_oca/models/edi_backend.py
```diff
@@ -12,8 +12,8 @@
 
 
 class EDIBackend(models.Model):
-
-    _inherit = "edi.backend"
+    _name = "edi.backend"
+    _inherit = ["edi.backend", "server.env.mixin"]
```
simahawk, line 16: I wouldn't do this here IMO (maybe edi_storage_server_env?) and we should use the tech_name mixin.

edi_storage_oca/models/edi_backend.py
```diff
@@ -37,6 +37,9 @@ class EDIBackend(models.Model):
     where A and B are the partners exchanging data and they are in turn
     sender and receiver and vice versa.
     """
+    directory_path = fields.Char(
```
simahawk, line 40:
```suggestion
    storage_root_path = fields.Char(
```

simahawk (changes requested):

edi_storage_oca/models/edi_backend.py
ivantodorovich: `storage_backend` is already adding the `server_environment` dependency, so do we really need a separate module?
Is there a downside in doing this?

edi_storage_oca/components/listener.py
ivantodorovich: I'm not sure we need that PR: https://github.com/OCA/edi/pull/556#issuecomment-1097929628

ivantodorovich:
After giving it a little bit more though, this PR helps but what we really need to be able to reuse the `edi.backend` is to be able to set the paths on `edi.exchange.type`, not here.

`edi.exchange.type` already holds the filename pattern, so it seems to be the good place to also hold the dir paths.
Different exchange types can share the same backend, even if they watch/push files on different directories.

I guess at some point this was considered, according to these comments: https://github.com/OCA/edi/blob/1dbca793600510777694a9848cb2432150eb937d/edi_storage_oca/models/edi_backend.py#L40-L41

What I'm not sure is if we can do this change in `15.0`. Even with a migration script, it will probably need some changes from people using it, if they're setting these fields through custom module data for example.

PR #562, opened 2022-04-19
Updated Copier from template

etobella:
From Travis to GH Actions


etobella:
@ernestotejeda Some of your modules have problems with the development status. Some of them comes from other repos (like sale_order_secondary_unit). I will propose the promotion for all modules...

olgamarcocb (approved):
LGTM :+1:

base_edi/__manifest__.py
```diff
@@ -5,8 +5,8 @@
 {
     "name": "Base EDI",
     "summary": """Base module to aggregate EDI features.""",
-    "version": "13.0.1.0.2",
-    "development_status": "Alpha",
+    "version": "13.0.2.0.0",
```
simahawk: IMO all these changes do not deserve a major release as they are backward compat...

simahawk (approved):

base_edi/__manifest__.py
etobella: True, changed

PR #575, opened 2022-04-27
fix(account.invoice.import): reemplaza supplier_invoice_number por re…

fedexin40:
…ference

De acuerdo a los últimos cambios realizados en el módulo de account el campo supplier_invoice_number
fue depreciado


simahawk (changes requested):
Thanks for your contrib :)

Please, use English and rewrite your commits to be shorter and keep the explanation inside the commit description.

Eg:

```
account_invoice_import: fix invoice reference value

bla bla bla explanation here
```
```
account_invoice_import: fix invoice onchange

bla bla
```

Also, note that `play_onchanges` is available only when using `onchange_helper` module which is not the case here.

PR #578, opened 2022-05-06
edi_storage: path configurable by type and param

simahawk:
TODO

- [x] add tests

PR #582, opened 2022-05-10
[15.0] edi_storage: path configurable by type and param

jcoux:
Backported on https://github.com/OCA/edi/pull/578

By setting specific storage path on exchange type advanced settings, we
can configure specific path for each exchange type (input or output).

Path can also be configured by config parameters to allow to easily set
different values depending of the environment.

### DIFF
--- a/edi_storage_oca/components/base.py
+++ b/edi_storage_oca/components/base.py
@@ -1,10 +1,13 @@
 # Copyright 2020 ACSONE
+# Copyright 2022 Camptocamp
 # @author: Simone Orsi <simahawk@gmail.com>
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
-
+import logging
 from pathlib import PurePath
 
 from odoo.addons.component.core import AbstractComponent
+
+_logger = logging.getLogger(__file__)
 
 
 class EDIStorageComponentMixin(AbstractComponent):
@@ -48,7 +51,21 @@ class EDIStorageComponentMixin(AbstractComponent):
         :param filename: string for file name
         :return: PurePath object
         """
+        _logger.warning(
+            "Call of deprecated function `_remote_file_path`. "
+            "Please use `_get_remote_file_path` instead.",
+        )
         return self._dir_by_state(direction, state) / filename.strip("/ ")
+
+    def _get_remote_file_path(self, state, filename=None):
+        """Retrieve remote path for current exchange record."""
+        filename = filename or self.exchange_record.exchange_filename
+        direction = self.exchange_record.direction
+        directory = self._dir_by_state(direction, state).as_posix()
+        path = self.exchange_record.type_id._storage_fullpath(
+            directory=directory, filename=filename
+        )
+        return path
 
     def _get_remote_file(self, state, filename=None, binary=False):
         """Get file for current exchange_record in the given destination state.
@@ -57,8 +74,7 @@ class EDIStorageComponentMixin(AbstractComponent):
         :param filename: custom file name, exchange_record filename used by default
         :return: remote file content as string
         """
-        filename = filename or self.exchange_record.exchange_filename
-        path = self._remote_file_path(self.exchange_record.direction, state, filename)
+        path = self._get_remote_file_path(state, filename=filename)
         try:
             # TODO: support match via pattern (eg: filename-prefix-*)
             # otherwise is impossible to retrieve input files and acks
--- a/edi_storage_oca/models/edi_backend.py
+++ b/edi_storage_oca/models/edi_backend.py
@@ -145,16 +145,24 @@ class EDIBackend(models.Model):
         return record.identifier
 
     def _storage_get_input_filenames(self, exchange_type):
+        full_input_dir_pending = exchange_type._storage_fullpath(
+            self.input_dir_pending
+        ).as_posix()
         if not exchange_type.exchange_filename_pattern:
-            # If there is not patter, return everything
-            return self.storage_id.list_files(self.input_dir_pending)
+            # If there is not pattern, return everything
+            filenames = [
+                x
+                for x in self.storage_id.list_files(full_input_dir_pending)
+                if x.strip("/")
+            ]
+            return filenames
 
         bits = [exchange_type.exchange_filename_pattern]
         if exchange_type.exchange_file_ext:
             bits.append(r"\." + exchange_type.exchange_file_ext)
         pattern = "".join(bits)
-        full_paths = self.storage_id.find_files(pattern, self.input_dir_pending)
-        pending_path_len = len(self.input_dir_pending)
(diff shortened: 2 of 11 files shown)


edi_storage_oca/models/edi_backend.py
```diff
@@ -145,16 +145,24 @@ def _storage_create_record_if_missing(self, exchange_type, remote_file_name):
...
         if not exchange_type.exchange_filename_pattern:
-            # If there is not patter, return everything
-            return self.storage_id.list_files(self.input_dir_pending)
+            # If there is not pattern, return everything
+            files = self.storage_id.list_files(full_input_dir_pending)
+            filenames = []
+            for file in files:
+                if file.strip("/"):
+                    filenames.append(file)
+            return filenames
```
simahawk:
```suggestion
            filenames = [x for x in self.storage_id.list_files(full_input_dir_pending) if x.strip("/")]
            return filenames
```
Wondering: when do you get `/` only?

edi_storage_oca/models/edi_exchange_type.py
```diff
@@ -1,12 +1,18 @@
...
+
+from odoo import api, fields, models
+
+from odoo.addons.base_sparse_field.models.fields import Serialized
 
 
 class EDIExchangeType(models.Model):
     _inherit = "edi.exchange.type"
 
+    storage_settings = Serialized(default={}, compute="_compute_storage_settings")
```
simahawk: ```help="Alias to `storage` key in type settings"```

edi_storage_oca/models/edi_exchange_type.py
```diff
@@ -19,3 +25,44 @@ class EDIExchangeType(models.Model):
...
+            icp = self.env["ir.config_parameter"].sudo()
+            path = icp.get_param(path_config_param)
+            if path:
+                return PurePath(path)
+
+    def _get_full_exchange_type_path(self, directory=None):
+        self.ensure_one()
+        path_prefix = self._get_exchange_type_path()
+        path = PurePath((directory or "").strip().rstrip("/"))
+        if path_prefix:
```
simahawk, line 58: isn't this the same done in `_make_remote_file_path`?

edi_storage_oca/models/edi_exchange_type.py
```diff
@@ -19,3 +25,44 @@ class EDIExchangeType(models.Model):
...
+        if path:
+            return PurePath(path)
+        path_config_param = self.storage_settings.get("path_config_param")
+        if path_config_param:
+            icp = self.env["ir.config_parameter"].sudo()
+            path = icp.get_param(path_config_param)
+            if path:
+                return PurePath(path)
+
+    def _get_full_exchange_type_path(self, directory=None):
```
simahawk:
```suggestion
    def _storage_fullpath(self, directory=None):
```

edi_storage_oca/models/edi_exchange_type.py
```diff
@@ -19,3 +25,44 @@ class EDIExchangeType(models.Model):
         "the files to be fetched from the pending directory in the related "
         "storage. E.g: `.*my-type-[0-9]*.\\.csv`"
     )
+
+    @api.depends("advanced_settings")
+    def _compute_storage_settings(self):
+        for rec in self:
+            rec.storage_settings = rec.advanced_settings.get("storage", {})
+
+    def _get_exchange_type_path(self):
```
simahawk:
```suggestion
    def _storage_path(self):
```

simahawk (changes requested):

edi_storage_oca/models/edi_backend.py
jcoux: On this unit test:
https://github.com/OCA/edi/blob/15.0/edi_storage_oca/tests/test_edi_backend_storage.py#L206
The `list_files` function returns:
`['/', '/some-file.csv', '/another-file.csv']`

edi_storage_oca/models/edi_exchange_type.py
jcoux: Just removed `_make_remote_file_path` to avoid duplicates parts of code.

edi_storage_oca/components/base.py
```diff
@@ -40,15 +43,15 @@ def _dir_by_state(self, direction, state):
...
-        :param state: string stating state of the exchange
-        :param filename: string for file name
-        :return: PurePath object
-        """
-        return self._dir_by_state(direction, state) / filename.strip("/ ")
+    def _get_remote_file_path(self, state, filename=None):
+        """Retrieve remote path for current exchange record."""
+        filename = filename or self.exchange_record.exchange_filename
+        direction = self.exchange_record.direction
+        directory = self.backend[direction + "_dir_" + state] or ""
```
simahawk: pls keep using `self._dir_by_state` here which is what you are doing here but w/out validation

edi_storage_oca/tests/test_components_base.py
```diff
@@ -18,14 +18,18 @@ def test_remote_file_path(self):
...
 
-        with self.assertRaises(AssertionError):
-            self.checker._remote_file_path("WHATEVER", "error", "foo.csv")
+        with self.assertRaises(ValueError):
+            self.checker.exchange_record.direction = "WHATEVER"
+            self.checker._get_remote_file_path("error", "foo.csv")
 
-        with self.assertRaises(AssertionError):
-            self.checker._remote_file_path("input", "WHATEVER", "foo.csv")
+        with self.assertRaises(KeyError):
```
simahawk: key error?

edi_storage_oca/components/base.py
```diff
@@ -40,15 +43,15 @@ def _dir_by_state(self, direction, state):
...
+        path = self.exchange_record.type_id._storage_fullpath(
+            directory=directory, filename=filename
+        )
+        return path
```
simahawk, line 68: why not having always PurePath objects till the end as it was before?

edi_storage_oca/tests/common.py
```diff
@@ -76,13 +76,13 @@ def _file_fullpath(
         record = record or self.record
         if not fname:
             fname = self._filename(record, ack=ack)
-        if not direction:
-            direction = record.direction
+        if direction:
```
simahawk: why this change? Is not the same thing as before. Also, the direction on the record cannot be changed because it's a related from the type.

simahawk (changes requested):

edi_storage_oca/components/base.py
simahawk: we should probably keep the old method and log deprecation in case someone is still using it

jcoux:
@simahawk All is done: you can update your review ;)

damdam-s (approved):

SilvioC2C (approved):
LGTM overall

simahawk (approved):

PR #585, opened 2022-05-14
[14.0] [FIX][IMP] account_invoice_import buildout create new partner context

bosd:
The create new partner form, did not have all the fields populated with data from the parsed invoice. This fixes that.


bosd:
@alexis-via Can you re-open this one? 
Just checked, the email adress is still not populated on partner creation.

PR #589, opened 2022-05-18
[15.0][FIX] edi_storage_oca: Fix errors with get done/failed directories

jcoux:
When trying to get done/failed directoris, if we don't found it on SFTP
server, we ignore the error.

Current implementation only ignore `FileNotFoundError`, but depending of
the SFTP server, we can have `OSError` instead of.

This PR allow to also ignore `OSError`.

### DIFF
--- a/edi_storage_oca/components/base.py
+++ b/edi_storage_oca/components/base.py
@@ -1,10 +1,12 @@
 # Copyright 2020 ACSONE
 # @author: Simone Orsi <simahawk@gmail.com>
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
-
+import logging
 from pathlib import PurePath
 
 from odoo.addons.component.core import AbstractComponent
+
+_logger = logging.getLogger(__file__)
 
 
 class EDIStorageComponentMixin(AbstractComponent):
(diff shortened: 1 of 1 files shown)


edi_storage_oca/components/base.py
```diff
@@ -65,4 +67,14 @@ def _get_remote_file(self, state, filename=None, binary=False):
             # (the date will never match)
             return self.storage.get(path.as_posix(), binary=binary)
         except FileNotFoundError:
+            _logger.info(
+                f"Ignored FileNotFoundError when trying to "
+                f"get file {filename} into path {path} for state {state}"
```
simahawk: please use std log msg interpolation, eg: `_logger.info("Bla bla %s", foo)`

simahawk (changes requested):
Can you rewrite the commit msg please?
Eg: 
```
edi_storage_oca: improve error handling

When trying to get done/failed directories,
if we don't found it on SFTP server, we ignore the error.

Current implementation only ignores FileNotFoundError,
but depending on the SFTP server, we can have OSError.
```
In particular:
1. odoo version is not needed and confusing when back/forward porting
2. [FIX] is redundant w/ the msg
3. Is nice to have the explanation in the commit (which BTW is used automatically for the PR description when open it).

simahawk (approved):
LG. Would you mind backporting to v14?

jcoux:
> Would you mind backporting to v14?

https://github.com/OCA/edi/pull/590

yankinmax (approved):

PR #591, opened 2022-05-20
edi_oca: allow to override dt for exchange filename

jcoux:
Backported on https://github.com/OCA/edi/pull/592

By setting specific timezone and datetime format on exchange type
advanced settings, we can easily override the datetime used for generation of
exchange record filename.

### DIFF
--- a/edi_oca/models/edi_exchange_type.py
+++ b/edi_oca/models/edi_exchange_type.py
@@ -1,9 +1,14 @@
 # Copyright 2020 ACSONE SA
+# Copyright 2022 Camptocamp SA (https://www.camptocamp.com).
 # @author Simone Orsi <simahawk@gmail.com>
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
 import logging
+from datetime import datetime
+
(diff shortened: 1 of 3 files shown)


SilvioC2C (approved):

damdam-s (approved):

edi_oca/models/edi_exchange_type.py
```diff
@@ -134,12 +143,30 @@ def _check_backend(self):
             if rec.backend_id.backend_type_id != rec.backend_type_id:
                 raise exceptions.UserError(_("Backend should respect backend type!"))
 
+    def _make_exchange_filename_datetime(self):
+        """
+        Returns current datetime (now) using filename pattern
+        which can be set using advanced settings.
```
simahawk, line 149: Didn't notice this on 14.0 but not a blocker:

```suggestion
        """Returns current datetime (now) using filename pattern.
        
        The pattern can be configured using `filename_pattern` key in advanced settings.
```
Next time, don't change it now as it is the same on 14.0.

simahawk (approved):

PR #592, opened 2022-05-20
edi_oca: allow to override dt for exchange filename

jcoux:
Backport of https://github.com/OCA/edi/pull/591

By setting specific timezone and datetime format on exchange type
advanced settings, we can easily override the datetime used for generation of
exchange record filename.

### DIFF
--- a/edi_oca/models/edi_exchange_type.py
+++ b/edi_oca/models/edi_exchange_type.py
@@ -1,9 +1,14 @@
 # Copyright 2020 ACSONE SA
+# Copyright 2022 Camptocamp SA (https://www.camptocamp.com).
 # @author Simone Orsi <simahawk@gmail.com>
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
 import logging
+from datetime import datetime
+
+from pytz import timezone, utc
 
 from odoo import _, api, exceptions, fields, models
+from odoo.tools import DEFAULT_SERVER_DATETIME_FORMAT as DATETIME_FORMAT
 
 from odoo.addons.base_sparse_field.models.fields import Serialized
 from odoo.addons.http_routing.models.ir_http import slugify
@@ -87,6 +92,10 @@ class EDIExchangeType(models.Model):
                 process:
                   usage: $comp_usage
 
+              filename_pattern:
+                force_tz: Europe/Rome
+                date_pattern: %Y-%m-%d-%H-%M-%S
+
             In any case, you can use these settings
             to provide your own configuration for whatever need you might have.
         """,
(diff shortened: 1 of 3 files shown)


edi_oca/models/edi_exchange_type.py
```diff
@@ -134,12 +134,15 @@ def _check_backend(self):
             if rec.backend_id.backend_type_id != rec.backend_type_id:
                 raise exceptions.UserError(_("Backend should respect backend type!"))
 
+    def _get_dt_to_make_exchange_filename(self):
```
simahawk: what's the main goal in your case? Change the date pattern?

jcoux: In our case: force a specific timezone to always have it, without considering user timezone.

simahawk: Can we make this configurable via type settings?
I would support:
```
filename_pattern:
    force_tz: Europe/Rome
    date_pattern: y/m/d
```

simahawk:
```suggestion
    def _make_exchange_filename_datetime(self):
```

simahawk (changes requested):

jcoux:
@simahawk Fixed with this push force:
https://github.com/OCA/edi/compare/0bd3f4256ad1b70792aa07531012a84643fb21ca..1a7795d6c98301f25df22f3805401342651a5fe6

If it's now ok for you, just remaining to me to complete unit tests with new options.

edi_oca/models/edi_exchange_type.py
```diff
@@ -87,11 +91,18 @@ class EDIExchangeType(models.Model):
...
+              filename_pattern:
+                force_tz: Europe/Rome
+                date_pattern: %Y-%m-%d-%H-%M-%S
+
             In any case, you can use these settings
             to provide your own configuration for whatever need you might have.
         """,
     )
     advanced_settings = Serialized(default={}, compute="_compute_advanced_settings")
+    filename_pattern_settings = Serialized(
```
simahawk: I'm not in favour of adding a serialized field for each an every key in adv settings. This will encourage others to do so and IMO there's no gain especially for something that is used in a small ctx like this.

edi_oca/tests/test_backend_base.py
```diff
@@ -10,6 +10,7 @@
 class EDIBackendTestCase(EDIBackendCommonTestCase):
     @freeze_time("2020-10-21 10:00:00")
     def test_create_record(self):
+        self.env.user.tz = None  # Have no timezone used in generated filename
```
simahawk, line 13: pls add a test for the new behavior

simahawk (changes requested):

jcoux:
@simahawk All is done: you can update your review ;)

simahawk (approved):

PR #595, opened 2022-05-20
[14.0] sale_order_import: increase test coverage

simahawk:
The wizard is almost not tested at all and there are several methods that are not even used at all.

Depends on:

- [x] #596


simahawk:
For `get_xml_files_from_pdf` apparently it should come from bs.document.import inheritance as done here https://github.com/OCA/edi/blob/cce308bbbe625a44f406bc127a48654ea445f9a5/account_invoice_import/wizard/account_invoice_import.py#L23 but I think we should remove the necessity for such intheritance and use `bs.document.import` explicitly.

Also, in the wizard there's a kind of abandoned support for CSV import which should come from `sale_order_import_csv` but it's nowhere to be seen.

@alexis-via can you enlighten me pls? :pray:

simahawk:
As for CSV import, found: https://github.com/OCA/edi/blob/10.0/sale_order_import_csv/wizard/sale_order_import.py 
It's ported only to v10. I would take the occasion to support adding more supported formats but remove all the pseudo machinery for imports that are not available at all.

OTOH it's probably better to refactor everything in v15 and/or v16 and get rid of all these mastodon-tic models using components -> plug by default the EDI framework.

simahawk:
I've added 22% more test cov and cleaned up some duplicated code. ATM I'm not willing to spend more time on this because I see too many things to refactor as a lot of logic is duplicated here and there (and not tested) across modules and it would be probably better to start fresh w/ new modules w/ a modular design in mind.

SilvioC2C (approved):
LGTM

ivantodorovich (approved):

ivantodorovich:
needs rebasing

PR #596, opened 2022-05-23
[14.0] Add pdf_helper: stop copy/pasting code :)

simahawk:
While working on https://github.com/OCA/edi/pull/595 I've found that in sale_order_import the PDF support is broken and going down the rabbit hole I found we have duplicated pieces of code to handle PDFs which are not even tested.

This module aims to provide a tool that can be shared across different modules.
In other repos we have some copy/pasted methods for splitting PDF pages: we might want to move the here too.


pdf_helper/utils.py
```diff
@@ -0,0 +1,73 @@
...
+        """
+        res = {}
+        with BytesIO(self.pdf_file) as fd:
+            xmlfiles = self._extract_xml_files(fd)
+            for filename, xml_obj in xmlfiles.items():
+                root = self._extract_xml_root(xml_obj)
+                if not root:
+                    continue
+                res[filename] = root
+                _logger.debug(
```
simahawk: this logging must be removed

TDu (approved):

SilvioC2C (approved):
LGTM

ivantodorovich (approved):
LG

PR #573, opened 2022-04-27
[14.0][IMP][FIX] account_invoice_import_invoice2data add line support and additional partner fields

bosd:
[a121bb6d-91b3-4233-b89a-af9dd03a6cf1.webm](https://user-images.githubusercontent.com/11499387/231365074-3e690dee-9345-4798-94f9-ad7bd1ef6ffb.webm)

add support for:
Invoice lines

- [x] `discount` 
- [x] `price_subtotal` 
- [x] `date_start` + `date_end`
- [x] `uom_dict`
    - [x] `unece_code`
    - [x] `name` 
- [x] Product dict 
    - [x] `barcode`
    - [x] `code`
- [x] tax dict 
    - [x] `amount`
    - [x]  `vat_percent`
    - [x] `price_include` True of False (preparation for future Support for ATI All tax included) 
- [x] Invoice line notes // upstream support in account_invoice_import to be implemented
- [x] Invoice line sectionheader // upstream support in account_invoice_import to be implemented

partner fields:
- [x] Chamber of Commerce Number
- [x] bic
- [x] iban
- [x] note
- [x] phone
- [x] street
- [x] street2
- [x] city
- [x] zip
- [x] country_code

Global invoice fields:
- [x] note (will be imported in the chatter)
- [x] narration  (no support yet in account_invoice_import) Will be imported in bottom of invoice page

- [x] Improve documentation

todo:
- [x] add unit tests and test invoice

Leave other keys and values in tact so they could be processed by addons downstream.

related to issue: https://github.com/OCA/edi/issues/568

Keep compatability with template builder: https://github.com/OCA/edi/pull/399
Use standard field names: https://github.com/invoice-x/invoice2data/pull/313#issuecomment-1079958934

PR #597, opened 2022-05-25
[14.0] Add lines management to create multiline invoices

MTantin:
Multiline invoice creation is not supported yet.

I propose those modifications to accept multiline creation.

Invoice2data library accept format (with parser or not) to extract invoice line by line with regex.
the result is stored in "lines" field as json list.
I iterate over this list to prepare data for account_invoice_import submodule.

### DIFF
--- a/account_invoice_import_invoice2data/tests/test_invoice_import.py
+++ b/account_invoice_import_invoice2data/tests/test_invoice_import.py
@@ -3,10 +3,13 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 import base64
+import logging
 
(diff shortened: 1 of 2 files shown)


bosd:
Thanks for all your efforts.
There is already a PR for this. 
Please review: https://github.com/OCA/edi/pull/573

As it does not include tests, do you want to contribute your tests code in that pr.

bosd:
There is already a PR for this. 
Please review: https://github.com/OCA/edi/pull/573

MTantin:
Thx @bosd i didn't noticed this PR before, I'll look at this and contibute if i can.

PR #600, opened 2022-06-02
[IMP]edi_purchase_oca: Add write listener

GuillemCForgeFlow:
@ForgeFlow

### DIFF
--- a/edi_purchase_oca/models/purchase_order.py
+++ b/edi_purchase_oca/models/purchase_order.py
@@ -19,3 +19,8 @@ class PurchaseOrder(models.Model):
         if self:
             self._event("on_button_cancel_purchase_order").notify(self)
         return result
+
+    def write(self, vals):
+        if self:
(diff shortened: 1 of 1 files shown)


edi_purchase_oca/models/purchase_order.py
```diff
@@ -19,3 +19,8 @@ def button_cancel(self):
         if self:
             self._event("on_button_cancel_purchase_order").notify(self)
         return result
+
+    def write(self, vals):
+        if self:
+            self._event("on_button_edit_purchase_order").notify(self, vals)
```
LoisRForgeFlow: better `on_write_purchase_order` to keep the same structure as the other

GuillemCForgeFlow: Perfect! Thanks :)

simahawk:
Are you sure we need this change? The `on_write` event is available by default on all models.

PR #601, opened 2022-06-02
[13.0][FIX] remove PyPDF2 dependency

LoisRForgeFlow:
PyPDF2 dependency is already included in odoo CE requirements. See https://github.com/odoo/odoo/blob/13.0/requirements.txt#L31

This problem has been detected when this week PyPDF2 2.0.0 was released, since the version was not pinned in this repository, the latest 2.0.0 was installed, however this new version is not supported by odoo and gives errors

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,4 @@
 factur-x
 lxml
-PyPDF2
 pyyaml
 xmlschema
--- a/base_business_document_import/__manifest__.py
+++ b/base_business_document_import/__manifest__.py
@@ -17,5 +17,4 @@
         "account_tax_unece",
         "uom_unece",
     ],
-    "external_dependencies": {"python": ["PyPDF2"]},
 }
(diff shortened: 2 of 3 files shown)


AaronHForgeFlow (approved):

HviorForgeFlow (approved):

HviorForgeFlow:
@etobella

requirements.txt
```diff
@@ -1,5 +1,5 @@
 factur-x
 lxml
-PyPDF2
+PyPDF2==1.26.0
```
etobella: If the problem is 2.0.0 Why not doing <2.0.0 ?

LoisRForgeFlow: I also thought about it, but 1.26.0 [is in the odoo requirements](https://github.com/odoo/odoo/blob/13.0/requirements.txt#L31). Now that I think about it again... is this requirement actually needed in these modules? since it is a standard odoo requirement?

etobella: True!! It is probably unnecessary

etobella (approved):

simahawk:
IMO we should rollback this change and pin the right version. This sounds like a bug.

simahawk:
@LoisRForgeFlow can you revert this? See https://github.com/OCA/edi/pull/602

PR #602, opened 2022-06-03
[14.0][FIX] remove PyPDF2 dependency.

LoisRForgeFlow:
PyPDF2 dependency is already included in odoo CE requirements.

Forward port of https://github.com/OCA/edi/pull/601

### DIFF
--- a/base_ubl/__manifest__.py
+++ b/base_ubl/__manifest__.py
@@ -10,6 +10,6 @@
     "author": "Akretion,Onestein,Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/edi",
     "depends": ["uom_unece", "account_tax_unece", "base_vat", "pdf_helper"],
-    "external_dependencies": {"python": ["PyPDF2"]},
+    "external_dependencies": {"python": ["PyPDF2<=1.28.4"]},
     "installable": True,
 }
--- a/pdf_helper/__manifest__.py
+++ b/pdf_helper/__manifest__.py
@@ -15,5 +15,5 @@
     "depends": [
         "base",
     ],
-    "external_dependencies": {"python": ["PyPDF2"]},
+    "external_dependencies": {"python": ["PyPDF2<=1.28.4"]},
 }
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,7 +6,7 @@ invoice2data
 ovh
 pdfplumber
 phonenumbers
-PyPDF2
+PyPDF2<=1.28.4
 pyyaml
 regex
 requests


bosd (approved):
LGTM
Requurement is also satisfied in pdf_helper

GSLabIt (approved):

francesco-ooops:
@etobella @simahawk could this be merged? it's blocking our local repos

are tests broken due to this PR?

etobella:
@LoisRForgeFlow Can you test locally if you have the same error?

GSLabIt:
> @LoisRForgeFlow Can you test locally if you have the same error?

The error is raised since `PyPDF2 - 2.1.0`:

 ```
[531](https://github.com/.../runs/6779205686?check_suite_focus=true#step:5:532)
  File "/github/home/odoo-14.0/odoo-bin", line 5, in <module>
[532](https://github.com/.../runs/6779205686?check_suite_focus=true#step:5:533)
    import odoo
[533](https://github.com/.../runs/6779205686?check_suite_focus=true#step:5:534)
  File "/github/home/odoo-14.0/odoo/__init__.py", line 84, in <module>
[534](https://github.com/../runs/6779205686?check_suite_focus=true#step:5:535)
    PyPDF2.filters.decompress = _decompress
[535](https://github.com/.../runs/6779205686?check_suite_focus=true#step:5:536)
AttributeError: module 'PyPDF2' has no attribute 'filters'
[536](https://github.com/.../runs/6779205686?check_suite_focus=true#step:5:537)
```

**Review by alan196 — APPROVED** [review]
👍 Technical review

**Comment by francesco-ooops** [comment]
@LoisRForgeFlow could you take a look?

**Comment by francesco-ooops** [comment]
@LoisRForgeFlow could you take a look? :)

**Comment by GSLabIt** [comment]
@etobella any chance to merge it?

**Comment by etobella** [comment]
Tests are failing

**Comment by GSLabIt** [comment]
> Tests are failing

ok, it seems that need PyPDF version different from the one in odoo requirements, but it is breaking other modules that depends from this repo. 

Any chance to use PyPDF < 2.1.0?

**Comment by GSLabIt** [comment]
> Tests are failing

It seems other pr tests are failing as well, have a look at: https://github.com/OCA/edi/runs/6829375584?check_suite_focus=true#step:8:1120

**Comment by simahawk** [comment]
Why don't we pin the right version then?
IMO having python dependencies that are explicitly used in a module and not declared is a bug.

**Comment by LoisRForgeFlow** [comment]
> Why don't we pin the right version then? IMO having python dependencies that are explicitly used in a module and not declared is a bug.

It was the first fix I did, see https://github.com/OCA/edi/pull/601#pullrequestreview-993830447. However I found it is an standard Odoo dependency, so I don't see why it should be different here.

**Comment by LoisRForgeFlow** [comment]
Rebase done after #614 is merged.

**Comment by simahawk** [comment]
It must be here. This is a python pkg that is a direct requirement for the module. 
If you support only version xyz then pin this version here.
By removing it you are just hiding the dependencies and you are not documenting the fact that you support only version xyz.

**Review by simahawk — CHANGES_REQUESTED** [review]
see former msg

**Comment by GSLabIt** [comment]
> It must be here. This is a python pkg that is a direct requirement for the module.
> If you support only version xyz then pin this version here.
> By removing it you are just hiding the dependencies and you are not documenting the fact that you support only version xyz.

It seems issue has been raised since pyPDF moved to 2.1.0, pinning version < 2.1.0 should do the job.

**Comment by etobella** [comment]
Well, the right approach would be to pin the exact same version we have on odoo, and all should be fixed then

**Comment by LoisRForgeFlow** [comment]
> It must be here. This is a python pkg that is a direct requirement for the module.
If you support only version xyz then pin this version here.
By removing it you are just hiding the dependencies and you are not documenting the fact that you support only version xyz.

@simahawk  So, for instance if I have a module that depends on `xlrd` that it is an odoo standard requirement, it should be also listed in an OCA moudule using the library? Isn't this a bit redundant?

**Comment by LoisRForgeFlow** [comment]
> > It must be here. This is a python pkg that is a direct requirement for the module.
> > If you support only version xyz then pin this version here.
> > By removing it you are just hiding the dependencies and you are not documenting the fact that you support only version xyz.
> 
> It seems issue has been raised since pyPDF moved to 2.1.0, pinning version < 2.1.0 should do the job.

If we go this way, I would suggest to pin 1.26.0 to match [odoo requirements](https://github.com/odoo/odoo/blob/15.0/requirements.txt#L30).

**Comment by GSLabIt** [comment]
> So, for instance if I have a module that depends on `xlrd` that it an odoo standard requirement, it should be also listed in an OCA moudule using the library? Isn't this a bit redundant?

I think it needs to be added in case version needed is different from the odoo standard requirements, is this the case?

**Comment by LoisRForgeFlow** [comment]
> > So, for instance if I have a module that depends on `xlrd` that it an odoo standard requirement, it should be also listed in an OCA moudule using the library? Isn't this a bit redundant?
> 
> I think it needs to be added in case version needed is different from the odoo standard requirements, is this the case?

It is not the case, and if the version is different we would not be respecting Odoo requirements, which doesn't sound correct to me either.

**Comment by simahawk** [comment]
> > It must be here. This is a python pkg that is a direct requirement for the module.
> > If you support only version xyz then pin this version here.
> > By removing it you are just hiding the dependencies and you are not documenting the fact that you support only version xyz.
> 
> @simahawk So, for instance if I have a module that depends on `xlrd` that it is an odoo standard requirement, it should be also listed in an OCA moudule using the library? Isn't this a bit redundant?

it's not redundant, it's explicit.

Regarding which version to pin: be flexible and pin a range if possible.

**Comment by simahawk** [comment]
> requirements

You'll notice that another version is required and you'll be forced to test it and support it in these modules if needed.

**Comment by LoisRForgeFlow** [comment]
> > > It must be here. This is a python pkg that is a direct requirement for the module.
> > > If you support only version xyz then pin this version here.
> > > By removing it you are just hiding the dependencies and you are not documenting the fact that you support only version xyz.
> > 
> > 
> > @simahawk So, for instance if I have a module that depends on `xlrd` that it is an odoo standard requirement, it should be also listed in an OCA moudule using the library? Isn't this a bit redundant?
> 
> it's not redundant, it's explicit.
> 
> Regarding which version to pin: be flexible and pin a range if possible.

But why?, there a lot of modules in OCA that depend on standard python odoo dependencies and do not explicitly list them. It is like depending on "base" module in any module.

**Comment by simahawk** [comment]
> > > @simahawk So, for instance if I have a module that depends on `xlrd` that it is an odoo standard requirement, it should be also listed in an OCA moudule using the library? Isn't this a bit redundant?
> > 
> > 
> > it's not redundant, it's explicit.
> > Regarding which version to pin: be flexible and pin a range if possible.
> 
> But why?, there a lot of modules in OCA that depend on standard python odoo dependencies and do not explicitly list them. It is like depending on "base" module in any module.

Then they are wrong from my pov. If it is a python a **direct** dep it should be documented. With "direct" I mean that you type `from foo import bar` -> `foo` must be documented.

**Comment by LoisRForgeFlow** [comment]
@simahawk I don't really have an strong opinion, so let's follow your suggestion. Please, review again.

**Review by simahawk — APPROVED** [review]

**Comment by LoisRForgeFlow** [comment]
@simahawk All green, merge :rocket: ?

File: base_ubl/__manifest__.py [hunk 1]
```diff
@@ -10,6 +10,6 @@
     "author": "Akretion,Onestein,Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/edi",
     "depends": ["uom_unece", "account_tax_unece", "base_vat", "pdf_helper"],
-    "external_dependencies": {"python": ["PyPDF2"]},
+    "external_dependencies": {"python": ["PyPDF2<=1.28.4"]},
```
sbidoul, line 13:
```suggestion
    "external_dependencies": {"python": ["PyPDF2<2"]},
```

sbidoul: But actually, what happens if we simply leave "PyPDF2". Since it is already installed, pip should not upgrade it.

Could you try that so I can see the test log ?

sbidoul (commented):
But I don't understand why this change is necessary. In the log we see pip saying "requirement already installed", so it should not upgrade even if we don't set the upper bound.

GSLabIt:
> But I don't understand why this change is necessary. In the log we see pip saying "requirement already installed", so it should not upgrade even if we don't set the upper bound.

Our module depends from this repo and it raise an error, see:

https://github.com/OCA/edi/pull/602#issuecomment-1150009088

sbidoul:
> Our module depends from this repo and it raise an error, see:

@GSLabIt what module is that ? Can I see the install log ?

I mean, if base_ubl has no specific constraint on the PyPDF2 version it must **not** set an upper bound in its dependencies.

sbidoul:
(sorry missed a **not** word in the previous comment - updated)

GSLabIt:
> > Our module depends from this repo and it raise an error, see:
> 
> @GSLabIt what module is that ? Can I see the install log ?
> 
> I mean, if base_ubl has no specific constraint on the PyPDF2 version it must **not** set an upper bound in its dependencies.

It is a custom module for commission, it depends from sale-workflow repo, that depends from this repo. And since PyPDF 2.1.0 we're experiencing this issue. What happen is that, if not pinned, PyPDF will be updated to latest version, raising the issue.

sbidoul:
Ok, but the line that cause problem is in Odoo itself, not in base_ubl.
So there is nothing to change in base_ubl, and you should look at your install process why it did upgrade PyPDF2 and pin there if needed, but not here.

sbidoul (changes requested):
The issue does not seem to be related to base_ubl nor pdf_helper.

GSLabIt:
> Ok, but the line that cause problem is in Odoo itself, not in base_ubl.
> So there is nothing to change in base_ubl, and you should look at your install process why it did upgrade and pin there if needed, but not here.

I'm not getting ur point. We don't have any module with depends directly from this repo, neither other modules dependencies. 
One or more modules in sale-workflow depends on this repo; so while running tests, and runbots (using old oca modules version),  it install as first PyPDF from odoo requirements then install the one referenced in this repo requirements. 

So what i should do? it seems i can't do nothing, i've added PyPDF in my repo but still having the issue because version 2.1.0 will get installed.  

So i really don't know what to do; need to depend from sale-workflow but it comes out with this issue. 

if you have any hint it will be appreciated.

sbidoul:
@GSLabIt without knowning how you install I can't really help. 

If you do `pip install -r odoo-requirements.txt` it will install the PyPDF2 version that Odoo wants. Then if you do `pip install PyPDF2` pip should say `requirement already satisfied`.

Maybe you do it in the other order ? Or you `pip install --upgrade` somewhere ?

GSLabIt:
> @GSLabIt without knowning how you install I can't really help.
> 
> If you do `pip install -r odoo-requirements.txt` it will install the PyPDF2 version that Odoo wants. Then if you do `pip install PyPDF2` pip should say `requirement already satisfied`.
> 
> Maybe you do it in the other order ? Or you `pip install --upgrade` somewhere ?

Order is odoo first than other repos. and nope, double checked and we do not have `pip install --upgrade`

GSLabIt:
> @GSLabIt without knowning how you install I can't really help.
> 
> If you do `pip install -r odoo-requirements.txt` it will install the PyPDF2 version that Odoo wants. Then if you do `pip install PyPDF2` pip should say `requirement already satisfied`.
> 
> Maybe you do it in the other order ? Or you `pip install --upgrade` somewhere ?

At the end of the day, what's the problem to add `PyPDF2<2` in requirements on this repo?

sbidoul:
> At the end of the day, what's the problem to add PyPDF2<2 in requirements on this repo?

Because pinning / bounding requirements when not necessary cause unresolvable version conflicts down the line.
And as you can see in the installation log of this repo, pip says `requirement already installed` so the current config is not an issue.

We really need to understand what is going wrong in your installation process. Searching for pypdf2 in your installaiton logs should help.

GSLabIt:
> > At the end of the day, what's the problem to add PyPDF2<2 in requirements on this repo?
> 
> Because pinning / bounding requirements when not necessary cause unresolvable version conflicts down the line. And as you can see in the installation log of this repo, pip says `requirement already installed` so the current config is not an issue.
> 
> We really need to understand what is going wrong in your installation process. Searching for pypdf2 in your installaiton logs should help.

I finally found the issue, it is on clone_oca_dependencies:

https://github.com/OCA/maintainer-quality-tools/blob/d705b9d68abd2616ad786bd29f760ee0ca182692/travis/clone_oca_dependencies#L119

michelerusti:
I'm experiencing a similar issue on local repos, caused by the same non-pinned version. I too would like his to be merged

PR #604, opened 2022-06-04
[14.0][IMP] business_document_import Usability improvement for multiline product imports

bosd:
Be less strict on product matching.
If one line fails to match a product, the whole (invoice) import fails.
This is unwanted in a professional environment when importing files with a lot of lines is very common.

Also it was too strict as it did not support the import of lines without any products.
Example, line notes and section headers.

This PR changes the behavior.
Instead of throwing an user error and stopping the import, an chatter message is displayed that not all products could automatically be matched.
This is in line with how uom imports are being handled.
The user can manually adjust the lines with the feedback he received from the chatter.

### DIFF
--- a/base_business_document_import/tests/test_business_document_import.py
+++ b/base_business_document_import/tests/test_business_document_import.py
@@ -164,12 +164,10 @@ class TestBaseBusinessDocumentImport(TransactionCase):
         )
         self.assertEqual(res, product1)
         raise_test = True
-        try:
-            bdio._match_product(product_dict, [], seller=False)
-            raise_test = False
-        except Exception:
-            pass
         self.assertTrue(raise_test)
+        chatter = []
+        bdio._match_product(product_dict, chatter_msg=chatter, seller=False)
+        self.assertGreaterEqual(len(chatter), 1)
 
     def test_match_uom(self):
         bdio = self.env["business.document.import"]
--- a/base_business_document_import/models/business_document_import.py
+++ b/base_business_document_import/models/business_document_import.py
@@ -600,9 +600,7 @@ class BusinessDocumentImport(models.AbstractModel):
                 and len(sinfo.product_tmpl_id.product_variant_ids) == 1
             ):
                 return sinfo.product_tmpl_id.product_variant_ids[0]
-        raise self.user_error_wrap(
-            "_match_product",
-            product_dict,
+        chatter_msg.append(
             _(
                 "Odoo couldn't find any product corresponding to the "
                 "following information extracted from the business document:\n"
@@ -616,6 +614,7 @@ class BusinessDocumentImport(models.AbstractModel):
                 seller and seller.name or "",
             ),
         )
+        return None
 
     @api.model
     def _match_product_search(self, product_dict):


bosd:
@i-vyshnevska can you please review?

bosd:
@simahawk Can you re-trigger tests? 🙏

simahawk:
tests are broken

This one I'll check myself: 

https://github.com/OCA/edi/runs/6740399998?check_suite_focus=true#step:8:235


this is probably yours https://github.com/OCA/edi/runs/6740399998?check_suite_focus=true#step:8:1348

simahawk:
you have to rebase. The 1st issue is fixed already here https://github.com/OCA/edi/commit/abaccd983dc9dff71f1c2ae6f49ade2522e8b159

bosd:
Rebased. tests are green now :)

base_business_document_import/tests/test_business_document_import.py
```diff
@@ -164,11 +164,6 @@ def test_match_product(self):
         )
         self.assertEqual(res, product1)
         raise_test = True
-        try:
```
simahawk, line 167: This change makes little sense to me. You are not testing anything anymore.
Moreover, for testing exceptions we should use `assertRaises` and not catch them.

You should test that you have the right msg instead.

simahawk (changes requested):
I'm not sure that changing the behavior for all cases by default is safe in this version.
You should probably do this w/ a ctx key or explicit param.

Additionally, please keep commits short w/ 1 line summary and extended description. Eg:

```
base_business_document_import: improve match_product

Make product match fault tolerant in case of missing match.

If one line fails to match a product, the whole (invoice) import fails.
This is unwanted in a professional environment when importing files with a lot of lines is very common.

Also it was too strict as it did not support the import of lines without any products.
Example, line notes and section headers.

This PR changes the behavior.
Instead of throwing an user error and stopping the import, an chatter message is displayed that not all products could automatically be matched.
This is in line with how uom imports are being handled.
The user can manually adjust the lines with the feedback he received from the chatter.
```

base_business_document_import/tests/test_business_document_import.py
bosd: Thanks for your help, writing test is still a bit complicated for me :)
Let's hope this will do the trick.

bosd: @simahawk Can you re-review?

bosd:
@simahawk  Can you re-review?
- [X]  tests are updated
- [X] commit message changed

Tests are failling, but does not seem to be mine..

simahawk:
@bosd sorry but my comment https://github.com/OCA/edi/pull/604#pullrequestreview-1064735946 has not been addressed. 
Nor for the conditional change, nor for the commit msg (which I proposed above in the right form).

bosd:
> Nor for the conditional change, nor for the commit msg (which I proposed above in the right form).

@simahawk  Sorry, had a different interpretation of your comment for the commit message.
I changed it into
''Make product match fault tolerant in case of missing match''

>I'm not sure that changing the behavior for all cases by default is safe in this version.
You should probably do this w/ a ctx key or explicit param.

Sorry, I don't understand what would be unsafe in this current form. Or what you mean with a ctx key or explicit param.
I don't see the danger.
Don't get me wrong, I am willing to implement it the way you like. But I do need some more explanation/guidance to do so.

bosd:
Switched to a linux system. Now was able to edit the commit message.
@simahawk Is it oke now?

bofiltd (approved):
Looks Good To me! I prefer to have it more user friendly. Only chatter message on matching error is ok.

bosd:
> Nor for the conditional change

@simahawk Can you elaborate / make a code suggestion how to achieve this?

rvalyi:
@bosd as for not failing completely when one invoice line importation fails, you could consider depending (or in an extension) on https://github.com/OCA/account-invoicing/blob/14.0/account_move_exception which is pretty much designed for this kind of automation use case.

bosd:
@rvalyi Thanks for pointing to that module. However, I don't see how that could be tied in into the `business_document_import`.
It looks more like a module which is there tho handle after the fact (exception) has occured.

PR #606, opened 2022-06-05
[14.0][FIX] sale_order_import_ubl: super().parse_xml_order() with all arguments

ghost:
`parse_xml_order(xml_root)` needs the argument `detect_doc_type` to avoid errors.

### DIFF
--- a/sale_order_import_ubl/wizard/sale_order_import.py
+++ b/sale_order_import_ubl/wizard/sale_order_import.py
@@ -26,7 +26,9 @@ class SaleOrderImport(models.TransientModel):
             else:
                 return self.parse_ubl_sale_order(xml_root)
         else:
-            return super(SaleOrderImport, self).parse_xml_order(xml_root)
+            return super(SaleOrderImport, self).parse_xml_order(
+                xml_root, detect_doc_type
+            )
 
     @api.model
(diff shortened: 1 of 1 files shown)


ghost:
@alexis-via Here is a bugfix that you might want to merge.

I am not using this module in production, but I used it as a template for a new account statement import module.

sale_order_import_ubl/wizard/sale_order_import.py
```diff
@@ -26,7 +26,9 @@ def parse_xml_order(self, xml_root, detect_doc_type=False):
             else:
                 return self.parse_ubl_sale_order(xml_root)
         else:
-            return super(SaleOrderImport, self).parse_xml_order(xml_root)
+            return super(SaleOrderImport, self).parse_xml_order(
+                xml_root, detect_doc_type
```
simahawk, line 30: Well, this is not broken but it will have a strange behavior if you have installed several modules handling SO import based on the order their override is executed.

Also, in v14 I don't find any other module but this one, checking for `detect_doc_type`. I guess you overridden it as well?

Which error in particular have you faced?

simahawk:
@appstogrow please check my comment above. Also, a rebase + force push is needed.

PR #620, opened 2022-07-01
Update of the translations of ``base_facturx`` and ``account_invoice_facturx``.

eferriere:
This is an update of the translations of ``base_facturx`` and ``account_invoice_facturx``.


pedrobaeza:
This is not correct picked and it's more than translations. Besides, translations are done on https://translation.odoo-community.org/

Issue #610, opened 2022-06-10
account_invoice_import : improving handling of simple PDF invoices 


alexis-via:
This issue is about PDF supplier invoices that don't have any XML attachment. For those invoices, we have 2 modules:
- account_invoice_import_invoice2data
- account_invoice_import_simple_pdf

So far, I have always considered that account_invoice_import_simple_pdf would replace account_invoice_import_invoice2data. But, following discussions with some users, I'm starting to change my mind. Each module has it's own advantage:
- account_invoice_import_invoice2data: ability to share invoice templates among different Odoo servers by sharing the yaml files
- account_invoice_import_simple_pdf: ability to add support for a new vendor without technical skills

So it could be interesting to be able to use the 2 modules on the same Odoo server:
1) if the PDF invoice can be read by an invoice2data template, use it
2) otherwise, try to find a simple_pdf configuration

Technically, it may be as simple as removing the following raise in account_invoice_import_invoice2data: https://github.com/OCA/edi/blob/14.0/account_invoice_import_invoice2data/wizard/account_invoice_import.py#L56

Another related topic: If there is no invoice2data template nor simple_pdf configuration, we raise an error, cf https://github.com/OCA/edi/blob/14.0/account_invoice_import/wizard/account_invoice_import.py#L97
Starting from Odoo v13, it is possible to create an "empty" draft invoice with the PDF as attachment without partner and without any line. This is what Odoo EE does when it is unable to recognize the invoice.

Advantages:
- if you have setup a mail gateway to create vendor bills, you get all the PDF invoices that you receive as draft invoice (currently, the PDF invoices which are not "recognized" are skipped and the accountant has no way to see it in Odoo, cf https://github.com/OCA/edi/blob/14.0/account_invoice_import/wizard/account_invoice_import.py#L1420)
- if you are importing the PDF invoice manually via the Invoice import wizard, you avoid the error message of the wizard and you save a few seconds creating the vendor bill manually and adding the PDF invoice as attachment (of course, you still have to select the partner, set the date and create the invoice line manually).

Drawback:
- If you use the mail gateway, any spam that has a PDF attachment will create a draft vendor bill !

It would be interesting to have the feedback of people who have used Odoo EE to know their opinion about these draft empty invoices.

avmediadesign:
Being able to handle all invoices (via e-mail gateway or manual import) regardless if there is or not an invoice2data template or simple_pdf configuration would be an valuable enhancement.
[...]

PR #622, opened 2022-07-04
[ADD] New module account_invoice_import_main_attachment

NL66278:
account_invoice_import already can work on emails fetched from email, but as far as I can see discards all emails it does not recognize an import definition for.

With this module you can create (draft) vendor invoices in the standard Odoo way, using a mail alias linked to a journal. Then look at the attachment linked to the invoice, create an import definition based on what you see, and then import the invoice using the definition.

Of course subsequent invoices for the same supplier can just be imported.


account_invoice_import_main_attachment/wizards/account_invoice_import.py
```diff
@@ -0,0 +1,57 @@
...
+            )
+        return vals
+
+    def update_invoice(self):
+        """Can override partner in invoice, if not set already."""
+        self.ensure_one()
+        invoice = self.invoice_id
+        parsed_inv = self.get_parsed_invoice()
+        partner = self._match_partner(
+            parsed_inv["partner"], parsed_inv["chatter_msg"], partner_type="supplier"
```
TeoGoddet, line 53: I think the invoice partner is automatically set by odoo account module when receiving invoice from email

TeoGoddet (commented):
This is great, i'm just wondering about the usefulness of of setting the partner from the chatter as odoo natively do it in v14

account_invoice_import_main_attachment/wizards/account_invoice_import.py
NL66278: I have taken this from the base code. But had to set the partner before calling super(), otherwise the super() method would complain about the partner being different. It might be argued that it would be better to change the original code, because why not set the partner if we find the partner in the imported document and it has not been set on the invoice yet?

alexis-via:
@NL66278 You may not have followed this discussion:
https://github.com/OCA/edi/issues/610
The second part of the issue deals about the reception of invoices by email and the creation of "empty" invoices". It was implemented in PR https://github.com/OCA/edi/pull/630 which was merged a few days ago.
Does this PR still make sense in this context ?

PR #630, opened 2022-07-14
[14.0] account_invoice_import: allow to create empty invoices when nothing is extracted from the PDF

alexis-via:
Implement what is described in issue #610 :
- account_invoice_import_simple_pdf and account_invoice_import_invoice2data can now be use
d together
- when importing an invoice for an unknown partner, it creates an empty invoice with PDF a
s attachment
- when importing an invoice via mail gateway and the from email is set
on a partner that doesn't have any import config, create a supplier
invoice with PDF as attachment and with that partner
- when importing an invoice via mail gateway and the from email is set
on a partner that has a single-line import config, create a supplier
invoice with PDF as attachment, that partner and 1 invoice line with price_unit=0.0


alexis-via:
Travis is red. It is caused by the fact that my Bouygues Telecom test invoice that I shipped in account_invoice_import_simple_pdf also matches a template of invoice2data ! Now that the 2 modules are compatible, I should ship an invoice that will not match an invoice2data template.

bosd:
Can you add the feature to toggle the "To Check" value if an empty invoice is created?
(maybe also force auto_post to false)

alexis-via:
Cool, tests are green now!

alexis-via:
@bosd: ok, I'll have a look at it, but not today.

alexis-via:
@bosd When the invoice import doesn't match any partner, and you try to confirm it, you get the message "You need to add a line before posting."
When the invoice import matches a partner that has a single line import config and you try to confirm it without modification, you get the message "The Bill/Refund date is required to validate this document.".
So, anyway, you have to update the vendor bill before being able to confirm it. So I don't see any advantage of checking the "To check" boolean ; if must update/verify the invoice anyway to be able to confirm it. And once you have updated the invoice, you may not want it to have the "To check" option.
Setting auto_post to False doesn't have any advantage either, because the empty invoice is created without date, so it will never be auto-posted anyway.

bosd:
That makes sense. So it is not nedeed here.


However it might be useful to add support of the to_check parameter for businesses_document_import.

In case when product match fails.
https://github.com/OCA/edi/pull/604

Or when a uom, tax is not correctly matched.

nibraun (approved):
Used in production, no code review

njeudy (approved):
I use this in production, very basic code review

PR #640, opened 2022-08-09
[14.0] edi_oca: improve auto btns

simahawk:
* auto btns on/off flag per exchange type
* offer hook to customize how btns are rendered
* improve UI
* rename config field to be more generic, rename form field to be more specific


edi_oca/models/edi_exchange_type.py
```diff
@@ -115,6 +115,11 @@ class EDIExchangeType(models.Model):
         It will be executed if variable result has been defined as True
         """,
     )
+    model_manual_btn = fields.Boolean(
+        string="Manual button on form",
+        help="Automatically display a btn on related models' form."
```
sebalix:
```suggestion
        help="Automatically display a button on related models' form."
```

sebalix (approved):
LG

simahawk:
@etobella can you have a look here?

simahawk:
I still have have few things to rework.

edi_oca/migrations/14.0.1.10.0/post-migrate.py
```diff
@@ -0,0 +1,19 @@
...
+from odoo import SUPERUSER_ID, api
+
+_logger = logging.getLogger(__name__)
+
+
+def migrate(cr, version):
+    if not version:
+        return
+
+    env = api.Environment(cr, SUPERUSER_ID, {})
```
etobella, line 15: Why not using openupgradelib?

etobella (approved):
Small comment. Seems right for me on a basic review.

edi_oca/migrations/14.0.1.10.0/post-migrate.py
simahawk: because AFAIS there's no need. What's the benefit in this case?

simahawk:
@etobella please have a look at my last commit https://github.com/OCA/edi/pull/640/commits/0acaa54cfedb182fd572aaba77795961138669e2 -> it renames fields to be more explicit on what they are used for.
Being all computed non stored fields I don't see the need for backward compat but: please tell me if you ended up overriding/customizing those fields and view for any reason (or you know someone did).

edi_oca/models/edi_exchange_consumer_mixin.py
```diff
@@ -24,21 +25,19 @@ class EDIExchangeConsumerMixin(models.AbstractModel):
         domain=lambda r: [("model", "=", r._name)],
     )
     exchange_record_count = fields.Integer(compute="_compute_exchange_record_count")
-    expected_edi_configuration = Serialized(
-        compute="_compute_expected_edi_configuration",
+    edi_form_config = Serialized(
```
simahawk: this can be probably made even more generic and named `edi_config` while keeping `edi_has_form_config` which is specific.

simahawk: Ok, I changed this to be more generic so that we can reuse it for any kind of lookup of configuration on consumer models.

edi_oca/migrations/14.0.1.10.0/post-migrate.py
etobella: Well, usually it is good in order to avoid the env definition, obviously not needed, but makes it all easier IMO. Obviously this is not required

etobella:
LGTM 

Some tests might fail on l10n-spain as soon as it is migrated. I can check it as soon as it has been changed on 14.0 and 15.0. Changes are not used as far as I can see :smile:

simahawk:
> LGTM
> 
> Some tests might fail on l10n-spain as soon as it is migrated. I can check it as soon as it has been changed on 14.0 and 15.0. Changes are not used as far as I can see smile

Which tests? Can you point them here?
However, if it's only on v13 then they'll be fixed when ported, yes.

victoralmau:
Please can you apply these changes also to v15?

simahawk:
> Please can you apply these changes also to v15?

I can find some time. I won't be able to test manually tho. Can you do that?

I can also port this https://github.com/OCA/edi/pull/645 -> you can have a look before I merge.

PR #642, opened 2022-08-12
[FIX][14.0]purchase_order_ubl - fix stack-trace in _render_qweb_pdf

luc-demeyer:
You get a stack trace when validating a bank statement on a system with the purchase_order_ubl module installed.

This caused by the fact that the purchase_order_ubl/models/report.py expects res_ids to a list whereas it can also be an integer.
cf. https://github.com/odoo/odoo/blob/10378872eddd6037f479fa8cbb3ed65d5e4b52c6/addons/account/models/account_bank_statement.py#L440

The problem is fixed by adding a check on the res_ids type.

### DIFF
--- a/purchase_order_ubl/models/report.py
+++ b/purchase_order_ubl/models/report.py
@@ -31,7 +31,14 @@ class IrActionsReport(models.Model):
         """
         This is only necessary when tests are enabled.
         It forces the creation of pdf instead of html."""
-        if len(res_ids or []) == 1 and not self.env.context.get("no_embedded_ubl_xml"):
+        # Even if not expected, we might get an ID.
+        # Eg https://github.com/odoo/odoo/blob/10378872eddd6037f479fa8cbb3ed65d5e4b52c6
+        # /addons/account/models/account_bank_statement.py#L440
+        if (
+            isinstance(res_ids, int)
+            or len(res_ids or []) == 1
+            and not self.env.context.get("no_embedded_ubl_xml")
+        ):
             if len(self) == 1 and self.is_ubl_xml_to_embed_in_purchase_order():
                 self = self.with_context(force_report_rendering=True)
         return super()._render_qweb_pdf(res_ids, data)


simahawk:
Hmm to me it seems that the core method always wants a list of ids:

https://github.com/odoo/odoo/blob/14.0/odoo/addons/base/models/ir_actions_report.py#L892
https://github.com/odoo/odoo/blob/14.0/odoo/addons/base/models/ir_actions_report.py#L603

vanderperre (approved):

BenjaHe (approved):
This is working for me

luc-demeyer:
@simahawk 
I checked the _post_pdf method in standard addons via the links you posted here and you are right: in _post_pdf it should always be a list of ids hence I limited the fix to the _render_qweb_pdf where both id or ids are allowed (and the stack trace I experienced happened only in this method).
Hence I think we are ok now for merge.

simahawk:
IMO the bug to fix is in the core on the line you linked: they shouldn't pass `statement.id` but `statement.ids`.

https://github.com/odoo/odoo/blob/10378872eddd6037f479fa8cbb3ed65d5e4b52c6/addons/account/models/account_bank_statement.py#L440

Now, we could keep the fix here but I would add a comment on the reason why we need and I would open an issue on odoo/odoo.

luc-demeyer:
@simahawk 
the statement.id in the core is correct since the _render_qweb_pdf method transforms the integer to a list, hence we should fix it on our end.
cf. https://github.com/odoo/odoo/blob/2256d8075d8e6bf2b09461aa947a1c8ee2bdcbf3/odoo/addons/base/models/ir_actions_report.py#L807

simahawk:
I still think that it's wrong to pass an ID. The real "magic" happens here https://github.com/odoo/odoo/blob/2256d8075d8e6bf2b09461aa947a1c8ee2bdcbf3/odoo/addons/base/models/ir_actions_report.py#L794 where `browse` can handle an ID or a list. But everywhere in the code they assume they get a list of IDS, not a single integer.

Anyway, if you want to fix it here fine, but please add a reference in the code eg:

```
# Even if not expected, we might get an ID.
# Eg: https://github.com/odoo/odoo/blob/10378872eddd6037f479fa8cbb3ed65d5e4b52c6
# /addons/account/models/account_bank_statement.py#L440
```
Last but not least: please rewrite the commit msg as 

```
purchase_order_ubl: fix pdf render w/ single ID

In some places odoo is passing a single ID instead of a list as expected.
```
NOTE: the odoo version `[14.0]` should not be included because is not needed and not useful when you migrate changes from version to version (and seems we need this for 15.0).

Thanks!

luc-demeyer:
@simahawk 
comments added to the code + commit message adapted.

simahawk:
@luc-demeyer thanks for your patience :)
/ocabot merge patch

PR #623, opened 2022-07-08
[IMP] edi: archive edi_backend

JaumeBforgeFlow:
Adding the feature to archive edi.backend so they can be ignored on searches.

PR #643, opened 2022-08-12
[14.0] edi: backend archive (port from v13) + integrate w/ edi_endpoint

simahawk:
https://github.com/OCA/edi/pull/623

### DIFF
--- a/edi_endpoint_oca/models/edi_backend.py
+++ b/edi_endpoint_oca/models/edi_backend.py
@@ -2,7 +2,7 @@
 # @author: Simone Orsi <simone.orsi@camptocamp.com>
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
 
-from odoo import api, fields, models
+from odoo import _, api, exceptions, fields, models
 
 
 class EDIBackend(models.Model):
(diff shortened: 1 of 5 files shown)


edi_endpoint_oca/models/edi_backend.py
```diff
@@ -16,10 +16,16 @@ class EDIBackend(models.Model):
     )
     endpoints_count = fields.Integer(compute="_compute_endpoints_count")
 
-    @api.depends("endpoint_ids")
+    @api.depends("endpoint_ids.active")
     def _compute_endpoints_count(self):
+        data = self.env["edi.endpoint"].read_group(
+            [("backend_id", "in", self.ids), ("active", "=", True)],
```
SilvioC2C, line 22: `("active", "=", True)` shouldn't be necessary

SilvioC2C (approved):
LGTM, one minor comment

edi_endpoint_oca/models/edi_backend.py
simahawk: It's just more explicit.

TDu (approved):

PR #647, opened 2022-08-27
[14.0] edi_oca: misc fix/imp

simahawk:

### DIFF
--- a/edi_exchange_template_oca/models/edi_exchange_template_mixin.py
+++ b/edi_exchange_template_oca/models/edi_exchange_template_mixin.py
@@ -13,6 +13,19 @@ from odoo.tools import DotDict, safe_eval
 _logger = logging.getLogger(__name__)
 
 
+def date_to_datetime(dt):
+    """Convert date to datetime."""
+    if isinstance(dt, datetime.date):
+        return datetime.datetime.combine(dt, datetime.datetime.min.time())
+    return dt
+
+
+def to_utc(dt):
+    """Convert date or datetime to UTC."""
+    # Gracefully convert to datetime if needed 1st
+    return date_to_datetime(dt).astimezone(pytz.UTC)
+
+
 class EDIExchangeTemplateMixin(models.AbstractModel):
     """Define a common ground for EDI exchange templates."""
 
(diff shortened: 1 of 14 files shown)


edi_oca/models/edi_backend.py
```diff
@@ -341,7 +341,7 @@ def _cron_check_output_exchange_sync(self, **kw):
             backend._check_output_exchange_sync(**kw)
 
     # TODO: consider splitting cron in 2 (1 for receiving, 1 for processing)
-    def _check_output_exchange_sync(self, skip_send=False):
+    def _check_output_exchange_sync(self, skip_send=False, skip_sent=True):
```
TDu, line 344: The docstring could be updated for the new parameter

edi_exchange_template_oca/models/edi_exchange_template_mixin.py
```diff
@@ -13,6 +13,19 @@
 _logger = logging.getLogger(__name__)
 
 
+def date_to_datetime(dt):
+    """Convert date to datetime."""
+    if isinstance(dt, datetime.date):
+        return datetime.datetime.combine(dt, datetime.datetime.min.time())
+    return dt
```
SilvioC2C, line 20: Could it make sense to add timezone info?
Also, `datetime.datetime.min.time()` can be replaced by importing `time` from `datetime` and then using `time.min`.

edi_exchange_template_oca/models/edi_exchange_template_mixin.py
```diff
@@ -77,22 +90,42 @@ def _utc_now():
...
     def _date_to_string(dt, utc=True):
+        if not dt:
+            return ""
         if utc:
-            dt = dt.astimezone(pytz.UTC)
+            dt = to_utc(dt)
         return fields.Date.to_string(dt)
 
+    @staticmethod
+    def _datetime_to_string(dt, utc=True):
```
SilvioC2C, line 100: `_date_to_string` and `_datetime_to_string` are the same method, maybe we can avoid defining one of them and simply execute the other one when called. Something like this:
```
    @staticmethod
    def _date_to_string(dt, utc=True):
        if not dt:
            return ""
        if utc:
            dt = to_utc(dt)
        return fields.Date.to_string(dt)

    _datetime_to_string = _date_to_string
```

edi_oca/models/edi_exchange_consumer_mixin.py
```diff
@@ -232,3 +238,10 @@ def get_edi_access(self, doc_ids, operation, model_name=False):
         else:
             check_operation = operation
         return check_operation
+
+    def _edi_set_origin(self, exc_record):
+        self.sudo().update({"origin_exchange_record_id": exc_record.id})
+
+    def _edi_get_origin(self):
+        self.ensure_one()
+        return self.origin_exchange_record_id
```
SilvioC2C, line 247: Any specific reason for having setter/getter methods for this field?

SilvioC2C (approved):
Overall LGTM, just some minor comments

edi_oca/models/edi_exchange_consumer_mixin.py
simahawk: they are not setter/getter for a field, they are api methods. And no that you let me think about them, they should not be private.

edi_exchange_template_oca/models/edi_exchange_template_mixin.py
simahawk: they are similar, not the same. One uses `fields.Date` and the other `fields.Datetime` :wink:

edi_exchange_template_oca/models/edi_exchange_template_mixin.py
simahawk: TZ I don't think is needed here. I don't see the point of replacing the min time var. What's the pro?

PR #651, opened 2022-09-02
[14.0] edi: misc improvements

simahawk:

### DIFF
--- a/edi_oca/models/edi_backend.py
+++ b/edi_oca/models/edi_backend.py
@@ -341,7 +341,9 @@ class EDIBackend(models.Model):
             backend._check_output_exchange_sync(**kw)
 
     # TODO: consider splitting cron in 2 (1 for receiving, 1 for processing)
-    def _check_output_exchange_sync(self, skip_send=False, skip_sent=True):
+    def _check_output_exchange_sync(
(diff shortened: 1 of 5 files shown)


edi_oca/models/edi_backend.py
```diff
@@ -379,17 +383,20 @@ def _check_output_exchange_sync(self, skip_send=False, skip_sent=True):
...
             ("type_id.direction", "=", "output"),
             ("edi_exchange_state", "=", "new"),
             ("exchange_file", "=", False),
         ]
+        if record_ids:
+            domain.append(("id", "in", record_ids))
+        return domain
 
-    def _output_pending_records_domain(self, skip_sent=True):
+    def _output_pending_records_domain(self, skip_sent=True, record_ids=None):
```
etobella, line 399: Why do yo do this?

I mean, isn't it used on the cron only? I assume you need it for something else, isn't it?

etobella (approved):
In general LGTM. Just one question

edi_oca/models/edi_backend.py
etobella: Understood after checking #653

PR #653, opened 2022-09-05
[14.0] edi: add quick exec option by type

simahawk:
Allow scheduling jobs immediately based on type configuration.
Very often you want an action to be executed right away.
You still get a job, meaning that the action is not sync
but the job will be scheduled immediately
instead of waiting for the cron to run.

Plus fix `retryable` flag. See atomic commits for details.


etobella (approved):
LGTM

PR #657, opened 2022-09-09
[13.0][FIX] edi_backend_partner

AnnaPForgeFlow:
correct partner_id field location


LoisRForgeFlow:
@AnnaPForgeFlow Could you extend a bit the commit message describing the fix?

PR #662, opened 2022-09-13
[14.0] Add edi_party_data_oca

simahawk:
Technical module for the EDI suite module to retrieve data for parties of an exchange.

This module provides default component
and a mixin to be used for registering new components for specific backends.

It's based on `partner_identification`
so that the party information will include allowed ID numbers for a given exchange.

You can configure which ID number categories are allowed on the exchange type.

ref: cos-3752

### DIFF
--- a/edi_party_data_oca/README.rst
+++ b/edi_party_data_oca/README.rst
@@ -1,0 +1,1 @@
+wait for the bot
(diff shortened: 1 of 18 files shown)


simahawk:
Fix moved to https://github.com/OCA/edi/pull/664 and getting merged

TDu (approved):

PR #664, opened 2022-09-13
[14.0] edi_webservice: fix tests

simahawk:
'webservice' now requires auth params to be set.


simahawk:
Fast tracking as tested here https://github.com/OCA/edi/pull/662 

/ocabot merge patch

PR #665, opened 2022-09-14
[FIX] Removed PyPDF2 library from external dependencies

luiseevaquer:
In the requirement file of odoo/odoo the library is pinned to version 1.26.0
    (https://github.com/odoo/odoo/commit/021de21e8cdcf35276c945d8070d04435fa1a19c),
    if library is unpinned, it is forced to install the current version, which
    generates the following error when running odoo:

    AttributeError: module 'PyPDF2' has no attribute 'filters'

### DIFF
--- a/account_invoice_download_ovh/__manifest__.py
+++ b/account_invoice_download_ovh/__manifest__.py
@@ -12,7 +12,7 @@
     "maintainers": ["alexis-via"],
     "website": "https://github.com/OCA/edi",
     "depends": ["account_invoice_download"],
-    "external_dependencies": {"python": ["requests", "ovh"]},
+    "external_dependencies": {"python": ["ovh"]},
     "data": [
         "wizard/ovh_api_credentials_view.xml",
         "views/account_invoice_download_config.xml",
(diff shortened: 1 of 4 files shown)


luisg123v (changes requested):
Since requirements are built up from manifest's requirements, you need to remove dependencies also from there.

In addition, the `requests` requirement is in the same situation.

CC @suniagajose

suniagajose:
> Since requirements are built up from manifest's requirements, you need to remove dependencies also from there.
> 
> In addition, the `requests` requirement is in the same situation.
> 
> CC @suniagajose

Complementing your comment @luisg123v 

https://github.com/odoo/odoo/blob/14.0/requirements.txt#L38

etobella:
requirements.txt is generated automatically from manifests. You need to remove the library from external_dependancies key on the manifest

etobella:
Or pin it there to the expected value

luisg123v (changes requested):
Hi @lescobarvx,

- You need to perform a commit per module. That's because, when modules are migrated, history is brought taking commits related to that module. So, if you leave only only commit, if both modules are migrated, commit will be duplicated in next version
- Missing removing `requests`, that is in the module `account_invoice_download_ovh`

luisg123v (approved):
LGTM :+1: 

@moylop260 could you review, please?

moylop260 (approved):
👍

moylop260:
The CI is even green and the PR is already approved so...

PR #666, opened 2022-09-16
[14.0] edi: drop ack record auto create

simahawk:
Creating the ack was kind of... an hack :)

When I added this behavior we didn't have yet the storage machinery that looks automatically for a new file.
Also for webservices seems not relevant since we can configure endpoints that can create the ack when needed.

All in all, I don't think this feature is needed.
Worse, it adds some overhead when a record gets created and can generate write conflicts when exchanges are processed immediately overlapping with the creation of the ack record.

@etobella @LoisRForgeFlow can you please check if this is something you still need?
If yes, I can deprecate it and make it optional (maybe w/ a flag off by default on the type or on the backend).

### DIFF
--- a/edi_oca/models/edi_backend.py
+++ b/edi_oca/models/edi_backend.py
@@ -381,8 +381,6 @@ class EDIBackend(models.Model):
                 # TODO: run in job as well?
                 self._exchange_output_check_state(rec)
 
-        self._exchange_check_ack_needed(pending_records)
-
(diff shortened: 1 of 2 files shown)


LoisRForgeFlow (approved):
I agree, from our side there is no problem dropping it :+1:

etobella (approved):
If I remember properly, we never used it, so it shouldn't be a problem

PR #672, opened 2022-10-05
[14.0] edi: fix _get_component ctx propagation

simahawk:
The env_ctx that might come from a exchange type settings was not propagated to the exchange record
which is made directly available to the work ctx of the component.

Before this change, if you had any method on the exchange record or on the related record relying on a ctx key,
the ctx key was not propagated to all records' context.


etobella (approved):

etobella:
Makes sene to me :smiley:

florentx (approved):

PR #673, opened 2022-10-08
[14.0] account_invoice_import_simple_pdf: allow to match partners on additionnal fields

alexis-via:
Add a hook to easily inherit the module
account_invoice_import_simple_pdf to match partners on fields other than "vat".

For example, the module l10n_fr_account_invoice_import_simple_pdf takes advantage of this hook to match partners on SIREN (French identification number for companies), cf https://github.com/OCA/l10n-france/pull/377


Doscaal:
:+1:

njeudy (approved):
+1 for the refactoring with possibility to inherit this.

PR #685, opened 2022-11-14
[14.0]edi_account_statement_import module adding

CybroOdooDev:
contributing  edi_account_statement_import module to OCA/edi repository


TeoGoddet:
Hi ! 
Checkout #542 it does the same thing with tests

PR #686, opened 2022-11-15
[14.0] Update pre-commit conf

simahawk:


simahawk:
```
fatal: could not read Username for 'https://gitlab.com/': No such device or address
```
WTH?

simahawk:
> `fatal: could not read Username for 'https://gitlab.com/': No such device or address`

https://github.com/OCA/oca-addons-repo-template/pull/170

PR #687, opened 2022-11-15
[14.0][ADD] product_import

florentx:
Import product catalogues with new prices.


product_import/wizard/product_import.py
```diff
@@ -0,0 +1,209 @@
...
+        if doc_type is None:
+            return {"warning": self._unsupported_file_msg(self.product_filename)}
+
+    def _parse_file(self, filename, filecontent, detect_doc_type=False):
+        assert filename, "Missing filename"
+        assert filecontent, "Missing file content"
+        filetype = mimetypes.guess_type(filename)
+        logger.debug("File mimetype: %s", filetype)
+        mimetype = filetype[0]
+        supported_types = {
```
simahawk: pls drop this abandoned support for CSV, I know it comes from sale_order_import but it should be removed and refactored there too, thanks!

product_import/wizard/product_import.py
```diff
@@ -0,0 +1,209 @@
...
+
+    @api.model
+    def create_product(self, parsed_product, chatter_msg, seller=None, filename=None):
+        product_vals = self._prepare_product(parsed_product, chatter_msg, seller=seller)
+        product = product_vals.pop("recordset", None)
+        if product:
+            product.write(product_vals)
+            logger.info("Product %d updated", product.id)
+        else:
+            product = self.env["product.product"].create(product_vals)
```
simahawk, line 169: We should probably collect data for all prods and call create just once.
Not a blocker, you can add a TODO.

product_import/wizard/product_import.py
```diff
@@ -0,0 +1,209 @@
...
+            logger.info("Product %d created", product.id)
+        return product
+
+    def import_button(self):
+        self.ensure_one()
+        file_content = b64decode(self.product_file)
```
simahawk: Can you delegate the real import to another method that returns something meaningful?
Eg: `_do_import`

product_import/wizard/product_import_view.xml
```diff
@@ -0,0 +1,45 @@
...
+  <record id="product_import_form" model="ir.ui.view">
+    <field name="name">product.import.form</field>
+    <field name="model">product.import</field>
+    <field name="arch" type="xml">
+      <form string="Import Products">
+        <group name="help-import">
+          <div>
+            <p>Upload below the product Catalogue as CSV or XML file. When you click on the Import button:</p>
+            <ol>
+              <li>If it is a CSV file, Odoo will parse it if the appropriate module is installed.</li>
```
simahawk: no csv

simahawk (changes requested):
Can you add some tests too?
You can see how I tested sale_order_import w/ mocks.
You could simulate a method returning catalog information and check that products get updated/created and supplierinfo updated.

product_import/wizard/product_import_view.xml
```diff
@@ -0,0 +1,55 @@
...
+
+    <record id="product_import_form" model="ir.ui.view">
+        <field name="name">product.import.form</field>
+        <field name="model">product.import</field>
+        <field name="arch" type="xml">
+            <form string="Import Products">
+                <group name="help-import">
+                    <div>
+                        <p
+                        >Upload below the product Catalogue as XML file. When you click on the Import button:</p>
```
sebalix: Below it says "If it is an XML file", so I guess we could support different formats there, no?
```suggestion
                        >Upload below the product Catalogue. When you click on the Import button:</p>
```

sebalix (approved):

product_import/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+TODO
```
sebalix: TODO

product_import/wizard/product_import.py
```diff
@@ -0,0 +1,194 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class ProductImport(models.TransientModel):
+    _name = "product.import"
+    _description = "Product import from files"
+
+    product_file = fields.Binary(
+        string="Product Catalogue",
```
simahawk, line 21:
```suggestion
        string="Product Catalogue",
        attachment=True,
```
AFAIR this is not the default in v14

product_import/readme/DESCRIPTION.rst
simahawk: :fire:  :smile:

simahawk (commented):
LG, README missing

product_import/wizard/product_import_view.xml
simahawk: Potentially it can support different formats. Is not the by default. But yeah, this msg should come from the specific module.

product_import/wizard/product_import.py
florentx: It is default in Odoo 14

simahawk (approved):

PR #688, opened 2022-11-15
[14.0][ADD] product_import_ubl

florentx:
Import UBL product catalogues

Depends on:
- [x] #687
- [x] #692


product_import_ubl/wizard/product_import.py
```diff
@@ -0,0 +1,123 @@
...
+        items = self._xpath(path, namespaces=self._ns)
+        return items and items[0].text or False
+
+    get = xpath_get
+    text = xpath_text
+
+
+class ProductImport(models.TransientModel):
+    _name = "product.import"
+    _inherit = ["product.import", "base.ubl"]
```
simahawk: If you use methods from base.ubl that do not depend on an instance, please do not inherit here.
Use it as a tool like for bs doc import.

product_import_ubl/wizard/product_import.py
```diff
@@ -0,0 +1,123 @@
...
+        #  - code: cac:Item/cac:SellersItemIdentification/cbc:ID
+        #  - barcode: cac:Item/cac:StandardItemIdentification/cbc:ID[@schemeID='GTIN']
+        product_vals = self.ubl_parse_product(line, ns)
+
+        # Fill available data
+        ele_price = xline.get(
+            "cac:RequiredItemLocationQuantity/cac:Price/cbc:PriceAmount"
+        )
+        currency = ele_price.attrib.get("currencyID") or False
+        min_qty = float(
```
simahawk, line 65: is this float call safe? No risk to get a bad formatted value? Eg: comma instead of dot?

simahawk (changes requested):
tests tests tests :pray: :)

Also, can you add a test dependency on the other PR?

product_import_ubl/wizard/product_import.py
florentx: OK, I copied it from `sale_import_ubl` I think. I don't know why tool is preferred over composition.  I'll change as you suggest.

product_import_ubl/wizard/product_import.py
florentx: It's as safe as we can ... usually it's even integer in UBL, but the Odoo field is float.

product_import_ubl/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+TODO
```
simahawk: TODO!

product_import_ubl/tests/test_ubl_catalogue_import.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+class TestUblOrderImport(SavepointCase):
+
+    maxDiff = None
+
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        # cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
```
simahawk:
```suggestion
        # cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
```
uncomment this unless you need tracking features

test-requirements.txt
```diff
@@ -4,3 +4,4 @@ responses
 # Python Dependencies
 PyYAML
 xmlunittest
+odoo14-addon-product_import @ git+https://github.com/OCA/edi@refs/pull/687/head#subdirectory=setup/product_import
```
simahawk: Please, always keep this change in a separated commit like `Add TMP test dependency`.
Makes it prominent and, when done, we can simply trash the commit.

simahawk (changes requested):
LG. Missing readme + squash and clean commits

simahawk:
dependency merged, pls check pre-commit

simahawk (commented):
https://github.com/OCA/edi/pull/688/commits/e8991d57dfc69b033e7ba000dfdb6467fd63defc LG but... no specific tests?

simahawk (approved):

PR #689, opened 2022-11-17
[MIG] edi_pdf2data_oca: Migration to 14.0

etobella:
Migration of #433


edi_pdf2data_oca/__manifest__.py
```diff
@@ -0,0 +1,25 @@
...
+    "name": "Pdf2data Import",
+    "summary": """
+        Module that allows to import data from a pdf""",
+    "version": "14.0.1.0.0",
+    "license": "AGPL-3",
+    "author": "CreuBlanca,Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/edi",
+    "depends": ["edi_oca"],
+    "external_dependencies": {
+        "python": ["dateparser", "unidecode"],
```
bosd: Why are you depending on unidecode? I don't see it used here. FYI, We are considering removing it from invoice2data because of license incompatability.

requirements.txt
```diff
@@ -1,4 +1,5 @@
 # generated from manifests external_dependencies
+dateparser
 dateparser==1.1.1
```
bosd: This is double, You need to pin a specific version here?

sanchonuria (approved):

kluna1998 (approved):

edi_pdf2data_oca/__manifest__.py
etobella: True! Removed

PR #691, opened 2022-11-25
[14.0][ADD] edi_account_invoice_import_ubl

LoisRForgeFlow:
This module plugs account_invoice_import_ubl into EDI machinery.

@ForgeFlow


JoanSForgeFlow (approved):

edi_account_invoice_import_ubl/data/edi_exchange_type.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
```
simahawk, line 2: I'd recommend setting noupdate here.

Additionally, due to the nature of the exchange type, that can be highly configurable, I tend to say we should only have demo data for them and use post init hooks to setup default data if needed.

I know you found the example that I left w/ edi_sale_order_import but I'm using this approach since a while and I find it more reliable. 

Furthermore, I think is not worth to have one module for import and one module for export and I will deprecate edi_sale_order_import to go for `edi_sale_ubl_oca`like a did for purchase in #669 (not finished yet)

simahawk: Well, this is more a question for `edi_account_invoice_import` rather than this module :sweat_smile:

LoisRForgeFlow: @simahawk I think that it makes sense to add everything together (outgoing and incoming UBL invoices) in one module, I just split it because of what you mentioned: I followed your template in sales and because the "base" modules are already split account_invoice_import_ubl (incoming) and account_invoice_ubl (outoing). I think this "merge" could be part of a migration because it needs to be done in the sales module anyway, and at least now the structure is consistent.

Regarding the installation of default exchange types, it is the same, I followed your example on sales, but I have no problem to switch to the pre-init hooks style if you provide me another example.

Thanks for having a look!

LoisRForgeFlow: ping @simahawk

PR #692, opened 2022-11-25
[14.0][IMP] product_import: handle company and active flag

florentx:
Changed:
 - import product catalogue for a specific company
 - allow to archive/unarchive products

### DIFF
--- a/product_import/wizard/product_import.py
+++ b/product_import/wizard/product_import.py
@@ -92,9 +92,30 @@ class ProductImport(models.TransientModel):
         return catalogue
 
     @api.model
+    def _get_company_id(self, catalogue):
+        company_vals = catalogue.get("company")
+        if not company_vals:
+            return False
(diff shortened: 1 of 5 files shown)


product_import/wizard/product_import.py
```diff
@@ -91,10 +91,22 @@ def parse_product_catalogue(self, product_file, product_filename):
...
 
+    @api.model
+    def _get_company(self, catalogue):
+        part = self._bdimport._match_partner(
+            catalogue["company"],
+            catalogue["chatter_msg"],
+            partner_type="contact",
+            domain=[("is_company", "=", True)],
+        )
+        return part.ref_company_ids[:1]
```
simahawk: This field comes from `account` module. I don't think we want this dependency.
You can search for companies instead.
Am I missing anything?

simahawk (approved):

PR #693, opened 2022-11-29
[FIX] edi: output might be not UTF-8 encodable, like PDFs

sanchonuria:
This was not ported on 14

@etobella

PR #695, opened 2022-12-01
[FIX] edi: output might be not UTF-8 encodable, like PDFs

sanchonuria:
This was not ported on 15 #693

@etobella


etobella (approved):

etobella:
Can you review CI failures please?

simahawk (approved):

PR #698, opened 2022-12-06
[14.0] edi_exchange_template: fix rendering

simahawk:

ref: cos-3786


edi_exchange_template_party_data/models/edi_exchange_template_output.py
```diff
@@ -13,9 +13,9 @@ class EDIExchangeOutputTemplate(models.Model):
     _inherit = "edi.exchange.template.output"
 
     def _get_render_values(self, exchange_record, **kw):
```
sebalix, line 15: Is the `kw` parameter still used? It's not pass to `super` anymore?

sebalix (approved):

PR #700, opened 2022-12-21
[13.0][IMP] edi_backend_partner: modify backend view

JaumeBforgeFlow:
Right now when configuring a partner for a backend it appears right next to the backend type when saving. Now the partner will be shown in the next line.

### DIFF
--- a/edi_backend_partner/views/edi_backend_views.xml
+++ b/edi_backend_partner/views/edi_backend_views.xml
@@ -7,7 +7,7 @@
         <field name="model">edi.backend</field>
         <field name="inherit_id" ref="edi.edi_backend_view_form" />
         <field name="arch" type="xml">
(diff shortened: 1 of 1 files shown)


AaronHForgeFlow (approved):

JordiMForgeFlow (approved):

MiquelRForgeFlow (approved):

LoisRForgeFlow (commented):
Thanks for taking care of this! In my opinion it would be better placed below and with the label as any other field.

JaumeBforgeFlow:
Hi @LoisRForgeFlow, i've updated the code following your feedback and moved the partner field. 

image

MiquelRForgeFlow:
FYI https://github.com/OCA/edi/pull/657 exists.

LoisRForgeFlow (approved):

PR #707, opened 2022-12-26
[16.0][MIG] account_invoice_facturx_py3o

alexis-via:


alexis-via:
Requires https://github.com/OCA/edi/pull/706

PR #710, opened 2022-12-30
[14.0][MIG] edi_sale_oca

JaumeBforgeFlow:
Migration to v14 for #709


simahawk (changes requested):
Sorry, it maybe passed unnoticed, but I have this module here already and it's more "complete" https://github.com/OCA/edi/pull/759.

In there I don't use a specific event for the action confirm because I use edi_exchange_type_auto but I can add it if you need it.

JaumeBforgeFlow:
Hi @simahawk, it's ok I will close this one. We don't need the listener I was adding it since other modules have them.

PR #720, opened 2023-02-06
[13.0][ADD] edi_punchout

hbrunn:
This module implements several punchout protocols:

- IDS (very DE specific)
- OCI (quite EU specific)

Allowing users to transfer orders from a webshop supporting the one of the above to Odoo and the other way around.

Note that the current implementation doesn't protect the login credentials of the webshop because of the way the protocols work, so you should enable this functionality only for trusted users.

The above could be helped by implementing cXML or the secure OCI extension, which is not currently the case.


gurneyalex:
Hello @hbrunn 

For the record I'm currently designing for a customer (in 14.0) a cXML punchout implementation. I think there are things we could share in there, but probably making things modular would help. Among the things I had in mind in my (WIP) design:

* base_punchout module: providing a punchout_session model, recording the punchout protocol, and adding that field on a sale.order. Possibly adding a way to enable punchout on a customer (res.partner) or on a portal user. In my vision the punchout session would use OCA components for protocol implementation, but this is not really a "must have" feature, at least not in the first version. 

* punchout_cxml_quotation: implement the base cXML support (Profile endpoint, punchout initiation, punchout "checkout"). 

* website_sale_punchout_cxml: implement the cXML checkout on the odoo eshop (split out because our customer is using shopinvader for the eshop). 

Do you think it would be possible to refactor your module to split out a base layer?

chienandalu:
@gurneyalex I'd say @hbrunn is working in the opposite behavior with Odoo being in the buyer position instead of the supplier.

In any case, we're indeed working in what you're devising :) Here's a sneak peak o very early work on it (more in progress to be released by my partner @ernesto-garcia-tecnativa ): https://github.com/Tecnativa/e-commerce/commit/ce99a5bfa08600e510d49d74857d99b15b43f462 

I you want to we can meet and devise possible common efforts :smiley:

hbrunn:
yes @gurneyalex @chienandalu this is indeed the other way around.

If I may provide input for the server side: Some suppliers allow to create an account with permissions to only create carts and check them out via punchout (no option to really order at that point), and only send them readonly when they come in via IDS/OCI/cXML. That's a very handy feature for implementing purchase workflows with those protocols to avoid mistakes with manual changes.

bosd:
@hbrunn Can yuo attend to this PR? 🙏

hbrunn:
@bosd done

edi_punchout/data/uom_uom.xml
```diff
@@ -0,0 +1,34 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!-- Copyright 2023 Hunki Enterprises BV
+     License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). -->
+<odoo>
+
+    <record id="uom.product_uom_unit" model="uom.uom">
```
bosd, line 6: This file contains the same data as the oca module `uom_unece`.
https://github.com/OCA/community-data-files/blob/13.0/uom_unece/data/unece.xml
Maybe easier to depend on that module?

edi_punchout/views/edi_punchout_account.xml
```diff
@@ -0,0 +1,199 @@
...
+                        <field
+                                name="ids_kndnr"
+                                attrs="{'invisible': [('protocol', '!=', 'ids')]}"
+                            />
+                        <field
+                                name="ids_name_kunde"
+                                attrs="{'invisible': [('protocol', '!=', 'ids')]}"
+                            />
+                        <field
+                                name="ids_pw_kunde"
```
bosd, line 41:
```suggestion
                                name="ids_pw_kunde" password="1"
```

edi_punchout/views/templates.xml
```diff
@@ -0,0 +1,65 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!-- Copyright 2023 Hunki Enterprises BV
+     License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). -->
+<odoo>
+
+    <template id="ids_send_cart">
+        <Warenkorb xmlns="http://www.itek.de/Shop-Anbindung/Warenkorb/">
```
bosd, line 7: Why is this hard coded to this URL?
I don't see this view appear in runboat. Wher is it supposed to be?

bosd (commented):
Thanks for coming back to this one! 

Basic code review and test on runboat.
Some questions :face_in_clouds:

edi_punchout/data/uom_uom.xml
hbrunn: no, that's conceptually different things. Unece just happens to have the the same value here.

edi_punchout/views/templates.xml
hbrunn: that's what the [docs](https://www.itek.de/fileadmin/user_upload/itek-website/Beratung/Standardisierung/IDS_Connect/2.5/20201225_IDS_Schnittstelle_2_5.pdf) say the correct namespace for this kind of xml is. And you won't see this view in runboat, this is what generates the xml being sent to the webshop when using the ids protocol.

bosd: Thanks, makes sense.

bosd (approved):

bofiltd (approved):
LGTM

edi_punchout/views/edi_punchout_account.xml
bosd: @hbrunn What do you think about this suggestion?

hbrunn: seems sensible. Note however I'm not on the project using this any more, so I'll be slow in updating this PR

bosd (approved):

bosd:
Should we merge this one?

bosd:
> Note that the current implementation doesn't protect the login credentials of the webshop because of the way the protocols work, so you should enable this functionality only for trusted users.

Is this still applicable?

hbrunn:
yes

PR #721, opened 2023-02-07
[IMP] edi_storage_oca: Delete processed files if you want it

etobella:
This change allows us to delete the processed file if you want to

@simahawk @LoisRForgeFlow

### DIFF
--- a/edi_storage_oca/components/listener.py
+++ b/edi_storage_oca/components/listener.py
@@ -18,15 +18,23 @@ class EdiStorageListener(Component):
             # The file might have been moved after a previous error.
             return False
         self._add_after_commit_hook(
-            storage._move_files, [(from_dir / filename).as_posix()], to_dir.as_posix()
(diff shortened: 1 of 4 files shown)


edi_storage_oca/components/listener.py
```diff
@@ -18,15 +18,23 @@ def _move_file(self, storage, from_dir_str, to_dir_str, filename):
             # The file might have been moved after a previous error.
             return False
         self._add_after_commit_hook(
-            storage._move_files, [(from_dir / filename).as_posix()], to_dir.as_posix()
+            storage.move_files, [(from_dir / filename).as_posix()], to_dir.as_posix()
```
etobella, line 21: _move_files is deprecated https://github.com/OCA/storage/blob/14.0/storage_backend/models/storage_backend.py#L128

LoisRForgeFlow (approved):
Nothing against :+1:

PR #567, opened 2022-04-23
[FIX] invoice2data Fallback to tesseract

bosd:
The invoice2data module usage instructions include:

> If you want the invoice2data library to fallback on OCR if the PDF doesn't contain text (only a very small minority of PDF invoices are image-based and require OCR), you should also install [Imagemagick](http://www.imagemagick.org/)

However, it does not fallback if pdftotext fails.
This code fixes that..

PR #722, opened 2023-02-10
[FIX] invoice2data: Fallback to tesseract

bosd:
This is a continuation of https://github.com/OCA/edi/pull/567
Rebased on top of the latest code. Which allows import of the pdf document, even if no data could be extracted.

### DIFF
--- a/account_invoice_import_invoice2data/wizard/account_invoice_import.py
+++ b/account_invoice_import_invoice2data/wizard/account_invoice_import.py
@@ -5,7 +5,7 @@
 import logging
 import os
 import shutil
-from tempfile import mkstemp
+from tempfile import NamedTemporaryFile
 
 from odoo import _, api, fields, models, tools
 from odoo.exceptions import UserError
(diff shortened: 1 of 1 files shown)


bosd:
@bofiltd Would you like to review this one as well?
Made some minor improvements.. 
The logger now shows when it is trying to fallback on tesseract.

bosd:
Just pushed an empty commit. To recreate runboat, using the latest release of invoice2data (0.4.2) where the tesseract bug has been fixed.

bofiltd (approved):
LGTM! :+1:

bosd:
@etobella Could you please review? 🙏

account_invoice_import_invoice2data/wizard/account_invoice_import.py
```diff
@@ -55,9 +60,30 @@ def invoice2data_parse_invoice(self, file_data):
...
         if not invoice2data_res:
-            return False
+            if not shutil.which("tesseract"):
+                logger.warning(
+                    "Fallback on tesseract impossible, Could not find the utility. "
+                    "Hint: sudo apt install tesseract-ocr"
+                )
+                return False
+            # Fallback on tesseract
+            logger.info("PDF Invoice parsing failed: Falling back on Tesseract ocr")
```
etobella, line 72: If you channed the log message for Invoice parsing failed on top, can you keep the same logic?

etobella (commented):
In general, ILGM, just a small comment

account_invoice_import_invoice2data/wizard/account_invoice_import.py
bosd: @etobella Thanks for the review. 👍 
Sorry, But I don't understand the change you suggesting

bosd: Reverted the change, as hard coding the file types to PDF.
 
Invoice2data is capable of handling different file formats.
Sadly, we cannot support those different file types.

Because the uploaded file is converted to a base64 string.
When it is passing trough all the account_invoice_import functions, some data got lost.
(file name, mimetype)

Ideally we would have that data available to use in other functions.
(it could improve tesseract ocr results)
But currently we don't have it.

So, I'm reverting the log message.

bosd:
Also found 2 small bugs.
Will include an separate commit in this pr to fix it.

1. To be able to parse the pdf file correctly by the latest versions of the invoice2data lib. The tempfile must have the extension `.pdf`
2. The generated temp files are not correctly handled. They never get deleted when the parsing has errored or finished.

bosd:
@etobella  Is this ok now?

etobella (approved):
Makes sense to me

bosd:
@simahawk Can you merge this? 🙏

PR #550, opened 2022-03-22
[MIG] base_ubl: Migration to 15.0

tbanevicius:
Migration of module base_ubl to 15.0 Odoo version

PR #730, opened 2023-02-27
Migrate base_ubl to 15.0

aisopuro:
This PR replaces #550 

That PR has grown stale, due to the CI being unable to pass tests. So I wanted to rebase to see where the tests are actually failing, in hopes I can help move this forward.

Original commit authors should remain, I don't want to take undue credit.


base_ubl/tests/test_ubl_generate.py
```diff
@@ -58,7 +52,15 @@ def setUpClass(cls, chart_template_ref=None):
         cls.invoice.partner_id.street2 = "Test Street 2"
         cls.invoice.partner_id.parent_id = cls.invoice.company_id.partner_id
 
+    def disable_other_embeddings(self, company):
+        # In a test context, we may be running with other modules from this
+        # same repository that have activated embedding documents that we are
+        # not testing. So for our tests we turn them off if we know about them.
+        if "xml_format_in_pdf_invoice" in company._fields:
+            company.xml_format_in_pdf_invoice = "none"
```
aisopuro, line 60: I'm not 100% sure about this approach, but couldn't think of anything better with my schedule. A proper fix would likely be to question whether it is reasonable for the Factur-X module to make embedding it the default functionality, and that seems like a discussion that is out of scope for this PR.

But as a heads-up for reviewers, you may want to look at this one.

PR #731, opened 2023-03-01
[14.0] edi_endpoint: cross worker fix 

simahawk:
Replaces #633 

Former version of endpoint_route_handler had a major flaw:

routing rule registry was not properly shared across workers
forcing us to restart the instance to make sure all envs were inline w/ it.

This change adapts edi_endpoint_oca to the new version which contains
some refactoring.

Depends on:

- [x] https://github.com/OCA/web-api/pull/3


simahawk:
I don't really understand WTF is going on here. I'll try to explain here (also to talk to my GH rubber duck :grin: ).
On HttpCase tests that run on github the mixins are not loaded properly. Works smoothly on local tests or real instance.
Here's the hierarchy:

* `edi.endpoint` intherits from `endpoint.mixin`
* `endpoint.mixin` intherits from `endpoint.route.handler`
* `endpoint.route.handler` inherits from `endpoint.route.sync.mixin`

The error:
```
 2023-05-12 07:53:20,861 1792 ERROR odoo unittest.suite: ERROR: setUpClass (odoo.addons.edi_endpoint_oca.tests.test_edi_endpoint_controller.EDIEndpointHttpCase)
  File "/opt/odoo/odoo/tests/common.py", line 173, in _handleClassSetUp
    setUpClass()
  File "/__w/edi/edi/edi_endpoint_oca/tests/test_edi_endpoint_controller.py", line 23, in setUpClass
    cls.env["edi.endpoint"].search([])._handle_registry_sync()
AttributeError: 'edi.endpoint' object has no attribute '_handle_registry_sync'
```
`_handle_registry_sync` is a method of  `endpoint.route.sync.mixin` so it should available on all inheriting models (`edi.endpoint` in this case).

It's like the model is not fully registered in the registry at that point.

Right now I added [a dummy test]([test sync on GH](https://github.com/OCA/edi/pull/731/commits/b20a7f730769ab3536b7e810292df5bf88daa8c4)) to call that method in a normal SavepointCase test

simahawk:
> Right now I added [a dummy test]([test sync on GH](https://github.com/OCA/edi/pull/731/commits/b20a7f730769ab3536b7e810292df5bf88daa8c4)) to call that method in a normal SavepointCase test

Ok. it fails on SavepointCase too... at least some consistency... :smile:

simahawk:
Ok, I can't believe it... now it passes :thinking:

simahawk:
> Ok, I can't believe it... now it passes thinking

I think I've found the reason: declaring again the dependency on the sync mixin messed up the inheritance.
No clue why only on GH and not locally :shrug:

simahawk:
As the main [PR](https://github.com/OCA/web-api/pull/3) has been merged  this is required now to make edi_endpoint work. I tested everything again and it works smoothly on my side. I merge here but if anyone spots something wrong/weird just ping me :)

/ocabot merge nobump

edi_endpoint_oca/models/edi_endpoint.py
```diff
@@ -18,7 +13,7 @@ class EDIEndpoint(models.Model):
     """
 
     _name = "edi.endpoint"
-    _inherit = "endpoint.mixin"
+    _inherit = ["endpoint.mixin"]
```
sebalix, line 16: Your test issue was coming from there? Should it deserves a comment explaining this, someone that will migrate this later could remove these square brackets.

simahawk: no no, this is a left over / non necessary change. I added  the sync mixin here 1st then I removed it ;)

PR #733, opened 2023-03-09
account_invoice_import: FIX Partner creation ref + Improve comments

bosd:


yaseentai (approved):
LGTM
code review

bofiltd (approved):
LGTM :+1:

alexis-via (approved):

alexis-via:
While forward-porting this change in v15 (to prepare v16 mig), I was thinking again on the use of 'default_ref' in the method _prepare_new_partner_context():
AFAIK, the field 'ref' on res.partner is used as an internal reference. So there is no reason to have it in the UBL or Factur-X invoice you're receiving from your supplier. That's why I don't think it's useful to use it in this method and to mention it in the pivot dict format. OK, it doesn't hurt to have it, but it's not useful in real life.

Or maybe I miss something about the use of this 'ref' field ?

bosd:
> Or maybe I miss something about the use of this 'ref' field ?

It has a usecase as some invoices contains a customer number generated by the supplier.
In hindsight, it is'nt a critical thing to miss. As no autmatic partner matching seems to be done based on the field.

bosd:
Oopsie,
It seems it is being used for matching.


https://github.com/OCA/edi/blob/be96fdfa94334bf44675f649f1ea40358604ea53/base_business_document_import/models/business_document_import.py#L152-L159

PR #735, opened 2023-03-14
[14.0] Update oca_dependencies.txt, webservice module in web-api

angelmoya:


yaseentai (approved):
LGTM
code review

simahawk:
Thanks for your change. However, this file is not needed anymore.

@sbidoul IMO we should delete all oca_dependencies.txt from all the repos where we switched CI to use PIP mode.
WDYT?

sbidoul:
> IMO we should delete all oca_dependencies.txt from all the repos where we switched CI to use PIP mode.
WDYT?

@simahawk AFIK this file is not present in any of 15 and 16 branches. For older branches I guess I don't mind if people maintain it manually if they need it for existing workflows. But for sure we should not reintroduce it in newer branches.

PR #741, opened 2023-03-24
[16.0][MIG] account_invoice_import_simple_pdf

alexis-via:


icf20:
is this abandoned?

alexis-via:
@icf20 no ! It works well. It depend on account_invoice_import that is not merged because I want to make important changes in it

icf20:
> @icf20 no ! It works well.

thx for the reply, while my reply is late here it is, have to drop the interest in this the addon cause is lagging too much behind, odoo 18 (17.4) has electronic invoices and automatic tax reporting, so upgrade is a must before 2024-12-31, and this add-on is stuck on 14 or 16PR hell, i dont know if is intentional from odoo part but feels like they are pushing out all 3rd party companies offering services with their yearly upgrades
best of luck

victor-champonnois:
@alexis-via Hi, do you think this PR is usable in production, if we also deploy this https://github.com/OCA/edi/pull/753 released at a given stable commit ?

bguillot:
> @alexis-via Hi, do you think this PR is usable in production, if we also deploy this #753 released at a given stable commit ?

Hello, The answer is late but we use it in production at Akretion. hope it helps

bguillot (approved):
Used in production, works correctly

PR #742, opened 2023-03-24
[16.0][MIG] account_invoice_import_facturx

alexis-via:


smaddlsoft:
Hi everyone,

and thanks @alexis-via for the great work on this module 🙌

Is this the latest version of the v16 module of account_invoice_import_facturx, and is it working for you?

While testing the import of a Factur-X invoice, we ran into the following error when clicking Import in the wizard:
"AssertionError
account_invoice_import.py, line 1011
assert not inv_cur.compare_amounts(invoice.amount_total, parsed_inv['amount_total'])"

Any help appreciated
best

Jan

smaddlsoft:
@alexis-via thanks for the work on this module. 

Can you confirm you removed the "Update" scenario in account_invoice_import as outlined in your proposed changes [here](https://github.com/OCA/edi/issues/803)?

Did you also manage to work on the "wizard that makes the links between the invoice lines and the purchase order lines and compare the prices. This wizard would be independent from the account.invoice.import wizard." ?

Thanks in advance
Jan

alexis-via:
@smaddlsoft I confirm that the "Update" scenario as been removed.
I haven't worked myself on the wizard that makes the links between the invoice lines and the purchase order lines and compare the prices, but I think there are already some existing solutions (I heard about one at least, but not in the OCA). We should investigate more to identify the existing solutions and see if we can use them or if we should develop a new one.

alexis-via:
@smaddlsoft We should start by evaluating the feature available in the purchase module : https://github.com/odoo/odoo/commit/7ba5cbb00cb651d384ead556f8550d8314b4ab7b (available since 18.0).

bguillot (approved):
Used in production

smaddlsoft:
@alexis-via thanks for the feedback on the update scenario and wizard. I agree working on top of the native Odoo bill matching sounds meaningful. If you recall the non-OCA options you already saw or heard about and want to share, appreciated!

PR #744, opened 2023-03-28
[13.0][FIX] edi_voxel_sale_order_import: Tests fail when testing with other modules. Remove unnecessary company

sergio-teruel:
cc @Tecnativa TT42192

ping @ernestotejeda @stefan-tecnativa 

The tests fails with other modules like as account_payment_sale. The module account_payment_mode adds default values for payment mode in partners, so if ww create a so in other company we have an error because the companies are different. I Think that we do not need this extra company record in tests for this module...


ernestotejeda (approved):

carlosdauden (approved):

CarlosRoca13 (approved):

pedrobaeza:
/ocabot merge patch

Is this needed in upper versions?

PR #746, opened 2023-04-12
[16.0][MIG] edi_exchange_template_oca: Migration to 16.0

john-herholz-dt:


simahawk:
To be moved to [OCA/edi-framework](https://github.com/OCA/edi-framework)

PR #748, opened 2023-04-12
[16.0][MIG] edi_storage_oca: Migration to 16.0

john-herholz-dt:
Basic migration. Is blocked by storage dependend modules to be migrated.


simahawk:
To be moved to [OCA/edi-framework](https://github.com/OCA/edi-framework)

PR #749, opened 2023-04-13
[MIG][16.0] edi_oca

john-herholz-dt:


simahawk:
@john-herholz-dt thanks for the work.
Regarding runboat: seems green.

Overall looks good but there are some left overs from v15 to be removed. Could you check pls?
Last but not least: can you move this PR to OCA/edi-framework pls?

You can:
1. checkout that repo
2. add your OCA/edi fork as remote
3. use the usual format-patch command but use your remote and branch for this PR

test-requirements.txt
```diff
@@ -0,0 +1,2 @@
+mock
```
simahawk: pls remove `mock` from here. Use `from unittest import mock`.

simahawk (changes requested):
to be moved and fixed

simahawk:
Moved here https://github.com/OCA/edi-framework/pull/1

PR #758, opened 2023-05-10
[MIG] account_invoice_facturx: Migration to v16

bfigueiredo-odoogap:
Migrating the Module **account_invoice_facturx**, and staying all the functionality.

@alexis-via , i watched your migration of this module and i make some fixes on the country_id and on the unit_tests


pedrobaeza:
Thanks for the contribution.

Please preserve commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-16.0.

If the jump is between several versions, you have to modify the source branch in the main command to accommodate it to this circumstance.

PR #760, opened 2023-05-15
[14.0][ADD] edi_product_import

florentx:
New glue add-on for `product_import` ~and `product_import_ubl`~


edi_product_import_ubl/__manifest__.py
```diff
@@ -0,0 +1,15 @@
...
+{
+    "name": "EDI Product import UBL",
+    "summary": """Plug product_import_ubl into EDI machinery.""",
+    "version": "14.0.1.0.0",
+    "development_status": "Alpha",
+    "license": "AGPL-3",
+    "website": "https://github.com/OCA/edi",
+    "author": "Camptocamp, Odoo Community Association (OCA)",
+    "depends": ["edi_ubl_oca", "edi_product_import", "product_import_ubl"],
+    "auto_install": True,
```
simahawk: I would not enable auto install here

edi_product_import_ubl/data/edi_exchange_type.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
```
simahawk: This should be flagged as noupdate at least.
However, recently I've been moving such data to demo (hence making this module just an example) because the risk of impacting existing edi configurations is too high from my POV.

simahawk (changes requested):
Can you please split the commit in 1 per module?

edi_product_import_ubl/data/edi_exchange_type.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8" ?>
```
simahawk, line 1: As you decided to keep it, can you please put this in the roadmap? "Evaluate dropping this module in future versions since it's just a default configuration that can be described in docs."
TBH, the more I think / write about this, the more I'm not convinced we want to carry on this module :sweat_smile:

simahawk: @florentx ping :)

simahawk:
Replaced by https://github.com/OCA/edi/pull/882

PR #762, opened 2023-05-16
[14.0] edi + edi_webservice: add 'as_bytes' option for binary files

simahawk:

Pulled out of #669

### DIFF
--- a/edi_oca/models/edi_exchange_record.py
+++ b/edi_oca/models/edi_exchange_record.py
@@ -198,13 +198,16 @@ class EDIExchangeRecord(models.Model):
             output_string = bytes(output_string, encoding)
         self[field_name] = base64.b64encode(output_string)
 
-    def _get_file_content(self, field_name="exchange_file", binary=True):
+    def _get_file_content(
+        self, field_name="exchange_file", binary=True, as_bytes=False
+    ):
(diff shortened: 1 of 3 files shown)


edi_oca/models/edi_exchange_record.py
```diff
@@ -198,13 +198,16 @@ def _set_file_content(
...
         self[field_name] = base64.b64encode(output_string)
 
-    def _get_file_content(self, field_name="exchange_file", binary=True):
+    def _get_file_content(
+        self, field_name="exchange_file", binary=True, as_bytes=False
+    ):
         """Handy method to not have to convert b64 back and forth."""
         self.ensure_one()
         if not self[field_name]:
             return ""
```
florentx, line 207: For consistency, do we need to manage this case as well?
```suggestion
            return "" if not as_bytes else b""
```

florentx (approved):
See comment about corner case.
https://github.com/OCA/edi/pull/762/files#r1194767535

edi_oca/models/edi_exchange_record.py
simahawk: hmm I don't think we need it here.
This is only for sending. If you want to manipulate the file in other circumstances you can do it in custom code.

PR #763, opened 2023-05-16
[14.0] edi_oca: avoid ghost exchange records

florentx:
Unrestricted access to admin users (group "Settings")

Having exchange records with non-existent `res_id` does not happen normally.

Main use case is during testing phase, when related records can be deleted for cleaning.
I've proposed PR with minimal change: to be able to retrieve/delete these records when needed (group Admin).

### DIFF
--- a/edi_oca/models/edi_exchange_record.py
+++ b/edi_oca/models/edi_exchange_record.py
@@ -4,9 +4,12 @@
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
 
 import base64
+import logging
 from collections import defaultdict
 
 from odoo import _, api, exceptions, fields, models
+
+_logger = logging.getLogger(__name__)
 
 
 class EDIExchangeRecord(models.Model):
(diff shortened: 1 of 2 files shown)


edi_oca/models/edi_exchange_record.py
```diff
@@ -420,8 +420,8 @@ def _search(
             count=False,
             access_rights_uid=access_rights_uid,
         )
-        if self.env.is_superuser():
-            # rules do not apply for the superuser
+        if self.env.is_system():
+            # rules do not apply to group "Settings"
```
simahawk: I'd say that EDI admins should be able to see them too. WDYT?

simahawk (commented):
Can we do anything at UI level to make clear that a record was deleted?
Can't we determine that state by checking a model is set but no res_id?

edi_oca/models/edi_exchange_record.py
florentx: Well, for security purpose I prefer not to allow EDI admins at this stage. I see no use case for this, since deleting consumer records is not expected in Production. This PR is mainly to troubleshoot and remove ghost records before going live.

When consumer record is deleted, then `model` and `res_id` are still set on `edi.exchange.record`. So no easy way to show it in UI.
Since it's corner case, I don't think we need more time on it.

simahawk: Ok, but we have the relation to the original EDI record: I would flush them or set them in a way that makes it easier to recognize that they are "orphaned".
Not a blocker here, however, could you please add a TODO? Thanks!

simahawk (approved):

edi_oca/models/edi_exchange_record.py
```diff
@@ -453,11 +461,21 @@ def _search(
         for model, targets in model_data.items():
             if not self.env[model].check_access_rights("read", False):
                 continue
-            target_ids = list(targets)
+            recs = self.env[model].browse(list(targets))
```
simahawk, line 464: Just a remark: as this runs on every search it might add overhead when working on large datasets.

PR #765, opened 2023-05-17
[14.0] edi_oca: assertion was not executed

florentx:


edi_oca/tests/test_record.py
```diff
@@ -42,9 +42,7 @@ def test_record_validate_state(self):
                 "edi_exchange_state": "output_pending",
             }
             self.backend.create_record("test_csv_input", vals)
-            self.assertEqual(
-                err.exception.name, "Exchange state must respect direction!"
-            )
+        self.assertEqual(err.exception.name, "Exchange state must respect direction!")
```
simahawk: It would be probably better to use `assertRaisesRegex` instead

florentx: Well. Assertion is explicit enough. Using regex flavor is matter of taste :slightly_smiling_face:

simahawk: Actually is not. Something else in the stack could throw that same type of exception and you won't notice it.
Now, ideally, we should have a specific exception instead of this generic one... then you'll be 100% right :wink: 
I mean, the 1st assert alone, of course :)

simahawk (approved):
Not blocking but I'll take a note about using specific exceptions for framework specific errors.

PR #768, opened 2023-05-22
[14.0] base_business_document_import: Fix company domain

florentx:
Read company from Env, not from the current user.
This way the `Env.with_company(...)` works as expected.

Note: `with_context(force_company=...)` is deprecated in odoo 14 (or earlier?) in favor of `with_company(...)`

### DIFF
--- a/base_business_document_import/models/business_document_import.py
+++ b/base_business_document_import/models/business_document_import.py
@@ -288,9 +288,8 @@ class BusinessDocumentImport(models.AbstractModel):
         partner = self._direct_match(partner_dict, rpo, raise_exception=raise_exception)
         if partner:
             return partner
-        company_id = self._context.get("force_company") or self.env.company.id
         domain = domain or []
-        domain += ["|", ("company_id", "=", False), ("company_id", "=", company_id)]
+        domain += self._match_company_domain()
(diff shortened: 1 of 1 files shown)


base_business_document_import/models/business_document_import.py
```diff
@@ -506,12 +505,9 @@ def _match_partner_bank(
...
         bankaccount = rpbo.search(
-            [
-                "|",
-                ("company_id", "=", False),
-                ("company_id", "=", company_id),
+            self._match_company_domain()
+            + [
                 ("sanitized_acc_number", "=", iban),
                 ("partner_id", "=", partner.id),
             ],
```
mmequignon, line 513: Can't we use `odoo.osv.expression.AND(self._match_company_domain, […])`?

mmequignon (approved):
Not blocking, but I would use `odoo.osv.expression.AND`

TDu (approved):

base_business_document_import/models/business_document_import.py
florentx: If we do that, it should be done uniformly across the file, to stay consistent.
YAGNI.

simahawk (approved):
LG

klodr (approved):

PR #769, opened 2023-05-22
[14.0] product_import: Fix product's company

florentx:
And also:

- [ ] OCA/edi#768

### DIFF
--- a/product_import/wizard/product_import.py
+++ b/product_import/wizard/product_import.py
@@ -129,6 +129,8 @@ class ProductImport(models.TransientModel):
             for s_info in product.seller_ids:
                 if s_info.name.id != seller_info["name"]:
                     continue
+                if s_info.company_id.id not in (seller_info["company_id"], False):
+                    continue
                 if s_info.date_end and s_info.date_end < today:
                     continue
                 if (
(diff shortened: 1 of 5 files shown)


mmequignon (approved):

TDu (approved):

product_import/wizard/product_import.py
```diff
@@ -147,9 +147,12 @@ def _prepare_supplierinfo(self, seller_info, product):
 
     @api.model
     def _prepare_product(self, parsed_product, chatter_msg, seller=None):
+        product_company = self.env.context.get("product_company", False)
```
grindtildeath: I find it weird to default to a False value and then pass it to `with_company` that won't actually change the company if we have a False value. Wouldn't it make more sense to default to `self.env.company`?

grindtildeath (approved):
Not blocking

product_import/wizard/product_import.py
florentx: I've refactored, and no longer use `with_company` here.

product_import/wizard/product_import.py
```diff
@@ -215,7 +231,7 @@ def import_button(self):
             raise UserError(_("This catalogue doesn't have any product!"))
         company_id = self._get_company_id(catalogue)
         seller = self._get_seller(catalogue)
-        self.with_context(company_id=company_id)._create_products(
+        self.with_context(product_company=company_id)._create_products(
```
simahawk:
```suggestion
        self.with_context(product_company_id=company_id)._create_products(
```
to be more consistent

PR #771, opened 2023-05-25
[10.0] account_invoice_ubl : Add support for non vat taxes in Peppol BIS Billing 3.0

jbaudoux:
* [FIX] account_invoice_ubl: Fix embedded pdf report - Fix the call to generate the embedded pdf report
* [FIX] base_ubl: schemeID for GTIN is 0160
*  [FIX] account_invoice_ubl: Support non-VAT taxes
While Odoo allows to define taxes other than VAT, the Peppol BIS Billing 3.0 does not allow taxes that are non VAT. To solve this, those taxes not subject to VAT are converted as a global extra charge exempt of VAT.
* [FIX] Compute properly the line unit price
* [ADD] Support line discount (and triple discount). The discount is declared in a new line section as AllowanceCharge.
* [FIX] The base unit price is for 1 Unit. In case of tax included in the price, the base unit price must be for the total quantity to prevent rounding issue.
* [FIX] Credit Note: a credit note must follow the CreditNote schema and not the InvoiceSchema

I will fwd this to the other versions after reviews

@alexis-via @thomaspaulb @lmignon

### DIFF
--- a/account_invoice_ubl/models/account_invoice.py
+++ b/account_invoice_ubl/models/account_invoice.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 # © 2016-2017 Akretion (Alexis de Lattre <alexis.delattre@akretion.com>)
+# Copyright 2023 Jacques-Etienne Baudoux (BCIM) <je@bcim.be>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 from odoo import models
@@ -22,11 +23,13 @@ class AccountInvoice(models.Model):
         doc_id.text = self.number
         issue_date = etree.SubElement(parent_node, ns['cbc'] + 'IssueDate')
         issue_date.text = self.date_invoice
-        type_code = etree.SubElement(
-            parent_node, ns['cbc'] + 'InvoiceTypeCode')
         if self.type == 'out_invoice':
+            type_code = etree.SubElement(
+                parent_node, ns['cbc'] + 'InvoiceTypeCode')
             type_code.text = '380'
         elif self.type == 'out_refund':
+            type_code = etree.SubElement(
+                parent_node, ns['cbc'] + 'CreditNoteTypeCode')
             type_code.text = '381'
         if self.comment:
             note = etree.SubElement(parent_node, ns['cbc'] + 'Note')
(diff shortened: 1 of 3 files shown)


account_invoice_ubl/models/account_invoice.py
```diff
@@ -111,14 +146,33 @@ def _ubl_add_legal_monetary_total(self, parent_node, ns, version='2.1'):
...
+        # ChargeAllowance node of type discount
+        dpo = self.env["decimal.precision"]
+        qty_precision = dpo.precision_get("Product Unit of Measure")
+        price_precision = dpo.precision_get("Product Price")
+        qty = iline.quantity
+        if float_is_zero(qty, precision_digits=qty_precision):
+            price_unit = 0.0
+        else:
+            price_unit = float_round(
+                iline.price_subtotal / float(qty),
```
thomaspaulb: Maybe it's a silly question but why not just get [iline.price_unit](https://github.com/OCA/OCB/blob/10.0/addons/account/models/account_invoice.py#L1241) directly here?

thomaspaulb:
@OCA/edi-maintainers Since the quality of the module is currently quite poor, and this PR is an obvious improvement, I'm inclined to just treat this module as an "alpha module" and advocate proactively merging improvements in and see where we end up.

What's your opinion on the matter?

account_invoice_ubl/models/account_invoice.py
jbaudoux: The line price_unit can be VAT included. Ideally, we should take the price_unit and remove the included VAT from any tax configured as such.
Taking the price_unit also requires to support any discount and declare it in another section.
Currently, it is the price after the discounts. It was already like that and I didn't want to change and add more complexity on this

rvalyi:
it will take me some time so, I cannot do it right now, but I would like to give you a perspective on this cause I have an long term alternative proposal. I'll try over the week end. Please ask me again if I forget. I don't want to block or slow down anything meanwhile.

account_invoice_ubl/models/account_invoice.py
jbaudoux: I finally implemented it properly. I compute the VAT excluded unit price and declare the discount in an additional separate section.

The previous implementation was causing rounding issues in some specific amount.

jbaudoux:
I fixed the credit note generation following the CreditNote schema instead of Invoice schema

PR #774, opened 2023-05-29
[14.0] edi: fix related record missing broken views

simahawk:
Before this change, when a related record was deleted you could not access anymore any exchange record (tree or form) due to the permission check.

Now:

* the check is bypassed
* related record buttons are hidden
* an informative message is displayed in the form


etobella (approved):
Makes sense to me

simahawk:
CC @florentx 

/ocabot merge patch

PR #776, opened 2023-05-30
[ADD][14.0] edi_webservice_oca_response 

petrus-v:
This module store webservice responses while using webservice in edi framework.

This can be usefull to understand what's wrong while some webservice return usefull informations within http error code.

### DIFF
--- a/edi_webservice_oca_response/components/send.py
+++ b/edi_webservice_oca_response/components/send.py
@@ -21,14 +21,15 @@ class EDIWebserviceSend(Component):
         except HTTPError as err:
             response_content = err.response.content
             status_code = err.response.status_code
-            http_error = err
+            raise err from err
+        except Exception as ex:
+            status_code = False
+            response_content = ""
+            raise ex from ex
         else:
             status_code = 200
-            http_error = None
         finally:
             self.exchange_record._set_file_content(
                 response_content, field_name="ws_response_content"
             )
             self.exchange_record.ws_response_status_code = status_code
-            if http_error:
-                raise http_error from http_error
--- a/edi_webservice_oca_response/components/send.py
+++ b/edi_webservice_oca_response/components/send.py
@@ -1,0 +1,33 @@
+# Copyright 2022 Foodles (http://www.foodles.co).
+# @author Pierre Verkest <pierreverkest84@gmail.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from requests import HTTPError
+
+from odoo.addons.component.core import Component
+
+
+class EDIWebserviceSend(Component):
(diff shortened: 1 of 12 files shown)


edi_webservice_oca_response/models/edi_exchange_record.py
```diff
@@ -0,0 +1,30 @@
...
+    ws_response_content = fields.Binary(
+        "Response", attachment=True, copy=False, help="Web service response content"
+    )
+    ws_response_content_filename = fields.Char(
+        compute="_compute_ws_response_content_filename"
+    )
+
+    def _compute_ws_response_content_filename(self):
+        for rec in self:
+            rec.ws_response_content_filename = _("response_") + rec.exchange_filename
```
simahawk: it's a tech name, it should not be translatable IMO

edi_webservice_oca_response/components/send.py
```diff
@@ -0,0 +1,34 @@
...
+
+from odoo.addons.component.core import Component
+
+
+class EDIWebserviceSend(Component):
+    _inherit = "edi.webservice.send"
+
+    def __init__(self, work_context):
+        super().__init__(work_context)
+        self.ws_settings = getattr(work_context, "webservice", {})
```
simahawk, line 15: this is already defined

simahawk (changes requested):
Hi @petrus-v , thanks for your contrib.

I'd prefer to see this in the `webservice` module.
We can have:
1. a flag on the webservice backend to enable this behavior
2. a `webservice.consumer.mixin`  to reuse where needed 
3. use this mixin in `edi_webservice` to extend edi.exchange.record

Food for thoughts: in the long term I'd like to make `rest_log` independent from `base_rest` and use it to track any kind of request or response.

petrus-v:
> Hi @petrus-v , thanks for your contrib.
> 
> I'd prefer to see this in the `webservice` module. We can have:
> 
>     1. a flag on the webservice backend to enable this behavior
> 
>     2. a `webservice.consumer.mixin`  to reuse where needed
> 
>     3. use this mixin in `edi_webservice` to extend edi.exchange.record

thanks for review and advices, I'll take care of them

pedrobaeza:
Shouldn't it be `edi_webservice_response_oca` or just `edi_webservice_response`?

petrus-v:
@pedrobaeza thanks for your review !

As asked by Simone I'll refactor and move all the code present here in `webservice` and `edi_webservice_oca` so I'll close this PR as soon the refactor is ready and would  like to keep in draft in the mean time (I should find time to do that in given weeks)

petrus-v:
It's time to close this PR in favor of https://github.com/OCA/web-api/pull/22 and https://github.com/OCA/edi/pull/787

PR #777, opened 2023-05-30
[14.0] account_invoice_import_simple_pdf: add a new type 'Any Character' on invoice number parsing

alexis-via:
Add a tests for this new type.


PierrickBrun (approved):
LGTM

PR #778, opened 2023-06-01
[14.0] edi: add disable flag in core mixin + exc type partner config

simahawk:

### DIFF
--- a/edi_oca/models/edi_exchange_consumer_mixin.py
+++ b/edi_oca/models/edi_exchange_consumer_mixin.py
@@ -44,6 +44,12 @@ class EDIExchangeConsumerMixin(models.AbstractModel):
         default={},
     )
     edi_has_form_config = fields.Boolean(compute="_compute_edi_config")
+    # TODO: rename to `edi_disable_auto`
+    disable_edi_auto = fields.Boolean(
+        string="Disable auto",
+        help="When marked, EDI automatic processing will be avoided",
+        # Each extending module should override `states` as/if needed.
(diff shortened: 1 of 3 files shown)


edi_oca/models/edi_exchange_consumer_mixin.py
```diff
@@ -44,6 +44,12 @@ class EDIExchangeConsumerMixin(models.AbstractModel):
         default={},
     )
     edi_has_form_config = fields.Boolean(compute="_compute_edi_config")
+    # TODO: rename to `edi_disable_auto`
+    disable_edi_auto = fields.Boolean(
```
simahawk, line 48: This is the old TODO I mentioned a long time ago. 
You can now reduce the override of this field only to the states.
CC @LoisRForgeFlow for edi_purchase and other modules.

simahawk: @LoisRForgeFlow @etobella ping

LoisRForgeFlow: Sorry @simahawk, were you expecting some action from our side? I thought it was more like a "FYI"

simahawk: @LoisRForgeFlow yeah, just wanted an ack and maybe a review on the rest :wink:

simahawk:
Tested on prod since a while, no objection here... let it go...

/ocabot merge minor

PR #759, opened 2023-05-11
[14.0] Add edi_sale_oca + edi_sale_ubl_oca

simahawk:
Provide full cycle for inbound Order + outbound OrderResponse.
These modules are now deprecated in favour of a more organic `edi_sale_oca` + `edi_sale_ubl_oca`:

- edi_sale_order_import
- edi_sale_order_import_ubl
- edi_sale_order_import_ubl_endpoint

Depends on:
- [ ] https://github.com/OCA/edi-framework/pull/4
- [x] https://github.com/OCA/edi-framework/pull/2
- [x] https://github.com/OCA/edi-framework/pull/3
- [x] #778
- [x] #797 

TODO:
- [ ] split PR by module
- [ ] document

PR #779, opened 2023-06-05
[14.0][ADD] sale_order_import_edifact & base_edifact

rmorant:
Plugs EDIFACT/D96A ORDER in sale_order_import process.


rmorant:
Hi @etobella, no tests yet.

rmorant:
Tests added in the OCA Days. I don't know if refactoring will be needed with the edi_sale_oca change (https://github.com/OCA/edi/pull/759), but I would mege it now and adapt it when necessary.
What do you think? ( Enric @etobella  )

simahawk:
@rmorant question: where is this module coming from? The history seems be incomplete...

rmorant:
> @rmorant question: where is this module coming from? The history seems be incomplete...
Hi Simone, 
I only found a PR in v12: https://github.com/PlanetaTIC/edi/tree/12.0-add-base_edifact
I migrated these module, but it did not parse the EDIFACT/D96A files correctly and I created a new one based on the pydifact library.
It has been running in production with Amazon orders for over 6 months, but may not be fully in tune with the latest changes of the framework.
I would like to help in this regard.

simahawk:
Ok, then this is not a migration but and addition in fact.
Could you please adapt the description accordingly?
Also, a rebase would be nice so that we can move on (with v16 too).

base_edifact/__manifest__.py
```diff
@@ -0,0 +1,29 @@
...
+    "license": "AGPL-3",
+    "application": False,
+    "installable": True,
+    # "preloadable": True,
+    "external_dependencies": {
+        "python": ["pydifact"],
+        "bin": [],
+    },
+    "depends": [
+        # "edi_party_data_oca"
```
simahawk, line 22:
```suggestion
```

base_edifact/models/edifact.py
```diff
@@ -0,0 +1,211 @@
...
+                # segms.append((segment.tag, segment.elements))
+                seg = dict()
+                seg[segment.tag] = segment.elements
+                segms.append(seg)
+            msg["segments"] = segms
+            obj.append(msg)
+        return obj
+
+    @api.model
+    def get_party_data(self, exchange_record, partner, raise_if_not_found=True):
```
simahawk, line 70: this must be dropped

sale_order_import_edifact/__manifest__.py
```diff
@@ -0,0 +1,31 @@
...
+    "external_dependencies": {
+        "python": [],
+        "bin": [],
+    },
+    "depends": [
+        "sale_management",
+        "partner_identification",
+        "partner_identification_gln",
+        "base_edifact",
+        "edi_sale_order_import",
```
simahawk, line 25: this module is not meant to plug into edi framework

simahawk (changes requested):
by looking at the current state is probably easier if we move on v16 first (where we are already applying fixes etc) and then backport changes here.
See https://github.com/OCA/edi/pull/813

rmorant:
Hi @simahawk, now that you have approved #821. I've been looking at how to backport the modules as you suggested above.
base_edifact does not need changes and i have created the PR: #847 
But backporting #821 (sale_order_import_edifact) require changes to the sale_order_import module as those in v16. 
What do you think is the best path?
Try to backporting sale_order_import too or add the changes applied in the migration to this one?

simahawk:
> Try to backporting sale_order_import too or add the changes applied in the migration to this one?

I'd go for this as long as we can do it w/o introducing breaking changes.

PR #780, opened 2023-06-07
[13.0][IMP] edi: relate exchange record with multiple records

JaumeBforgeFlow:
Introduces a new structure to be able to relate an exchange record to multiple records, while maintaining the old way of relating a record.


simahawk:
Hola @JaumeBforgeFlow, tnx for your interesting contrib :)
Early comments after a very fast check: can you please backport changes from v14 before introducing this? Otherwise fwd porting will be a hell. Also, maybe it's already in your plans but just in case: we'll need full test coverage for this.

PR #787, opened 2023-06-22
[14.0] edi_webservice_oca: save webservice responses on exchange records

petrus-v:
This is part of what has been discussed here: https://github.com/OCA/edi/pull/776#pullrequestreview-1462480583 and based on https://github.com/OCA/web-api/pull/22

In draft at the moment, I'd like to 
- [x] Improve and discuss about design: I felt some abstraction is missing... I'll be back on it later
- [x] add flag to be able to activate or deactivate this feature from the webservice backend and use the current behavior as default: do not save responses and do not catch HTTP ERRORS !


edi_webservice_oca/components/send.py
```diff
@@ -27,7 +29,25 @@ def __init__(self, work_context):
...
+            response_content = ""
+            raise ex from ex
+        else:
+            status_code = 200
+        finally:
+            self.exchange_record._set_file_content(
+                response_content, field_name="ws_response_content"
+            )
+            self.exchange_record.ws_response_status_code = status_code
+        return response_content
```
petrus-v: @simahawk at the moment I've move code from #776 here but I've the feeling this code should be part of the webservice module some how (by this https://github.com/OCA/web-api/pull/22/files).

I'm not sure how to make that nice what do you think of something likes ?

```suggestion
       return self.webservice_backend.call_tracking_response(method, *pargs, **kwargs, tracking_record=self.exchange_record)
```
but I'm not really happy with that solution too... any idea welcomes

petrus-v: I'll do in this way first !  it makes the code easier to read and will check record is an instance of consumer mixin.

Also I'll add the option on activate or not the feature

edi_webservice_oca/models/edi_backend.py
```diff
@@ -33,3 +35,10 @@ def _component_sort_key(self, component_class):
         return (
             1 if getattr(component_class, "_webservice_protocol", False) else 0,
         ) + res
+
+    def _swallable_exceptions(self):
+        exceptions = super()._swallable_exceptions()
+        self.ensure_one()
+        if self.webservice_backend_id.save_response:
```
simahawk: if you swallow it the job state will always be "done"... I'm not sure is desirable :thinking: 
See my comments on https://github.com/OCA/web-api/pull/22

edi_webservice_oca/models/edi_exchange_record.py
```diff
@@ -0,0 +1,21 @@
...
+    Define an exchange record.
+    """
+
+    _name = "edi.exchange.record"
+    _inherit = [
+        "edi.exchange.record",
+        "webservice.consumer.mixin",
+    ]
+
+    def _compute_ws_response_content_filename(self):
```
simahawk, line 19: I'm not sure this override is needed.... the file will always be attached to the current record and if you include the datetime (as mentiond in https://github.com/OCA/web-api/pull/22) we probably don't need this change.

simahawk (changes requested):

edi_webservice_oca/models/edi_backend.py
petrus-v: Yes job task is done and the EDI exchange is in `output_error_on_send` (as demonstrated by following test bellow `test_component_send_save_results_on_http_error`.

this is because the existing edi code: https://github.com/OCA/edi/blob/14.0/edi_oca/models/edi_backend.py#L297

Do you think the queue job state should be in error as well?

edi_webservice_oca/models/edi_exchange_record.py
petrus-v: I wanted to reuse efficiently all the business rules implemented here https://github.com/OCA/edi/blob/14.0/edi_oca/models/edi_exchange_type.py#L213

simahawk: ok, but the file name shouldn't be the same, especially if we include the datetime as mentiond on OCA/web-api#22

petrus-v: ok you are right ! I'll improve that

edi_webservice_oca/tests/test_send_saving_response.py
```diff
@@ -0,0 +1,114 @@
...
+            status=401,
+        )
+
+        with mock.patch(
+            "odoo.addons.webservice.models.webservice_backend."
+            "WebserviceBackend._consumer_record_env",
+            side_effect=self._consumer_record_no_new_env,
+        ):
+            self.backend._check_output_exchange_sync()
+        self.assertEqual(self.record.edi_exchange_state, "output_error_on_send")
```
petrus-v: Hum... since I've changed the behavior in `webserivce` to save data on the record in a different transaction, the state is not updated properly when HTTP error is in the _swallable_exceptions (Here the _consumer_record_env mock hide this issue as in the test there is only one transaction).

I've realized (thanks @simahawk ), this module can't take this decision, we have to let developers to choose at edi backend or exchange type level what's the expected behavior.

For instance in case of HTTP ERROR 500 we probably wants to retry the call and keep the edi exchange to  `output_pending`, in case of a web service returns some errors regarding the paylaods we probably want to move to `output_error_on_send`

So I'm going to stop `_swallable_exceptions` here and change `webservice` module in order to let developer change the state of the consumer exchange record.

edi_webservice_oca/tests/test_send_saving_response.py
```diff
@@ -0,0 +1,207 @@
...
+            url,
+            body=response,
+            status=401,
+        )
+        with mock.patch(
+            "odoo.addons.webservice.models.webservice_backend."
+            "WebserviceBackend._consumer_record_env",
+            side_effect=self._consumer_record_no_new_env,
+        ) as mo:
+            with self.assertRaisesRegex(HTTPError, "Unauthorized"):
```
petrus-v: Hi @simahawk,

I've rebase this PR and get new features :). So I found that [this PR](https://github.com/OCA/edi/pull/796/files#diff-6252f8654b549013124ad45c71930bd855b1ea7ad31ec323d546516e494ee7ccR354) add `OSError` as `retryable_exceptions`.

Are you aware that [`HTTPError` inherits from `OSError`](https://github.com/python/cpython/blob/3.12/Lib/urllib/error.py) so that all kind of HTTP Errors (3xx, 4xx, 5xx) are retried ? And is that a desired side effect ?

edi_webservice_oca/models/edi_backend.py
```diff
@@ -1,6 +1,7 @@
 # Copyright 2020 Creu Blanca
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
+
```
simahawk, line 4: nice change :stuck_out_tongue_winking_eye:

edi_webservice_oca/tests/test_send_saving_response.py
```diff
@@ -0,0 +1,208 @@
...
+        self.assertFalse(
+            self.record._get_file_content(field_name="ws_response_content")
+        )
+
+    def test_ws_response_content_filename(self):
+        self.record.exchange_filename = "test.json"
+        self.assertEqual(self.record.ws_response_content_filename, "response_test.json")
+
+
+class TestWebServiceSendHandledErrorChangingExchangeState(TestEDIWebserviceBase):
```
simahawk:
```suggestion
class TestSendCustomSaveResponse(TestEDIWebserviceBase):
```

simahawk (commented):
LG overall.
@etobella mind to have a look to these PRs?

PR #796, opened 2023-07-05
[14.0] edi: generate/send chain jobs + add identity exact match based on checksum

simahawk:
Depends on:

- https://github.com/OCA/queue/pull/594

Would be nice to use job chain for input too. Not a blocker tho.

### DIFF
--- a/edi_oca/utils.py
+++ b/edi_oca/utils.py
@@ -2,9 +2,15 @@
 # @author Simone Orsi <simahawk@gmail.com>
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
 
+import hashlib
+
 from odoo.addons.http_routing.models.ir_http import slugify
 
 
 def normalize_string(a_string, sep="_"):
     """Normalize given string, replace dashes with given separator."""
     return slugify(a_string).replace("-", sep)
+
+
+def get_checksum(filecontent):
+    return hashlib.md5(filecontent).hexdigest()
--- a/edi_oca/utils.py
+++ b/edi_oca/utils.py
@@ -5,6 +5,7 @@
 import hashlib
 
 from odoo.addons.http_routing.models.ir_http import slugify
+from odoo.addons.queue_job.job import identity_exact_hasher
 
 
 def normalize_string(a_string, sep="_"):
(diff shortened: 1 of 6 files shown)


simahawk:
Indenty key does not work fine w/ job deps. We'll need this https://github.com/OCA/queue/pull/546

simahawk:
@etobella @HviorForgeFlow if you could pass by... :pray:

edi_oca/utils.py
```diff
@@ -2,9 +2,25 @@
...
+
+def get_checksum(filecontent):
+    return hashlib.md5(filecontent).hexdigest()
+
+
+def exchange_record_job_identity_exact(job_):
+    hasher = identity_exact_hasher(job_)
+    # Include files checksum
+    hasher.update(
+        str(sorted(job_.recordset.mapped("exchange_filechecksum"))).encode("utf-8")
```
etobella, line 24: Why do yo need to add the file checksum? The function already contains ID of the exchange record and function :thinking:

etobella (approved):

etobella (commented):
Oops, I didn't want to set the approval, I just wanted to mark it :cry:

edi_oca/utils.py
simahawk: Because is the only to way to know if the file is the same.
If by error you try to schedule another job you won't get duplicated jobs unless the content of the file is the same.
In most cases, if you rely only on the cron to process files you'll have the filter done by the check methods that will skip records not in the rate state or not w/ the right values to be generated/sent/received/processed.
However, if you force those checks and try to schedule a job for the a record whereas the file didn't change you shouldn't be able to.

simahawk: It might come handy also to compare files w/o reading the whole file (eg: to check if you received the same file twice).

HviorForgeFlow (approved):
Code review :+1:

ping @MiquelRForgeFlow

simahawk:
@etobella good for you or you seen any possible regression on your side?

edi_oca/utils.py
etobella: What will happen if the file is not generated?

etobella:
I don't see any major regressions from my side. :thinking: 

I tested with other OCA reports and everything works fine

etobella:
@simahawk I migrated `edi_oca` yesterday on a Production database with this change and it took a long time due to this change :sob: . I have +100.000 exchange records (At the end, I had to make a `Manual migration`)

Just a question, could we change it to use `sha1` and not `md5` in order to make a migration script? (there is a checksum for sha1 for attachments, so we can recover data from there?) At the end, I solved the issue for me, but it might be a problem for other users.

WDYT?

PR #797, opened 2023-07-05
[14.0]  edi: refactor model rules w/ specific model

simahawk:
Long awaited improvement.
Allows to decouple rules by model giving much more flexibility.
Migration steps included.
Backward compat kept.


simahawk:
@etobella can you please have a look? I know you might have existing implementations for Spanish localization that rely on form btns and I'd like not to break them if possible :)

edi_oca/models/edi_exchange_consumer_mixin.py
```diff
@@ -52,34 +52,45 @@ def _compute_edi_config(self):
...
-            result[exchange_type.id] = self._edi_get_exchange_type_conf(exchange_type)
+            result[rule.id] = self._edi_get_exchange_type_rule_conf(rule)
         return result
 
     @api.model
-    def _edi_get_exchange_type_conf(self, exchange_type):
-        conf = {"form": {}}
-        if exchange_type.model_manual_btn:
-            conf.update({"form": {"btn": {"label": exchange_type.name}}})
+    def _edi_get_exchange_type_rule_conf(self, rule):
```
simahawk, line 87: Dropping a method here: risk of breaking existing things but chances that someone overridden this are quite low...

etobella:
I did not override that function, but the change might break installation, as the fields might be used on data :thinking: 
I know you added a migration script, but might be a problem...

https://github.com/OCA/l10n-spain/blob/14.0/l10n_es_facturae_face/data/edi.xml#L16-L21

simahawk:
> I did not override that function, but the change might break installation, as the fields might be used on data thinking I know you added a migration script, but might be a problem...
> 
> https://github.com/OCA/l10n-spain/blob/14.0/l10n_es_facturae_face/data/edi.xml#L16-L21

I see. What would you suggest? Can't we update those files? Shouldn't be that hard.

Other options:

1. full backward compat (eg: keep the fields and rely on the rule to lookup for conf on the type + hide fields on type form via specific group)
2. leave fields, hide and ignore them (they will get migrated anyway) + log deprecation

I would like to avoid both of them honestly :stuck_out_tongue_winking_eye:

etobella:
My suggestion would be to leave them for 14, 15 and 16, but make them non storable and store the data properly on the new table. We need to make an inverse and compute in order to fill the data and fields properly :thinking: 

I know it is much harder, but it is probably cleaner,  WDYT?

simahawk:
> My suggestion would be to leave them for 14, 15 and 16, but make them non storable and store the data properly on the new table. We need to make an inverse and compute in order to fill the data and fields properly thinking
> 
> I know it is much harder, but it is probably cleaner, WDYT?

Ok for 14 and 15 but definitely not for 16 (which is not even ready yet).

I'm not in favor of keeping full support for the old style tho:

1. hide fields, to not break those files
2. deprecate it, to force users to switch (maybe even w/ a nice warning on the form)
3. add an inverse field only to reflect updates if necessary if you keep using the outdated version in such files

Sounds good?
Otherwise let's make a call (like  in the old days) :wink:

simahawk:
@etobella ping :)

etobella:
Good for me @simahawk If it is not merged on 16, we can remove the old way there :wink:

simahawk:
@etobella backward compat layer added. Tests included. Could you please test on your side when you have time?

image

simahawk:
Small improvement: a btn to wipe fields carefully

image

OriolMForgeFlow (commented):
Hi @simahawk ,
I am encountering an error during testing in runboat, as I am unable to create Exchange Type Rules, since the field *type_id* is required but is not displayed within the form view.

simahawk:
> Hi @simahawk , I am encountering an error during testing in runboat, as I am unable to create Exchange Type Rules, since the field _type_id_ is required but is not displayed within the form view.

Sorry, I missed a commit. Pushed now

simahawk:
@etobella could you drop a review here pls?

etobella:
I will test it this weekend 😁

johnny-longneck:
Hey Simone, I said I will also look at it this weekend but at the moment I will leave it to Etobella,

I was checking runboat and used the existing form button rule for sales orders and so far I am not seeing the button.

So, I was working mostly on the output direction for EDI and I am struggling to understand how to test it manually.
Following my ticket with "Documentation required" I would have liked some update on the readme file, with a section how to setup the new decoupled rules.

But as far as I understand it, I think the change is really good.
For my case, I mostly just have a model in the exchange type rules and let the model listener do the rest so my knowledge ends here.

edi_oca/migrations/14.0.1.20.0/post-migrate.py
```diff
@@ -0,0 +1,35 @@
...
+        SELECT * FROM exc_type_model_rel_bkp
+    """
+    cr.execute(query)
+    res = cr.dictfetchall()
+    model = env["edi.exchange.type.rule"]
+    for item in res:
+        kind = "form_btn" if item.pop("form_btn", False) else "custom"
+        vals = dict(item, name="Default", kind=kind)
+        model.create(vals)
+        item.type_id.button_wipe_deprecated_rule_fields()
```
etobella: This is failing.

Maybe you ment:

```python
        model.create(vals).type_id.button_wipe_deprecated_rule_fields()
```

etobella (approved):
Tested locally against l10n-spain tests and all worked properly except for the comment.

Good for me then :smile:

edi_oca/migrations/14.0.1.20.0/post-migrate.py
simahawk: ahhhhhhh you are totally write, I think I've just messed up after testing w/ a typo. 
Good catch! Fix coming!

simahawk: fixed

simahawk:
> Hey Simone, I said I will also look at it this weekend but at the moment I will leave it to Etobella,
> 
> I was checking runboat and used the existing form button rule for sales orders and so far I am not seeing the button.
> 
> So, I was working mostly on the output direction for EDI and I am struggling to understand how to test it manually. Following my ticket with "Documentation required" I would have liked some update on the readme file, with a section how to setup the new decoupled rules.
> 
> But as far as I understand it, I think the change is really good. For my case, I mostly just have a model in the exchange type rules and let the model listener do the rest so my knowledge ends here.

You are right, thanks for your feedback. 
I'm adding a new section in the configuration part.

edi_oca/readme/CONFIGURE.rst
```diff
@@ -27,3 +27,24 @@ snippet of code.
...
+Go to an exchange type and go to the tab "Model rules".
+There you can add one or more rule, one per model.
+On each rule you can define a domain or a snippet to activate it.
+In case of a "Form button" kind, if the domain and/ the snippet is/are satisfied,
+a form btn will appear on the top of the form.
+This button can be used by the end user to manually generate an exchange.
+If there's more than a backend and the exchange type has not a backend set,
+a wizard will appear asking to select a backend to be used for the exchange.
+
+In case of "Custom" kind, you'll have to define your own logic to do something.
```
john-herholz-dt, line 50: Much clearer now, thanks!

Issue #789, opened 2023-06-23
14.0 edi_oca synchronize ACK state with parent state when parent isn't processed by itself


simgonzalez:
**Is your feature request related to a problem?**
When I have an EDI link with an ACK I would except the ACK state to also update the parent exchange state if no processing behaviors is implemented for the parent exchange.

**Describe the solution you'd like**
The process state of the ACK exchange to define the process state of the parent exchange. 

**Describe alternatives you've considered**
To have a checkbox to activate or not that behavior.

**Additional context**
This could also be linked to update the state of the linked model. (That would need to implement a generic next_state method for all odoo model)

ecino:
Can someone check this, as well as the PR that solves the issue?

simahawk:
What do you mean exactly w/ "if no processing behaviors is implemented for the parent exchange"?
Can you give more details on your real use case?

simgonzalez:
@simahawk hello,
I mean that in the EDI process (from what I understood, as there’s no documentation) the oca modules give a behavior until a standard done state but if we want the record to appear done and process (which seems to be the real final state) we have to implement a mixin to take care of our EDI Exchange record for our exchange type. Then what I mean is that when I have an ACK I except the ACK to take care of changing the final state for my original record. Except in the case where an implementation is done to process the original EDI exchange record.

The real use case concerns a bank SFTP where we upload file generated from payment orders those file are automatically uploaded to the SFTP and we except an ACK that will tell us if the bank successfully received the file.

available for any other details

simahawk:
I see. Thanks for you explanation! I'll review and comment on the PR.

PR #800, opened 2023-07-18
[14.0][IMP] edi_oca:  sync parent state on ack process

simgonzalez:
Mechanism to sync both EDI exchange together.
resolve (issue #789 )

### DIFF
--- a/edi_oca/models/edi_backend.py
+++ b/edi_oca/models/edi_backend.py
@@ -489,10 +489,12 @@ class EDIBackend(models.Model):
                 }
             )
             if origin_output_state:
-                exchange_record.parent_id.write({
-                    "edi_exchange_state": origin_output_state,
-                    "exchange_error": error,
-                })
+                exchange_record.parent_id.write(
+                    {
+                        "edi_exchange_state": origin_output_state,
+                        "exchange_error": error,
+                }
+                )
             if state == "input_processed_error":
                 exchange_record._notify_error("process_ko")
             elif state == "input_processed":
@@ -630,9 +632,9 @@ class EDIBackend(models.Model):
         self, exchange_type, extra_domain=None, count_only=False
     ):
         domain = [
-                     ("backend_id", "=", self.id),
-                     ("type_id", "=", exchange_type.id),
-                 ] + extra_domain or []
+             ("backend_id", "=", self.id),
+             ("type_id", "=", exchange_type.id),
+        ] + extra_domain or []
         return self.env["edi.exchange.record"].search(domain, count=count_only)
 
     def action_view_exchanges(self):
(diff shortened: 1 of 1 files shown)


gbrito (approved):

ecino (approved):

simahawk (changes requested):
I'm not sure to understand the use case.
I'll comment on the linked issue.

edi_oca/models/edi_backend.py
```diff
@@ -445,6 +445,10 @@ def _exchange_process_check(self, exchange_record):
             "input_processed_error",
         ]
 
+    def _update_related_ack(self, exchange_record):
```
simahawk, line 448:
```suggestion
    def _is_parent_update_needed(self, exchange_record):
```

edi_oca/models/edi_backend.py
```diff
@@ -445,6 +445,10 @@ def _exchange_process_check(self, exchange_record):
...
+        if exchange_record.parent_id and exchange_record.type_id.ack_for_type_ids:
```
simahawk, line 449: I'm not sure the check on `ack_for_type_ids` is significant. What about this?
```suggestion
        if exchange_record.parent_id.ack_expected:
```

edi_oca/models/edi_backend.py
```diff
@@ -456,6 +460,7 @@ def exchange_process(self, exchange_record):
         state = exchange_record.edi_exchange_state
         error = False
         message = None
+        origin_output_state = False
```
simahawk, line 463:
```suggestion
        parent_output_state = None
```

edi_oca/models/edi_backend.py
```diff
@@ -464,20 +469,32 @@ def exchange_process(self, exchange_record):
...
             res = True
+            if self._update_related_ack(exchange_record):
+                origin_output_state = "output_sent_and_processed"
         finally:
             exchange_record.write(
                 {
                     "edi_exchange_state": state,
                     "exchange_error": error,
                     # FIXME: this should come from _compute_exchanged_on
+                    # TODO: shouldn't this update processed_on ?
```
simahawk, line 486: `exchanged_on` is a generic field. In this case, it reflects the time of processing. To see when it was "received" just check the create date.

simahawk (changes requested):
Thanks for your proposal!
Beside the comments, here some more steps:
* tests
* squash commits
* rewrite commit msg. Something like this should work:
 
```
edi: sync parent state on ack process

$detailed_explanation_as_given on the linked issue
```

ecino:
I will probably make a PR myself to address the comments.

PR #805, opened 2023-08-17
[13.0][FIX] Fixed dependecy name on __manifest__.py

10537:
The right name is base_sparse_field_list_support because was rewrite in commit  a28bb05 in server-tools OCA's repo.


rvalyi (dismissed):

simahawk (changes requested):
https://github.com/odoo/odoo/tree/13.0/addons/base_sparse_field

PR #809, opened 2023-09-05
[FIX] edi_webservice_oca: add internal user webservice backend read a…

petrus-v:
…ccess

queue.job task are running in the context with the user that create the edi exchage record as those user are able to create exchange they should be able to read webservice backend in order to etablish the connexion to send payloads to the related webserivce

### DIFF
--- a/edi_webservice_oca/tests/test_send.py
+++ b/edi_webservice_oca/tests/test_send.py
@@ -40,6 +40,24 @@ class TestSend(TestEDIWebserviceBase):
                     endpoint: push/here
         """
         cls.record.type_id.set_settings(cls.settings1)
+        cls.a_user = (
+            cls.env["res.users"]
+            .with_context(no_reset_password=True)
+            .create(
+                {
+                    "name": "foo",
+                    "login": "a_user",
+                    "email": "foo@bar.com",
+                    "groups_id": [
+                        (
+                            6,
+                            0,
+                            (cls.env.ref("base.group_user")).ids,
+                        )
+                    ],
+                }
+            )
+        )
 
     def test_find_component(self):
         component = self.backend._get_component(self.record, "send")
@@ -79,9 +97,15 @@ class TestSend(TestEDIWebserviceBase):
     @responses.activate
     def test_component_send(self):
         self.record.type_id.set_settings(self.settings2)
+        # Internal user should be able to call the third party webservice
+        # without read access (no ir.access.model records)
+        # on `webservice.backend` model which store credentials
+        record = self.record.with_user(self.a_user)
+        backend = self.backend.with_user(self.a_user)
+
         url = "https://foo.test/push/here"
         responses.add(responses.POST, url, body="{}")
-        component = self.backend._get_component(self.record, "send")
+        component = backend._get_component(record, "send")
         result = component.send()
         self.assertEqual(result, b"{}")
         self.assertEqual(
--- a/edi_webservice_oca/models/edi_backend.py
+++ b/edi_webservice_oca/models/edi_backend.py
@@ -5,7 +5,6 @@ from odoo import fields, models
 
 
 class EdiBackend(models.Model):
-
     _inherit = "edi.backend"
 
     webservice_backend_id = fields.Many2one("webservice.backend")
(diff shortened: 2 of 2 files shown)


edi_webservice_oca/security/ir.model.access.csv
```diff
@@ -1,2 +1,3 @@
 id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
 access_webservice_backend_edi_manager,webservice_backend edi manager,webservice.model_webservice_backend,base_edi.group_edi_manager,1,1,1,1
+access_webservice_backend_user,webservice_backend user,webservice.model_webservice_backend,base.group_user,1,0,0,0
```
petrus-v: I've realized that this would gives read access to credential, it's probably better to use sudo instead, I'll changed that

simahawk (commented):
TBH I don't like this. It makes too easier to spoil sensitive data.
I'd rather use `sudo` in in the adapters only to read what's needed.

simahawk (changes requested):
TBH I don't like this. It makes too easy to spoil sensitive data.
I'd rather use sudo in webservice adapters only to read what's needed.

edi_webservice_oca/security/ir.model.access.csv
petrus-v: done !

edi_webservice_oca/components/send.py
```diff
@@ -23,7 +23,7 @@ class EDIWebserviceSend(Component):
     def __init__(self, work_context):
         super().__init__(work_context)
         self.ws_settings = getattr(work_context, "webservice", {})
-        self.webservice_backend = self.backend.webservice_backend_id
+        self.webservice_backend = self.backend.webservice_backend_id.sudo()
```
simahawk: If this is to make `call` happy, IMO this should be done in "call" in the base module otherwise we'll have to address the same problem for all depending modules, no?

edi_webservice_oca/tests/test_send.py
```diff
@@ -40,6 +40,20 @@ def _setup_records(cls):
                     endpoint: push/here
         """
         cls.record.type_id.set_settings(cls.settings1)
+        cls.a_user = cls.env["res.users"].create(
```
simahawk:
```suggestion
        cls.a_user = cls.env["res.users"].with_context(no_reset_password=True).create(
```
or use `odoo.tests.common.new_test_user`

simahawk (changes requested):

edi_webservice_oca/components/send.py
petrus-v: double checked... you're right this one is not necessary

petrus-v:
```suggestion
        self.webservice_backend = self.backend.webservice_backend_id
```

damdam-s (approved):

edi_webservice_oca/tests/test_send.py
```diff
@@ -88,3 +106,19 @@ def test_component_send(self):
             responses.calls[0].request.headers["Content-Type"], "application/xml"
         )
         self.assertEqual(responses.calls[0].request.body, "This is a simple file")
+
+    @responses.activate
+    def test_compenent_send_by_end_user(self):
```
simahawk: this test could be skipped: use the user in the test above. If the avg user is able to run that, admins can do it for sure. Unless you think is better to test both :wink:

simahawk (commented):
One last change if you don't mind... Plus, please squash when done and we are ready to merge :+1:

edi_webservice_oca/tests/test_send.py
petrus-v: done !

simahawk (approved):

petrus-v:
@simahawk I've just rebase this branch, every things looks okay ! not sure why oca-git-bot failed at some point (I've deep into)

petrus-v:
I've just rebase this branch, all CI seems ok, @simahawk could you re-try to merge this one ?

PR #813, opened 2023-09-12
[16.0][MIG] base_edifact: Migration to 16.0

QuocDuong1306:
Start migrating from:
- https://github.com/OCA/edi/pull/779


base_edifact/__manifest__.py
```diff
@@ -0,0 +1,29 @@
...
+    "license": "AGPL-3",
+    "application": False,
+    "installable": True,
+    # "preloadable": True,
+    "external_dependencies": {
+        "python": ["pydifact"],
+        "bin": [],
+    },
+    "depends": [
+        "edi_party_data_oca",
```
simahawk: hmm I don't think we want edi framework to be used here

base_edifact/__manifest__.py
```diff
@@ -0,0 +1,29 @@
...
+    "license": "AGPL-3",
+    "application": False,
+    "installable": True,
+    # "preloadable": True,
+    "external_dependencies": {
+        "python": ["pydifact"],
+        "bin": [],
+    },
+    "depends": [
+        # "edi_party_data_oca"
```
simahawk: this change should not belong to this commit

base_edifact/tests/test_base_edifact.py
```diff
@@ -0,0 +1,69 @@
...
+    def setUpClass(cls):
+        super().setUpClass()
+
+    @classmethod
+    def tearDownClass(cls):
+        super().tearDownClass()
+
+    def setUp(self):
+        super().setUp()
+        self.base_edifact_model = self.env["base.edifact"]
```
simahawk: this should go to setUpClass

base_edifact/tests/test_base_edifact.py
```diff
@@ -0,0 +1,69 @@
...
+        return fd.read()
+
+
+class TestBaseEdifact(TransactionCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+
+    @classmethod
+    def tearDownClass(cls):
```
simahawk: useless overrides

simahawk (changes requested):
Would be nice to see some more test cov

QuocDuong1306:
Thank @simahawk , I updated

base_edifact/models/edifact.py
```diff
@@ -0,0 +1,202 @@
+import datetime
```
simahawk, line 5: (c) missing

simahawk (approved):
LG overall

base_edifact/models/edifact.py
simahawk: @QuocDuong1306 can you fix this pls?  Then we are ready to go.

QuocDuong1306: Done here, thank @simahawk

PR #817, opened 2023-09-26
[ADD] punchout modules

benwillig:
Theses modules suite target to use the [punchout technology](https://punchoutcommerce.com/guides/punchout/) in order to connect Odoo to a supplier ecommerce website.
This is only an alpha version and simply handle the base connection and the reception of the response. We created submodules to handle the creation of the purchase orders, UOMs and products but they are not mature enough to be published.

User is redirected on the ecommerce website, then he can manage a shopping cart which can be sent back to Odoo. The ecommerce website trigger the form submission using javascript directly into the browser so the user is still connected.

This implementation has been done based on the [Wurth eshop](https://eshop.wurth.be).

cc @ThomasBinsfeld @sbidoul


chienandalu (commented):
Hi @benwillig and thanks for your contribution! 

@hbrunn started working in this same approach (Odoo does the punchout to the provider and comes back from it with a response to make the purchase.order) in https://github.com/OCA/edi/pull/720 It isn't using cXML though.

Also @gurneyalex [commented](https://github.com/OCA/edi/pull/720#issuecomment-1497192395) that they're developing another line of work to do the punchout the other way around: Odoo receives a punchout request and the customer uses the eCommerce as usual and finally at checkout the order is returned to the punchout requester to be recorded in their system. Although AFAIK there's no public code yet.

We at Tecnativa are working in this approach as well (quite focused in Ariba Network although we could go with a more broad approach) although is quite immature at the moment (our colleague for this task is on paternity leave :slightly_smiling_face: ): https://github.com/Tecnativa/e-commerce/tree/13.0-add-website_sale_punchout/website_sale_punchout We're using cXML requests as well, although our approach is to use qweb templating to render them avoiding struggling with lxml.

So I think that we should try and join efforts to have at least commonground that's useful for everyone and avoid implenting the same multiple times (well, that's OCA's deal after all :smile: )

benwillig:
Thank you for your feedback, indeed the qweb rendering might be a better option. It would be interesting to see and compare how our provider are working with the XML and how it is populated. I will work a bit on this during the OCA days so I'm open to talk about this.

chienandalu:
Hi @benwillig !

I think a good commonground could be to implement a `base_cxml` module wich would have:

- A set of cXML requests as documented in https://xml.cxml.org/current/cXMLReferenceGuide.pdf in the form of qweb templates. This way, we could contribute in both directions. An example of some of the request we already implemented: https://github.com/Tecnativa/e-commerce/blob/82d63815b5e85d4a9beb37904ebed7b93e6d2cff/website_sale_punchout/views/punchout_index_catalog_templates.xml#L1
- A cxml parser for the punchout server responses (I think you already started your work in that direction).

What we're struggling the most with is the SAP Ariba test environments which we couldn't get to make functional, so we're working in a theorethical base...

benwillig:
I didn't have time to finish what I was working on last week, but to resume what I'm planning to do:

- handle the base64 cxml if it exists
- validate the DTD of the response (uploaded on each punchout.backend)
- assign the operation to the punchout.request, each operation is associated to a CXML template (a ir.ui.view) which can be rendered to be sent
- allow to retrieve as a dict the cXml response to be used easily in a submodule

I will try to work on this regularly.

bosd:
> We at Tecnativa are working in this approach as well (quite focused in Ariba Network although we could go with a more broad approach) although is quite immature at the moment (our colleague for this task is on paternity leave 🙂 ): https://github.com/Tecnativa/e-commerce/tree/13.0-add-website_sale_punchout/website_sale_punchout We're using cXML requests as well, although our approach is to use qweb templating to render them avoiding struggling with lxml.

@chienandalu Are you stil working on this from the odoo webshop side?

Issue #818, opened 2023-09-26
Build red on 14.0 and 15.0 branch


simahawk:
```
 2023-09-22 11:11:08,170 1786 ERROR odoo odoo.addons.edi_oca.tests.test_security: FAIL: TestEDIExchangeRecordSecurity.test_rule_no_create
odoo.exceptions.AccessError: Due to security restrictions, you are not allowed to modify 'Model used only for test' (edi.exchange.consumer.test) records.

Contact your administrator to request access if necessary.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/opt/odoo/odoo/tools/misc.py", line 885, in deco
    return func(*args, **kwargs)
  File "/__w/edi/edi/edi_oca/tests/test_security.py", line 90, in test_rule_no_create
    self.create_record(self.user)
AssertionError: "Exchange Record rule demo" does not match "Due to security restrictions, you are not allowed to modify 'Model used only for test' (edi.exchange.consumer.test) records.

```
The same test was passing before. Something has probably changed somewhere else.

simahawk:
I have no time this week. @etobella if you have time to check, be my guest :wink:

PR #814, opened 2023-09-15
[15.0] FWD port of multiple PRs

etobella:
Small list of PRs pending to port to 15.0 :wink: 

https://github.com/OCA/edi/pull/639
https://github.com/OCA/edi/pull/651
https://github.com/OCA/edi/pull/652
https://github.com/OCA/edi/pull/653
https://github.com/OCA/edi/pull/645
https://github.com/OCA/edi/pull/648
https://github.com/OCA/edi/pull/662
https://github.com/OCA/edi/pull/666
https://github.com/OCA/edi/pull/672
https://github.com/OCA/edi/pull/694
https://github.com/OCA/edi/pull/696
https://github.com/OCA/edi/pull/697
https://github.com/OCA/edi/pull/698
https://github.com/OCA/edi/pull/721
https://github.com/OCA/edi/pull/762
https://github.com/OCA/edi/pull/764
https://github.com/OCA/edi/pull/763
https://github.com/OCA/edi/pull/765
https://github.com/OCA/edi/pull/767
https://github.com/OCA/edi/pull/774
https://github.com/OCA/edi/pull/778
https://github.com/OCA/edi/pull/797
https://github.com/OCA/edi/pull/790
https://github.com/OCA/edi/pull/808

PR #819, opened 2023-09-26
[FIX] edi_oca: Adapt tests to changes on error messages

etobella:
@simahawk 
Solves #818
I had to apply the same patch on #814

Might be related to https://github.com/odoo/odoo/pull/135028 ?

### DIFF
--- a/edi_oca/tests/test_security.py
+++ b/edi_oca/tests/test_security.py
@@ -86,18 +86,18 @@ class TestEDIExchangeRecordSecurity(EDIBackendCommonTestCase):
     def test_rule_no_create(self):
         self.user.write({"groups_id": [(4, self.group.id)]})
         self.consumer_record.name = "no_rule"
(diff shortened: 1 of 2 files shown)


etobella:
@simahawk I tried to do this without removing the regex, but odoo is now using has_user_group, this checks if the call comes from a request and we are on debug mode. Obviously it is not the case on tests, so the result will never be accessible :sob:

bosd (approved):
Functional review. Tests are passing again. :+1

etobella:
/ocabot merge patch

Merging as it is raising errors on current branch

simahawk:
> has_user_group

you can still check the msg in the exception in theory. I'll have a look as I don't like we are not testing the right thing

simahawk:
FTR fixed like this on v16 https://github.com/OCA/edi-framework/pull/5/commits/f1e6976f1bc510bbb014be53c343ebcf0f80a88f

PR #816, opened 2023-09-26
[16.0][MIG] sale_order_import: Migration to 16.0

QuocDuong1306:

PR #821, opened 2023-09-28
[16.0][MIG] sale_order_import_edifact: Migration to 16.0

QuocDuong1306:
Depends on:
- https://github.com/OCA/edi/pull/813
- https://github.com/OCA/edi/pull/816


nilshamerlinck:
Note: 
- `sale_order_import_edifact` is now just bringing EDIFACT support to `sale_order_import`, disconnected from the EDI framework
- it's `edi_sale_edifact_oca`, similarily to `edi_sale_ubl_oca`, that will glue `sale_order_import_edifact` with the EDI framework from now on

sale_order_import_edifact/wizard/sale_order_import.py
```diff
@@ -0,0 +1,237 @@
...
+        if self.order_filename:
+            path, ext = os.path.splitext(self.order_filename)
+            ok = ext and ext[1:] in extensions
+            if not ok and self.order_file:
+                ok = b64decode(self.order_file[:4]) == b"UNB"
+        return ok
+
+    @api.onchange("order_file")
+    def order_file_change(self):
+        if self.edifact_ok:
```
simahawk: see my comments on https://github.com/OCA/edi/pull/816

simahawk (changes requested):

QuocDuong1306:
Hi @simahawk , I updated

sale_order_import_edifact/readme/ROADMAP.rst
```diff
@@ -0,0 +1,5 @@
+..  Enumerate known caveats and future potential improvements.
+    It is mostly intended for end-users, and can also help
+    potential new contributors discovering new features to implement.
+
+* Refactor import by release
```
simahawk: can you elaborate on this?

sale_order_import_edifact/tests/test_edifact_order_import.py
```diff
@@ -0,0 +1,41 @@
...
+import base64
+
+from odoo.tests.common import TransactionCase
+from odoo.tools import mute_logger
+
+from .common import get_test_data
+
+
+class TestEdifactOrderImport(TransactionCase):
+    @mute_logger("odoo.addons.sale_order_import.wizard.sale_order_import")
```
simahawk, line 18: can you set tracking_disable on setUpClass pls?

sale_order_import_edifact/wizard/sale_order_import.py
```diff
@@ -0,0 +1,236 @@
...
+
+from odoo import _, api, fields, models
+from odoo.exceptions import UserError
+
+logger = logging.getLogger(__name__)
+
+
+class SaleOrderImportEdifact(models.TransientModel):
+    _name = "sale.order.import"
+    _inherit = ["sale.order.import", "base.edifact"]
```
simahawk: pls avoid extending base.edifact. That's a tool, not a mixin and should be used directly where needed.
We already removed base.business.document.import inheritance a long time ago ;)

sale_order_import_edifact/wizard/sale_order_import.py
simahawk: I don't understand this. We should leverage the call to super for loading defaults.
Also, here, the doc_type and the price_source are kind of hardcoded...
Shouldn't we rely on the parsing of the file as for the other formats?

sale_order_import_edifact/wizard/sale_order_import.py
```diff
@@ -0,0 +1,236 @@
...
+    def order_file_change(self):
+        if self.edifact_ok:
+            self.doc_type = self.env.context.get("doc_type", "rfq")
+            self.price_source = "order"
+        else:
+            res = super(SaleOrderImportEdifact, self).order_file_change()
+            if isinstance(res, dict):
+                return res
+
+    def parse_order_button(self):
```
simahawk: hmm this seems to be an additional method for the btn that is not defined by the base module..

sale_order_import_edifact/wizard/sale_order_import.py
```diff
@@ -0,0 +1,236 @@
...
+        return action
+
+    def _get_supported_types(self):
+        # Add more types for EDIFACT
+        res = super()._get_supported_types()
+        res.update({"edifact": ("text/plain")})
+        return res
+
+    @api.model
+    def parse_edifact_order(self, filecontent, detect_doc_type):
```
simahawk: Signature to respect
```suggestion
    def parse_edifact_order(self, filecontent, detect_doc_type=detect_doc_type):
```

sale_order_import_edifact/wizard/sale_order_import.py
```diff
@@ -0,0 +1,236 @@
...
+        # https://github.com/nerdocs/pydifact/blob/master/pydifact/segmentcollection.py
+        if not self.edifact_ok:
+            return None
+        interchange = self._loads_edifact(filecontent)
```
simahawk: Getting back to the onchange remark above, if you want to avoid heavy parsing here while only validating the doc type and stating if the file is good or not, you could pass a ctx key from the onchange like `validate_only` or similar, to just do a check.

sale_order_import_edifact/wizard/sale_order_import.py
```diff
@@ -0,0 +1,236 @@
...
+    import_type = fields.Selection(
+        selection_add=[("edifact", "EDIFACT")], ondelete={"edifact": "cascade"}
+    )
+    state = fields.Selection(
+        selection_add=[("tech", "Technical")], ondelete={"tech": "set default"}
+    )
+    parsed_json = fields.Text()
+
+    @property
+    def edifact_ok(self):
```
simahawk: this should be a computed field. It can be used in the UI and is going to be cached if you used it in other places like the import machinery below

sale_order_import_edifact/wizard/sale_order_import.py
```diff
@@ -0,0 +1,236 @@
...
+                ok = b64decode(self.order_file[:4]) == b"UNB"
+        return ok
+
+    @api.onchange("order_file")
+    def order_file_change(self):
+        if self.edifact_ok:
+            self.doc_type = self.env.context.get("doc_type", "rfq")
+            self.price_source = "order"
+        else:
+            res = super(SaleOrderImportEdifact, self).order_file_change()
```
simahawk:
```suggestion
            res = super().order_file_change()
```

test-requirements.txt
```diff
@@ -0,0 +1,4 @@
+odoo_test_helper
+odoo-addon-edi_party_data_oca @ git+https://github.com/OCA/edi.git@refs/pull/812/head#subdirectory=setup/edi_party_data_oca
```
simahawk: edi_party_data_oca is not a dep

simahawk (changes requested):

sale_order_import_edifact/readme/ROADMAP.rst
QuocDuong1306: This was an old item that was in the old version, so may I remove it?

sale_order_import_edifact/wizard/sale_order_import.py
QuocDuong1306: This button was added by the original author of the module in v14. So do you want me to remove this button?

QuocDuong1306:
I updated, but waiting for your reply to the comments above:
>This was an old item that was in the old version, so may I remove it?
>This button was added by the original author of the module in v14. So do you want me to remove this button?

After that, I will force push again

sale_order_import_edifact/wizard/sale_order_import.py
rmorant: This is to be able to force from configuration.

```yaml
sale_order_import:
      price_source: order
      confirm_order: false
      wiz_ctx:
         file_ext: 'edi'
         release: 'd96a'
         doc_type: 'rfq'
```
but yes, `price_source` should be taken from context like `doc_type`

sale_order_import_edifact/wizard/sale_order_import.py
rmorant: It's just for debugging (groups="base.group_system,base.group_no_one"). To take a quick look at the pydifact structure ('edifact' field) and the parsed object ('order') from the original document before importing the document.

It was very useful during development and I left it to detect changes in the future

simahawk: That's fine. Then let's make it clear: can't we name it like `button_parse_order_preview` or something similar?
Also a docstring explaining what has been described above would be nice :wink:

sale_order_import_edifact/wizard/sale_order_import.py
```diff
@@ -0,0 +1,248 @@
...
+        selection_add=[("tech", "Technical")], ondelete={"tech": "set default"}
+    )
+    parsed_json = fields.Text()
+    edifact_ok = fields.Boolean(compute="_compute_edifact_ok", store=True)
+
+    @api.depends("order_file", "order_filename")
+    def _compute_edifact_ok(self):
+        for rec in self:
+            if rec.order_file and rec.order_filename:
+                conf_ext = self.env.context.get("file_ext", "txt,d96a")
```
simahawk: Instead of depending purely on the ctx, I think it's better to have a field and via the ctx we set its default.

sale_order_import_edifact/wizard/sale_order_import.py
```diff
@@ -0,0 +1,248 @@
...
+    state = fields.Selection(
+        selection_add=[("tech", "Technical")], ondelete={"tech": "set default"}
+    )
+    parsed_json = fields.Text()
+    edifact_ok = fields.Boolean(compute="_compute_edifact_ok", store=True)
+
+    @api.depends("order_file", "order_filename")
+    def _compute_edifact_ok(self):
+        for rec in self:
+            if rec.order_file and rec.order_filename:
```
simahawk:
```suggestion
            if not (rec.order_file and rec.order_filename):
                rec.edifact_ok = False
                continue
```
then you can indent back the rest of the method

sale_order_import_edifact/wizard/sale_order_import.py
```diff
@@ -0,0 +1,248 @@
...
+                    ok = b64decode(self.order_file[:4]) == b"UNB"
+                rec.edifact_ok = ok
+            else:
+                rec.edifact_ok = False
+
+    @api.onchange("order_file")
+    def order_file_change(self):
+        # Support get context if using EDI machinery
+        res = super().order_file_change()
+        self.price_source = self.env.context.get("price_source")
```
simahawk: Again, in the edi machinery this can be supported w/ standard ctx key `default_price_source` instead of having another specific key (I'd like to cleanup this approach everywhere else indeed).

sale_order_import_edifact/wizard/sale_order_import.py
```diff
@@ -0,0 +1,248 @@
...
+        res.update({"edifact": ("text/plain")})
+        return res
+
+    @api.model
+    def parse_edifact_order(self, filecontent, detect_doc_type=False):
+        # https://github.com/nerdocs/pydifact/blob/master/pydifact/segmentcollection.py
+        if detect_doc_type:
+            if not self.edifact_ok:
+                return None
+            return self.env.context.get("doc_type", "rfq")
```
simahawk: I'm not sure to get this...

QuocDuong1306: Because, to my knowledge, this editfact standard does not contain `rfq` or `order` information. So we set the default as `rfq`

If this module integrates with EDI, we will get the value of `doc_type` context from EDI machinery. Also, if in EDI machinery, we set  `'doc_type: 'order'`, maybe we need to set `default_price_source` because of [this](https://github.com/OCA/edi/blob/32a0e4938f755cbcce815c5ab613794056a8ad2c/sale_order_import/wizard/sale_order_import_view.xml#L62)

cc @rmorant , could you help me to answer:
- Why do you set `doc_type = 'rfq'` and `price_source = 'order'` fixedly in v14? 

(Notes: In v16, `sale_order_import_edifact` disconnects from the EDI framework)

sale_order_import_edifact/wizard/sale_order_import.py
simahawk: that's the point. Instead of having special keys, we can rely on the fact that defaults will be taken automatically.
In they yml example above I'd like to get rid of the special keys `price_source` and `confirm_order`.
In the wiz ctx we should see something like:

```
wiz_ctx:
    default_file_ext: edi
    default_release: d96a
    default_doc_type: rfq
    default_price_source: order
    default_confirm_order: true
```
and if I'm not mistaken the only field missing here is `confirm_order` which I'd like to add as a boolean in the sale.order.import wizard to automatically confirm it instead for relying on a custom machinery in edi_sale_oca.
TBH is my "fault". Is something that I didn't introduce in v 14 and now I'd like to clear it up.
Hope is clear :wink:

sale_order_import_edifact/wizard/sale_order_import.py
simahawk: pls see my comment above regarding defaults via ctx

sale_order_import_edifact/__manifest__.py
```diff
@@ -0,0 +1,31 @@
...
+    "license": "AGPL-3",
+    "application": False,
+    "installable": True,
+    "external_dependencies": {
+        "python": [],
+        "bin": [],
+    },
+    "depends": [
+        "sale_management",
+        "partner_identification",
```
simahawk:
```suggestion
```
redundant

sale_order_import_edifact/models/business_document_import.py
```diff
@@ -0,0 +1,56 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import _, api, exceptions, models
+
+
+class BusinessDocumentImport(models.AbstractModel):
+    _inherit = "business.document.import"
+
+    @api.model
+    def _hook_match_partner(self, partner_dict, chatter_msg, domain, order):
```
simahawk, line 12: matching a partner via gln should be a generic thing... what is really needed here that is specific to edifact?
IMO there's something "wrong" or "too rigid" here https://github.com/OCA/edi/blob/d9309db8561785cdc5af8a5fa099245deba2bf9f/partner_identification_import/models/business_document_import.py#L15 whereas it is assumed that the scheme comes from a UBL file...

If you don't have time to solve it now, I would a TODO to refactor that piece of code and drop this one.

CC @jbaudoux

sale_order_import_edifact/readme/ROADMAP.rst
simahawk: unless is something useful.... @rmorant ?

sale_order_import_edifact/wizard/sale_order_import.py
```diff
@@ -0,0 +1,249 @@
...
+class SaleOrderImport(models.TransientModel):
+    _inherit = "sale.order.import"
+
+    import_type = fields.Selection(
+        selection_add=[("edifact", "EDIFACT")], ondelete={"edifact": "cascade"}
+    )
+    state = fields.Selection(
+        selection_add=[("tech", "Technical")], ondelete={"tech": "set default"}
+    )
+    parsed_json = fields.Text()
```
simahawk: why not a Serialized field?

sale_order_import_edifact/wizard/sale_order_import.py
```diff
@@ -0,0 +1,249 @@
...
+
+
+class SaleOrderImport(models.TransientModel):
+    _inherit = "sale.order.import"
+
+    import_type = fields.Selection(
+        selection_add=[("edifact", "EDIFACT")], ondelete={"edifact": "cascade"}
+    )
+    state = fields.Selection(
+        selection_add=[("tech", "Technical")], ondelete={"tech": "set default"}
```
simahawk: explanation on this "state"?

sale_order_import_edifact/readme/ROADMAP.rst
rmorant: Not at all, it was just a warning, unclear, that things were changing in the new versions.
can be removed.

sale_order_import_edifact/wizard/sale_order_import.py
rmorant: Maybe, but it's read only.

sale_order_import_edifact/wizard/sale_order_import.py
rmorant: It's just to show the object parsed for debug. 
We can remove the state and show the `parsed_json` field only if it contains data.

simahawk: I'd rather add a specific field or yes. check that you have a value there

QuocDuong1306:
I updated

sale_order_import_edifact/wizard/sale_order_import.py
QuocDuong1306: I created a PR for `confirm_order` field in `sale_order_import` module at https://github.com/OCA/edi/pull/839
Could you have a look?

sale_order_import_edifact/__manifest__.py
```diff
@@ -0,0 +1,31 @@
...
+    "website": "https://github.com/OCA/edi",
+    "author": "ALBA Software, Odoo Community Association (OCA)",
+    "maintainers": ["rmorant"],
+    "license": "AGPL-3",
+    "application": False,
+    "installable": True,
+    "external_dependencies": {
+        "python": [],
+        "bin": [],
+    },
```
simahawk:
```suggestion
```

sale_order_import_edifact/__manifest__.py
```diff
@@ -0,0 +1,31 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl-3.0)
+
+{
+    "name": "Import Business Document EDIFACT/D96A Order",
+    "summary": "EDIFACT/D96A Order",
+    "version": "16.0.1.0.0",
+    "development_status": "Alpha",
+    "category": "Sales Management",
+    "website": "https://github.com/OCA/edi",
+    "author": "ALBA Software, Odoo Community Association (OCA)",
```
simahawk:
```suggestion
    "author": "ALBA Software, Camptocamp, Odoo Community Association (OCA)",
```

sale_order_import_edifact/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1,3 @@
+* Rafa Morant <rmorant@albasoft.com> (https://www.albasoft.com)
+* Marc Poch <mpoch@planetatic.com>
+* Duong (Tran Quoc) <duongtq@trobz.com>
```
simahawk, line 3:
```suggestion
* Duong (Tran Quoc) <duongtq@trobz.com>
* Simone Orsi <simone.orsi@camptocamp.com>
```

sale_order_import_edifact/wizard/sale_order_import.py
```diff
@@ -0,0 +1,247 @@
...
+class SaleOrderImport(models.TransientModel):
+    _inherit = "sale.order.import"
+
+    import_type = fields.Selection(
+        selection_add=[("edifact", "EDIFACT")], ondelete={"edifact": "cascade"}
+    )
+    # Make doc_type field support for EDIFACT type
+    doc_type = fields.Selection(readonly=False)
+    parsed_json = fields.Text()
+    edifact_ok = fields.Boolean(compute="_compute_edifact_ok", store=True)
```
simahawk: is `store` necessary here? The value will be cached...

sale_order_import_edifact/wizard/sale_order_import.py
```diff
@@ -0,0 +1,247 @@
...
+
+class SaleOrderImport(models.TransientModel):
+    _inherit = "sale.order.import"
+
+    import_type = fields.Selection(
+        selection_add=[("edifact", "EDIFACT")], ondelete={"edifact": "cascade"}
+    )
+    # Make doc_type field support for EDIFACT type
+    doc_type = fields.Selection(readonly=False)
+    parsed_json = fields.Text()
```
simahawk:
```suggestion
    json_data_preview = fields.Text(help="Used by the btn `button_parse_order_preview` to preview data before importing")
```
And I would add a TODO to move this feature to the base module or to an additional module.

sale_order_import_edifact/wizard/sale_order_import.py
```diff
@@ -0,0 +1,247 @@
...
+            if not ok:
+                ok = b64decode(self.order_file[:4]) == b"UNB"
+            rec.edifact_ok = ok
+
+    def button_parse_order_preview(self):
+        """
+        This method is called by the button to take a quick look at the pydifact
+        structure ('edifact' type) and the parsed object ('order') from
+        the original document before importing the document.
+        """
```
simahawk:
```suggestion
        """
        Generate a preview of the data before importing.

        This method is called by the button to take a quick look at the pydifact
        structure ('edifact' type) and the parsed object ('order') from
        the original document before importing the document.
        """
```

sale_order_import_edifact/wizard/sale_order_import.py
```diff
@@ -0,0 +1,247 @@
...
+            order_file_decoded, self.order_filename, self.partner_id
+        )
+        del parsed_order["attachments"]
+        parsed_obj["order"] = parsed_order
+        self.write(
+            {
+                "parsed_json": json.dumps(parsed_obj, indent=4, default=str),
+            }
+        )
+        action = self.env["ir.actions.act_window"]._for_xml_id(
```
simahawk, line 71: Do you think returning the action is necessary for this feature? 
The purpose of the btn is to update the data for preview.... the wizard does not need any reload. 
Am I wrong?

rmorant: Yes, just update datda for preview.
What do you suggest, an @onchange instead?

QuocDuong1306: Yes, it's necessary. Because when we update the data for review, it will reload and make the wizard disappear. So we need the action to show the wizard again (after updating)

rmorant: You would change the 'Parse' button by 'Preview' and 'tech' status name to 'preview'.
In my opinion, i split 'parsed_json' to 'source_obj' and 'parsed_document', for example, in order to show next to each other in preview form. 
And a "Back" button in the preview step to go back to the 'import' state. 

And if @simahawk move this feature to the base module like he suggested before would be great.

QuocDuong1306: Thank @rmorant ,  @simahawk Do you think like Rmorant? And we should implement this feature here or `sale_order_import`?

simahawk: Ok for the action. As for moving it: I would add a TODO for now and add it to the roadmap.

test-requirements.txt
```diff
@@ -0,0 +1 @@
+odoo-addon-base_edifact @ git+https://github.com/OCA/edi.git@refs/pull/835/head#subdirectory=setup/base_edifact
```
simahawk: this can be trashed

sale_order_import_edifact/wizard/sale_order_import.py
```diff
@@ -0,0 +1,277 @@
...
+        res.update({"edifact": ("text/plain")})
+        return res
+
+    @api.model
+    def parse_edifact_order(self, filecontent, detect_doc_type=False):
+        # https://github.com/nerdocs/pydifact/blob/master/pydifact/segmentcollection.py
+        if detect_doc_type:
+            if not self.edifact_ok:
+                return None
+            return self.env.context.get("doc_type", "rfq")
```
simahawk:
```suggestion
            return self.env.context.get("default_doc_type", "rfq")
```
no?

sale_order_import_edifact/wizard/sale_order_import.py
```diff
@@ -0,0 +1,277 @@
...
+        header = interchange.get_header_segment()
+        # > UNB segment: [['UNOA', '2'], ['5450534000000', '14'],
+        # ['8435337000003', '14'], ['230306', '0435'], '5506']
+
+        msg_type, msg_type_release = edifact_model._get_msg_type(interchange)
+        supported = ["ORDERS"]
+        if msg_type not in supported:
+            raise UserError(_("%s document is not a Sale Order document"))
+
+        doc_type = self.env.context.get("doc_type", "order")
```
simahawk: default_doc_type?
Although, at this stage `self.doc_type` should be set properly no?

simahawk (commented):
Almost there! :mechanical_arm: 
Dependencies to drop and small remarks on the doc type.

sale_order_import_edifact/wizard/sale_order_import.py
simahawk: FTR https://github.com/OCA/edi-framework/pull/31#discussion_r1372739178

QuocDuong1306:
OK @simahawk , done here

simahawk (approved):
LG

PR #828, opened 2023-10-09
[14.0] account_invoice_import: start OCA wizard when user click on "Upload" in supplier invoice tree view

alexis-via:
In the supplier invoice tree view, when the user clicks on the "Upload" button at the top of the tree view, he now uses the OCA invoice import wizard instead of the native feature from the 'account' module.


alexis-via:
upload

NathanQj (approved):

bosd (approved):
LGTM :+1:

PR #835, opened 2023-10-20
[16.0][IMP] base_edifact: Support more EDIFACT format

thienvh332:
- Support "latin-1" format
- Get product fallback on SA if no EAN is given
- Fallback if address information is missing
- Add get price with Price qualifier is AAB
- support getting product description from the IMD segment


base_edifact/models/edifact.py
```diff
@@ -70,7 +70,10 @@ def pydifact_obj(self, docu):
 
     @api.model
     def _loads_edifact(self, order_file):
-        interchange = Interchange.from_str(order_file.decode())
+        try:
+            interchange = Interchange.from_str(order_file.decode())
+        except UnicodeDecodeError:
+            interchange = Interchange.from_str(order_file.decode('latin-1'))
```
simahawk: I would add a TODO for the long term to use chardet. 
Eg: https://github.com/OCA/connector-interfaces/blob/16.0/connector_importer/utils/import_utils.py#L20

simahawk (commented):
still draft?

base_edifact/models/edifact.py
thienvh332: Thanks @simahawk , updated

simahawk (approved):
LG

PR #839, opened 2023-10-25
[16.0][IMP] sale_order_import: Add confirm_order flag to wizard

QuocDuong1306:
When the flag is enabled, the order will be automatically confirmed.
Mostly used when the wizard is used as a tool to import orders.

### DIFF
--- a/sale_order_import/tests/test_order_import.py
+++ b/sale_order_import/tests/test_order_import.py
@@ -145,3 +145,30 @@ class TestOrderImport(TestCommon):
 
         self.assertEqual(len(so.order_line), 2)
         self.assertEqual(so.order_line[0].product_uom_qty, 3)
+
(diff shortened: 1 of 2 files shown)


simahawk (changes requested):
Cool, thank you!
Few remarks:
1. would be nice to have a test
2. please rewrite the commit msg
 
```
[IMP] sale_order_import: add confirm_order flag to wizard

When the flag is enabled, the order will be automatically confirmed.
Mostly used when the wizard is used as a tool to import orders.
```

QuocDuong1306:
Thank @simahawk , I updated

simahawk (approved):

simahawk:
FTR while working on https://github.com/OCA/edi/pull/1090 I've spotted a bug here. I'll fwd port from there when merged.

PR #840, opened 2023-10-25
[14.0][FIX] edi_oca: avoid flooding error in the record chatter

MiquelRForgeFlow:
If the exchange record is linked to a record, and the `process` fails, then it happens this, which is ugly:
Selection_2088

### DIFF
--- a/edi_oca/models/edi_backend.py
+++ b/edi_oca/models/edi_backend.py
@@ -453,7 +453,7 @@ class EDIBackend(models.Model):
         check = self._exchange_process_check(exchange_record)
         if not check:
             return False
-        state = exchange_record.edi_exchange_state
+        old_state = state = exchange_record.edi_exchange_state
         error = False
         message = None
         try:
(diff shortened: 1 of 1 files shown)


simahawk (commented):
Great, looks good.  Question: what about the other actions? Shall we introduce the same check?

simahawk:
Question 2: how are you triggering this? I guess you have the same job running N times? I'm working on a fix for jobs using RetryableJobError.... coming soon w/ some more improvements related to jobs.

MiquelRForgeFlow:
> Question: what about the other actions?

Which other actions. Only `process` has this `_notify_error` call.

> how are you triggering this? I guess you have the same job running N times?

A cron that calls `_cron_check_input_exchange_sync` every 5 minutes.
Selection_2089

MiquelRForgeFlow:
@simahawk can we merge this?

simahawk (approved):

PR #853, opened 2023-11-07
[16.0][IMP] base_edifact: add support for create segment and interchange based on pydifact library

thienvh332:


base_edifact/models/edifact.py
```diff
@@ -242,3 +242,47 @@ def map2odoo_description(self, seg):
             description = seg[2][3]
             return description
         return None
+
+    def _replace_empty(self, value):
```
simahawk:
```suggestion
    def _safe_segment_element(self, value):
```

base_edifact/models/edifact.py
```diff
@@ -242,3 +242,47 @@ def map2odoo_description(self, seg):
...
+        for element in elements:
+            if isinstance(element, list):
+                cleaned_elements.append(
+                    [self._replace_empty(value) for value in element]
+                )
+            else:
+                cleaned_elements.append(self._replace_empty(element))
+        return Segment(*cleaned_elements)
+
+    def create_interchange(self, sender, recipient, control_ref, syntax_identifier):
```
simahawk, line 263: The docstring should state what the method does

simahawk (approved):
LG, pre-approving

base_edifact/models/edifact.py
thienvh332: Thank you @simahawk for your review. I updated it.

base_edifact/models/edifact.py
thienvh332: Thank you @simahawk for your review. I updated it.

rmorant (approved):

simahawk (approved):
LG

PR #856, opened 2023-11-07
[16.0][ADD] base_wamas, base_wamas_ubl: Add new modules to handle the WAMAS document

tuantrantg:


.pre-commit-config.yaml
```diff
@@ -8,6 +8,8 @@ exclude: |
   .svg$|/tests/([^/]+/)?cassettes/|^.copier-answers.yml$|^.github/|
   # Maybe reactivate this when all README files include prettier ignore tags?
   ^README\.md$|
+  # Ignore the text files
```
simahawk, line 11: if needed this should be added to the template

base_wamas/models/wamas_document_default_field.py
```diff
@@ -0,0 +1,11 @@
+# Copyright 2023 Jacques-Etienne Baudoux (BCIM) <je@bcim.be>
+# Copyright 2023 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import models
+
+
+class WamasDocumentDefaultField(models.Model):
```
simahawk, line 8: Please, add docstrings on all classes and methods to explain what they represent/do

base_wamas/models/wamas_document_field.py
```diff
@@ -0,0 +1,279 @@
...
+        elif self.ttype == "int":
+            res = self._get_int(value)
+        elif self.ttype == "float":
+            res = self._get_float(value)
+        elif self.ttype == "date":
+            res = self._get_date(value)
+        elif self.ttype == "datetime":
+            res = self._get_datetime(value)
+        elif self.ttype == "bool":
+            res = self._get_bool(value)
```
simahawk, line 66:
```suggestion
        getters = dict(
            str=self._get_string,
            int=self._get_int
            [....]
        )
        return getters[self.ttype](value)
```
Those methods should be renamed like... `_get_value_from_string__[$ttype]`

An alternative to the dict of getters is:

```
handler = getattr(self, f"_get_value_from_string__{self.ttype}`)
return handler(value)
```

base_wamas/models/wamas_document_field.py
```diff
@@ -0,0 +1,279 @@
...
+
+        return res
+
+    def _get_bool(self, value):
+        self.ensure_one()
+
+        res = value.strip()
+
+        ICP = self.env["ir.config_parameter"]
+        lst_false = safe_eval(ICP.get_param("wamas_document_list_false_value", ["f"]))
```
simahawk, line 159: do we really need params for this?

base_wamas/models/wamas_document_element.py
```diff
@@ -0,0 +1,129 @@
...
+    field_ids = fields.One2many("wamas.document.field", "element_id", "Fields")
+
+    def get_values_from_element(
+        self, row, convert_type=False, result_type="ordered_dict"
+    ):
+        self.ensure_one()
+
+        res = False
+
+        if result_type == "list_tuple":
```
simahawk, line 52: What is the purpose of having a different result?
IMO it would be better to have 2 methods...

base_wamas/models/wamas_document_field.py
```diff
@@ -0,0 +1,279 @@
...
+
+
+class WamasDocumentField(models.Model):
+    _name = "wamas.document.field"
+    _description = "WAMAS Document Field"
+
+    name = fields.Char(required=True)
+    code = fields.Char(required=True)
+    ttype = fields.Selection(FIELD_TYPES, "Type", required=True)
+    len_field = fields.Integer("Length", required=True)
```
simahawk, line 45: Can you add  some `help` ?

base_wamas/tests/test_base_wamas.py
```diff
@@ -0,0 +1,68 @@
...
+
+import filecmp
+import os
+import tempfile
+
+from odoo.tests.common import TransactionCase
+from odoo.tools import file_open, file_path
+
+
+class TestBaseWamas(TransactionCase):
```
simahawk, line 13: It would be nice to have full test coverage, possibly via unit tests

base_wamas/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,7 @@
+Provides following models:
+
+* WAMAS Document Template, to define WAMAS document, it includes WAMAS document elements.
+* WAMAS Document Element, to define elements in WAMAS document, such as: WEAK, WEAP...
+* WAMAS Document Field, to define fields that WAMAS elment has, there are a lot of fields in a WAMAS element.
+* WAMAS Document Default Field, to define default fields of WAMAS document has.
+* WAMAS Document Default Field Template, to define the template of default fields of the elements.
```
simahawk, line 7: pls add a CONFIGURE.rst

base_wamas_ubl/data/ir_actions_server_data.xml
```diff
@@ -0,0 +1,196 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo noupdate="0">
+
+    <record id="ias_get_ubl_version" model="ir.actions.server">
```
simahawk, line 4: I don't get why you need server actions... :thinking:

simahawk (commented):
early comments

base_wamas/models/wamas_document_field.py
tuantrantg: it's needed, so that user can configure what value they want to get

base_wamas_ubl/data/ir_actions_server_data.xml
tuantrantg: because UBL has a lot of templates and we cannot add hard-coding functions, user can add functions that they need for their UBL documents, so I use `ir.actions.server` for this purpose.

it's may not right to use `ir.actions.server` like this, but user can control their document actively

base_wamas/tests/test_base_wamas.py
tuantrantg: sure, I'll do it, if I finish the other main functions

base_wamas/models/wamas_document_element.py
tuantrantg: > What is the purpose of having a different result?

- the `list_tuple`: `list` is each line of UBL document and each `tuple` (in `list`) is each element in the line -> one `list_tuple` related to one line -> it's so generic to use for any purpose
- the `ordered_dict`: `ordered_dict` is a line and each item of `ordered_dict` is each element in the line -> I use this to build the `ordered dict` structure for converting wamas -> ubl in `base_wamas_ubl`

> IMO it would be better to have 2 methods...

I agree, it should be better if have 2 methods, I'll split it into 2 methods

PR #862, opened 2023-11-18
[15.0][BPRT] edi_oca

etobella:
Trying to split https://github.com/OCA/edi/pull/814 in several PRs (maybe one for each module?)


etobella:
@simahawk I removed all the not migrated modules commits. Now it should be good. Can you review it?

pcastelovigo:
please

pcastelovigo:
@etobella @simahawk I merged locally and the conflicts are trivial, would it be possible to merge? It is necessary to advance with modules of a location

etobella:
Conflicts fixed!

@simahawk Can you review this please? This commits were not ported at the merge and we need now to do that.

Thanks !! :smile:

simahawk:
checking

simahawk:
@etobella it would be better to split by module... the version bump for instance won't be correct because you need to keep the specific version for edi_oca but the other modules must be updated (eg: edi_storage, which seems not an improvement required here).
Can we remove them or tests will fail?

Also, regarding the manual version bump, pls rewrite `[FIX] edi_oca: Bump version manually` -> `edi_oca: 15.0.1.6.0` or ``edi_oca: bump 15.0.1.6.0`. Is not a fix and is something that should be prominent.

etobella:
Bumped versions manually, so we could proceed for merge  following https://github.com/OCA/edi/pull/870#issuecomment-1824137485

PR #865, opened 2023-11-21
[14.0] account_invoice_ubl*: fix tests 

### LINKED ISSUE #864: [14.0] account_invoice_ubl_peppol tests broken
The module was migrated here https://github.com/OCA/edi/pull/795 but tests actually never passed.
Weird thing is that the build on 14.0 is green... but tests are broken and in fact they fail on  #855.

See https://github.com/OCA/edi/pull/795/#issuecomment-1820813357

simahawk:
Tests for account_invoice_ubl and account_invoice_ubl_peppol were broken.

PDF rendering was not fully done and account_invoice_ubl_peppol was not properly migrated to v14.

Closes https://github.com/OCA/edi/issues/864


simahawk:
~Great, now the test is working but it hangs forever on the rendering of the PDF~

PR #867, opened 2023-11-21
[14.0] Misc boost tests 

simahawk:


bosd (approved):
LGTM

bofiltd (approved):
LGTM

bosd:
Oops we merged this..

But overlooked this error..
image

simahawk:
I think there's something crazy in the TestCase suite of Odoo core: it does not detect import errors as errors in the tests...

simahawk:
voilá https://github.com/OCA/edi/pull/883

bosd:
> I think there's something crazy in the TestCase suite of Odoo core: it does not detect import errors as errors in the tests...

Indeed, seems something to do with test suite..
Also for cases like this https://github.com/OCA/server-ux/pull/779/files

Migration script appearantly does not cover. Tests pass even with these errors.
Should we discuss this somewehere? Maintainer repo?

PR #870, opened 2023-11-23
[15.0] Backport EDI_OCA: Test if splitting by module works properly

etobella:


etobella:
@simahawk It is not so easy to split by module, some changes on edi_oca affects other modules

simahawk:
> @simahawk It is not so easy to split by module, some changes on edi_oca affects other modules

ok, keep those changes together but take care of bumping versions w/ proper commit msg for each module :wink:

PR #737, opened 2023-03-23
[15.0][MIG] base_business_document_import

hhgabelgaard:
Std. migrations

Depends on:
- [x] #736

PR #872, opened 2023-11-23
[15.0][MIG] base_business_document_import

pcastelovigo:
As requested in #466

Module works, hopefully I'll make it pass the requirements


pcastelovigo:
@simahawk I'm reading plan is to deprecate this... Could you give me directions? I would like to clone from OCA but a repository of my own is acceptable

simahawk:
> @simahawk I'm reading plan is to deprecate this... Could you give me directions? I would like to clone from OCA but a repository of my own is acceptable

Where are you reading this? 
The module is already available on v16 https://github.com/OCA/edi/tree/16.0/base_business_document_import hence I doubt we'll deprecate or do major refactoring before v17 or even 18

etobella:
I think you are not maintaining Commit history

bosd:
Maybe better to base the work upon: https://github.com/OCA/edi/pull/737

pcastelovigo:
> Maybe better to base the work upon: #737

I have a test-passing clone on a repo, but it doesnt pass codecov

PR #874, opened 2023-11-25
[15.0][FIX] fix uncaught promise error

pcastelovigo:
Fix uncaught promise error happening when domain search not found because additonal space


etobella:
This kind of changes need to be done on weblate
https://translation.odoo-community.org

I did the change myself there.

PR #880, opened 2023-11-29
[14.0] edi: fix consumer mixin copy origin exc record

simahawk:
The origin record should never been copied.
Its values should always be set specifically by a framework action.


MiquelRForgeFlow (approved):

pedrobaeza (approved):
/ocabot merge minor

PR #882, opened 2023-12-01
[14.0][ADD] edi_product_import

florentx:
New glue add-on for `product_import`

(initial PR #760)


simahawk:
I'd say this module should go to https://github.com/OCA/edi-framework and be based on https://github.com/OCA/edi-framework/pull/43 (to be backported to 14.0) so that you can also see exchanges linked to the product.

mmequignon (approved):
Ok with that, though I don't know about @simahawk comment

bofiltd (approved):
LGTM

bosd (approved):

bofiltd:
Any update on this?

bosd:
> Ok with that, though I don't know about @simahawk comment

@simahawk Can you elaborate?
Should we merge it in this repo?

simahawk (changes requested):
my comment remains valid

florentx:
For Odoo 14, it would add some burden to depend on `edi_product_oca` because:
 - add-on `edi_oca` is still in `OCA/edi` repositiory and it will probably stay there (14.0 is old/stable enough)
 - add-on `edi_product_oca` does not exist for 14.0, neither in `OCA/edi-framework` and neither in `OCA/edi`
 - even if porting `edi_product_oca` to 14.0 seems doable, it seems it will just make the dependencies more complicated for not much benefits

simahawk:
Fair enough. I definitely see benefits in depending on it, in general... Yet, as for 14 let's go this way and simply add a TODO for migration.

PR #883, opened 2023-12-01
[14.0] account_invoice_*: fix tests import

simahawk:
fixes https://github.com/OCA/edi/pull/867


bosd (approved):
Thanks @simahawk 
This is ineed the proper way for V14

PR #884, opened 2023-12-04
[16.0][FIX] base_business_document_import tests

nilshamerlinck:
- default `base.main_partner`'s name is "YourCompany" but if stock module is installed, it is changed to "My Company (San Francisco)"
- so here's a fix for tests to pass with or without stock module installed

### DIFF
--- a/base_business_document_import/tests/test_business_document_import.py
+++ b/base_business_document_import/tests/test_business_document_import.py
@@ -96,7 +96,7 @@ class TestBaseBusinessDocumentImport(TransactionCase):
             "xmlid": "base.main_partner",
         }
         partner = bdio._direct_match(partner_dict, self.env["res.partner"], True)
-        self.assertEqual(partner.name, "YourCompany")
+        self.assertEqual(partner.id, self.env.ref("base.main_partner").id)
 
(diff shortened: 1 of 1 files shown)


base_business_document_import/tests/test_business_document_import.py
```diff
@@ -96,7 +96,9 @@ def test_direct_match_xmlid(self):
             "xmlid": "base.main_partner",
         }
         partner = bdio._direct_match(partner_dict, self.env["res.partner"], True)
-        self.assertEqual(partner.name, "YourCompany")
+        # by default base.main_partner's name is "YourCompany"
+        # but if stock module is installed, it is changed to "My Company (San Francisco)"
+        self.assertTrue(partner.name in ("YourCompany", "My Company (San Francisco)"))
```
simahawk: wouldn't be easier to check for the ID of the record instead?

nilshamerlinck: true, updated :+1:

bosd (approved):
LGTM

rvalyi (approved):

simahawk (approved):

PR #890, opened 2023-12-14
[16.0][MIG] account_invoice_export

TDu:
Along with the migration, I split the module to make the dependency on queue job optional. See commits for details.


sebalix (approved):
Fixup to squash, LG

account_invoice_export/models/account_move.py
```diff
@@ -65,12 +65,13 @@ def _export_invoice(self):
         file_data = self._get_file_for_transmission_method()
         headers = self.transmit_method_id.get_transmission_http_header()
         url = self.transmit_method_id.get_transmission_url()
-        res = requests.post(url, headers=headers, files=file_data)
+        res = requests.post(url, headers=headers, files=file_data, timeout=5)
```
grindtildeath: Isn't 5 seconds a bit low? Looking into odoo code it's generally between 15 and 30s. Could be good to add a parameter (a TODO in the code is also fine while working on the mig)

account_invoice_export/tests/common.py
```diff
@@ -0,0 +1,50 @@
+from odoo.tests.common import TransactionCase
```
grindtildeath, line 3: Please add header

grindtildeath (approved):
CI to fix but LG

account_invoice_export/tests/common.py
TDu: done

account_invoice_export/models/account_move.py
TDu: I changed it to 10.
And also left a comment to improve with configuaration.

TDu:
@simahawk  This one is ready to be merged IMO

PR #895, opened 2023-12-21
[15.0][FIX] edi_oca: avoid pre-migrate error

SergiCForgeFlow:
Column _model_manual_btn_ for table _edi_exchange_type_ does not exist for versions 13 and lower. So if a migration is being performed from some of those versions to version 15, the pre-migrate script may give an error when executing the SQL query with that column.


etobella (approved):
The same might happen on 14.0. Can you review it?

LoisRForgeFlow (approved):
LGTM

SergiCForgeFlow:
> The same might happen on 14.0. Can you review it?

Yes, the back port is here: [#897](https://github.com/OCA/edi/pull/897)

PR #896, opened 2023-12-21
[MIG] base_edi: Migration to 17.0

john-herholz-dt:


john-herholz-dt:
Didn't see it's duplicate of #894

PR #875, opened 2023-11-27
[16.0][ADD] base_wamas_ubl: Add new module to convert WAMAS files to UBL files and vice versa

tuantrantg:
- Refactoring of https://github.com/OCA/edi/pull/856 to get rid of the odoo abstraction
- Depends on:
  - https://github.com/OCA/oca-addons-repo-template/pull/240

PR #899, opened 2024-01-02
[16.0][ADD] partner_export_wamas

QuocDuong1306:
This module provides a method to export partner data as WAMAS format.

Depend on:
- https://github.com/OCA/edi/pull/875


.pre-commit-config.yaml
```diff
@@ -14,6 +14,8 @@ exclude: |
   ^docs/_templates/.*\.html$|
   # Don't bother non-technical authors with formatting issues in docs
   readme/.*\.(rst|md)$|
+  # Ignore test files in addons
```
simahawk, line 17: this got integrated in the template

partner_export_wamas/__manifest__.py
```diff
@@ -0,0 +1,16 @@
...
+{
+    "name": "Partner Export WAMAS",
+    "summary": """Module provides a method to export partner as WAMAS format""",
+    "version": "16.0.1.0.0",
+    "development_status": "Alpha",
+    "category": "Hidden",
+    "website": "https://github.com/OCA/edi",
+    "license": "AGPL-3",
+    "author": "Camptocamp,Odoo Community Association (OCA)",
+    "depends": ["contacts", "base_wamas_ubl"],
```
simahawk, line 14: if you depend on `_ubl` I suspect this module should be called `_ubl` too.

partner_export_wamas/models/res_partner.py
```diff
@@ -0,0 +1,27 @@
...
+        self.ensure_one()
+        if not telegram:
+            raise ValueError(_("Need telegram for exporting"))
+        # If having a specific dict for the partner, we use it
+        if specific_dict and isinstance(specific_dict, dict):
+            dict_partner = specific_dict
+        else:
+            dict_partner = self.read()[0]
+        base_wamas_ubl = self.env["base.wamas.ubl"]
+        return base_wamas_ubl.dict2wamas(dict_partner, telegram)
```
simahawk, line 27: To answer my point above regarding depending on _ubl module: IMO here we should look at the right helper model via configuration/parameter and use always the same method to get the data.
Something like:
```suggestion
    def _wamas_export(self, specific_dict=None, telegram=False, export_format=None):
        """
        Export the partner data as WAMAS format

        :return: partner data as WAMAS format
        :rtype: bytes
        """
        self.ensure_one()
        if not telegram:
            raise ValueError(_("Need telegram for exporting"))
        # If having a specific dict for the partner, we use it
        if specific_dict and isinstance(specific_dict, dict):
            dict_partner = specific_dict
        else:
            dict_partner = self.read()[0]

        export_format = export_format or self.export_format  # if we want to make it configurable by partner
        handler = self._wamas_export_get_handler(export_format)
        return handler.dict2wamas()

    def _wamas_export_get_handler(self, export_format):
        try:
            return self.env[f"base.wamas.{export_format}"]
        except KeyError:
            raise UserError(_("Not supported format %(fmt)s", fmt=export_format))
```
can you please add a TODO in the method at list?

partner_export_wamas/models/res_partner.py
```diff
@@ -0,0 +1,27 @@
...
+        :rtype: bytes
+        """
+        self.ensure_one()
+        if not telegram:
+            raise ValueError(_("Need telegram for exporting"))
+        # If having a specific dict for the partner, we use it
+        if specific_dict and isinstance(specific_dict, dict):
+            dict_partner = specific_dict
+        else:
+            dict_partner = self.read()[0]
```
simahawk, line 25: `self.read()` will read too many things. It can be quite huge.
IMO we should use cached data and reduce the fields to read somehow. 
IF you don't know yet what to do and you are only using the specific_dict just remove it and make the specific_dict mandatory.

simahawk (changes requested):

simahawk:
Another consideration regarding this `_wamas_export` method: it has nothing specific to partners.
I wonder if we really need this module at all. We could add that to `base` model and add some constraints on its usage.
As an alternative we could move it to a mixin and then use it where needed only.

It could also be a method on the helper itself. You could simply call `env["base.wamas"].record_to_wamas(record, ,....)` and avoid completely the need for this change here.
Can you elaborate a bit  on the use case?

QuocDuong1306:
Hello, this PR is replaced by: https://github.com/OCA/edi/pull/903

PR #900, opened 2024-01-03
[16.0][ADD] product_export_wamas

QuocDuong1306:
This module provides a method to export products and product packagings as WAMAS format

Depends on:
- https://github.com/OCA/edi/pull/875
- https://github.com/OCA/oca-addons-repo-template/pull/240


simahawk:
same remarks as #899 : I don't think we need all this code

QuocDuong1306:
Hello, this PR is replaced by: https://github.com/OCA/edi/pull/903

PR #903, opened 2024-01-11
[16.0][IMP] base_wamas_ubl: Add record_data_to_wamas method

QuocDuong1306:


base_wamas_ubl/models/base_wamas_ubl.py
```diff
@@ -42,3 +42,17 @@ def get_wamas_type(self, str_file):
     @api.model
     def wamas2wamas(self, str_file):
         return wamas2wamas(str_file)
+
+    @api.model
+    def record_to_wamas(self, specific_dict, telegram_type):
+        """
+        Export data as WAMAS format
+
```
simahawk: I'm not sure this method is adding anything useful compared to `dict2wamas`.
In the future we might have a real record passed here w/ the possibility to automatically read data.
However, as is not the case now I would change it like this:
```suggestion
    @api.model
    def record_data_to_wamas(self, data, telegram_type):
        """
        Convert Odoo record data to WAMAS format

```

QuocDuong1306: Thank @simahawk , I updated

simahawk (approved):

PR #904, opened 2024-01-15
[16.0][FIX] base_ubl: street3 in UBL generation

jbaudoux:
In an address, the real street is usually put on the last "street" field. Ensure we get this in StreetName or AdditionalStreetName but not in another element.

Example from https://en.wikipedia.org/wiki/Address
 - name: Ms H Williams
 - street: Finance and Accounting
 - street2: Australia Post
 - street3: 219–241 Cleveland St

Also improve UBL generation in case one of the street field is empty

cc @simahawk @nilshamerlinck

### DIFF
--- a/base_ubl/models/ubl.py
+++ b/base_ubl/models/ubl.py
@@ -38,17 +38,31 @@ class BaseUbl(models.AbstractModel):
     @api.model
     def _ubl_add_address(self, partner, node_name, parent_node, ns, version="2.1"):
         address = etree.SubElement(parent_node, ns["cac"] + node_name)
-        if partner.street:
+        if partner.street or partner.street2:
             streetname = etree.SubElement(address, ns["cbc"] + "StreetName")
-            streetname.text = partner.street
-        if partner.street2:
+            streetname.text = partner.street or partner.street2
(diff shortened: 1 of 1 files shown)


base_ubl/models/ubl.py
```diff
@@ -38,17 +38,30 @@ def _ubl_add_country(self, country, parent_node, ns, version="2.1"):
...
             streetname = etree.SubElement(address, ns["cbc"] + "StreetName")
-            streetname.text = partner.street
-        if partner.street2:
+            streetname.text = partner.street or partner.street2
+        if partner.street and partner.street2:
             addstreetname = etree.SubElement(
                 address, ns["cbc"] + "AdditionalStreetName"
             )
             addstreetname.text = partner.street2
         if hasattr(partner, "street3") and partner.street3:
```
nilshamerlinck, line 50:
```suggestion
        # if oca/partner-contact/partner_address_street3 is installed
        if hasattr(partner, "street3") and partner.street3:
```

jbaudoux: comment added

simahawk (approved):

PR #913, opened 2024-01-23
[16.0][IMP] base_wamas_ubl: update address export

thienvh332:


base_wamas_ubl/lib/wamas/utils.py
```diff
@@ -275,6 +319,9 @@ def generate_wamas_line(dict_item, grammar, **kwargs):  # noqa: C901
                 args = (dict_wamas_out,)
                 args += ast.literal_eval(re.search(r"\((.*?)\)", df_func).group(0))
                 df_func = "get_date_from_field"
+            elif "get_Adrs_" in df_func:
```
simahawk, line 324: TBH this way of passing arguments seems a bit weird, fragile and hard to understand overall.
The `df_func` or another key in the handlers mapping should provide a way to determine those args.
You could make them all classes instead of simple methods and accept parameters and get back arguments...
Do not change it now but can you please add a TODO to cleanup this?

simahawk (changes requested):

simahawk:
Can you rewrite the last commits pls?

```
[FIX] base_wamas_ubl: Fix bug wamas2ubl

AUSKQ has wrong value in 'IvAusk_ExtRef' field

[UPD] base_wamas_ubl: Update grammars ARTEAN, WEAK, WEAP
```
The explanation of the changes goes to the description. The fact that tests were updated is not relevant in general as it is expected whenever you have test coverage for the feature you touched :wink:

simahawk:
@thienvh332 pls avoid adding the odoo version in commit msg :pray:

simahawk (approved):

PR #914, opened 2024-01-23
[13.0][FIX] base_ubl: use variant attributes

JaumeBforgeFlow:
If we use `attribute_line_ids` then we will be adding and *AdditionalItemProperty* tag for each of the possible values defined on the product.template for each attribute. 
Using `product_template_attribute_value_ids` we only use the ones that apply to the variant

### DIFF
--- a/base_ubl/models/ubl.py
+++ b/base_ubl/models/ubl.py
@@ -450,14 +450,14 @@ class BaseUbl(models.AbstractModel):
                         node_name="ClassifiedTaxCategory",
                         version=version,
                     )
-            for attribute_value in product.attribute_line_ids.mapped("value_ids"):
+            for ptav in product.product_template_attribute_value_ids._only_active():
                 item_property = etree.SubElement(
                     item, ns["cac"] + "AdditionalItemProperty"
                 )
                 property_name = etree.SubElement(item_property, ns["cbc"] + "Name")
-                property_name.text = attribute_value.attribute_id.name
+                property_name.text = ptav.attribute_id.name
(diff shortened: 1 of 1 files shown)


JordiMForgeFlow (approved):
Makes sense 👍🏼

GuillemCForgeFlow (approved):
LGTM 👍🏽

base_ubl/models/ubl.py
```diff
@@ -450,14 +450,14 @@ def _ubl_add_item(
                         node_name="ClassifiedTaxCategory",
                         version=version,
                     )
-            for attribute_value in product.attribute_line_ids.mapped("value_ids"):
+            for ptav in product.product_template_attribute_value_ids:
```
MiquelRForgeFlow:
```suggestion
            for ptav in product.product_template_attribute_value_ids._only_active():
```
?

AaronHForgeFlow (approved):
:+1:

base_ubl/models/ubl.py
JaumeBforgeFlow: Hi @MiquelRForgeFlow i've applied your suggestion, makes sense

simahawk (approved):
LG. To be fwd ported?

JaumeBforgeFlow:
> LG. To be fwd ported?

Here are the PRs  :smile: 

- https://github.com/OCA/edi/pull/919
- https://github.com/OCA/edi/pull/920
- https://github.com/OCA/edi/pull/921

PR #917, opened 2024-02-01
[16.0][IMP] base_wamas_ubl: rebuild mapping dynamically and override default mapping

QuocDuong1306:


base_wamas_ubl/tests/test_base_wamas_ubl.py
```diff
@@ -75,6 +77,31 @@ def setUpClass(cls):
...
             },
+            "MAPPING_UNITCODE_WAMAS_TO_UBL": {
+                "unitCode": MappingDict(
+                    {
+                        "BOT": "XBQ",  # plastic bottle
+                        "BOUT": "C62",  # Unit
+                        "BOITE": "XBX",  # box
+                        "LITRE": "LTR",  # litre
+                        "PET": "XBO",  # glass bottle
+                        "TETRA": "X4A",  # tetra pack, changed 'X4B' to 'X4A'for testing
```
QuocDuong1306, line 88: Hi reviewer, [the default of 'TETRA'](https://github.com/OCA/edi/blob/644353cf7b25b9bfd91f36368c4f1c2b47ea9289/base_wamas_ubl/lib/wamas/const.py#L190) is X4B, so i just want to change to X4A to test

To make sure that it override the default no

base_wamas_ubl/lib/wamas/utils.py
```diff
@@ -211,9 +211,16 @@ def get_date_from_field(*args):
     return res
 
 
-def convert_unit_code(key, val):
+def convert_unit_code(key, val, extra_data=False):
```
simahawk: why not an explicit  key?

```suggestion
def convert_unit_code(key, val, unit_mapping=None):
```

base_wamas_ubl/models/base_wamas_ubl.py
```diff
@@ -25,19 +26,75 @@ def wamas2dict(self, str_file):
...
+                ] = MAPPING_UNITCODE_UBL_TO_WAMAS
+        return dict2ubl(template, data, extra_data=extra_data)
 
     @api.model
     def wamas2ubl(self, str_file, extra_data=False):
+        extra_data = extra_data if extra_data else {}
+        uom_records = self.env["uom.uom"].search([])
+        if uom_records:
+            (
+                MAPPING_UNITCODE_WAMAS_TO_UBL,
```
simahawk: Please, do not add this everywhere. Just define an helper method to get the result no matter where you call it from.

simahawk (changes requested):

base_wamas_ubl/lib/wamas/utils.py
```diff
@@ -211,9 +211,16 @@ def get_date_from_field(*args):
...
     if key in LST_FIELD_UNIT_CODE:
-        return MAPPING_UNITCODE_UBL_TO_WAMAS["unitCode"].get(val, val)
+        MAPPING = (
+            unit_mapping
+            and unit_mapping.get(
+                "MAPPING_UNITCODE_UBL_TO_WAMAS", MAPPING_UNITCODE_UBL_TO_WAMAS
+            )
+            or MAPPING_UNITCODE_UBL_TO_WAMAS
+        )
+        return MAPPING["unitCode"].get(val, val)
```
sebalix: Here `MAPPING` doesn't refer to a "constant" object at the module level, better to write it in lowercase:
```suggestion
        _mapping = (
            unit_mapping
            and unit_mapping.get(
                "MAPPING_UNITCODE_UBL_TO_WAMAS", MAPPING_UNITCODE_UBL_TO_WAMAS
            )
            or MAPPING_UNITCODE_UBL_TO_WAMAS
        )
        return _mapping["unitCode"].get(val, val)
```

base_wamas_ubl/lib/wamas/utils.py
```diff
@@ -510,13 +522,20 @@ def wamas2dict(
 
 def dict2ubl(template, data, verbose=False, extra_data=False):
     t = miniqweb.QWebXml(template)
+    MAPPING = (
```
sebalix: Same remark
```suggestion
    _mapping = (
```

simahawk (approved):
:see_no_evil: :see_no_evil: :see_no_evil: 
I definitely dislike this kind of implementation, sorry.
Not a blocker tho as this module's utils should be refactored anyway.

jbaudoux:
I dropped all the unitcode mapping in https://github.com/OCA/edi/pull/930/commits/46b79ba72898cf96110b80ca3f8d9de49e1a3d9e

@simahawk This PR can be closed

PR #929, opened 2024-02-06
[16.0] base_wamas_ubl: Update address export

thienvh332:


base_wamas_ubl/lib/wamas/utils.py
```diff
@@ -251,19 +251,67 @@ def get_Adrs_Name(a):
 
 
 def get_Adrs_Name2(a):
-    return next(filter(None, [a["PartyName"], a["Department"], a["StreetName"]]), "")
+    return next(
```
simahawk, line 254: Can we please do something less complex and more readable than this?
It's a dictionary, the keys are already there you don't save anything by using `iter` and `filter` + `list`.
Eg:
```
candidates = (
    "ContactName",
    "PartyName",
    "Department",
    "StreetName",
    "AdditionalStreetName",
)
res = [a[x] for x in candidates if a.get(x)]
return res[0] if res else None
```

Same everywhere else.  You can even have an helper func `get_first_valid_value(a_dict, candidates)` or similar.

simahawk (changes requested):

simahawk:
#930

PR #930, opened 2024-02-07
[16.0] base_wamas_ubl: Update address/uom/packaging/extref & kretk/p export 

jbaudoux:
Replaces as includes commits from:
* https://github.com/OCA/edi/pull/923
* https://github.com/OCA/edi/pull/928
* https://github.com/OCA/edi/pull/929
* https://github.com/OCA/edi/pull/931
* https://github.com/OCA/edi/pull/917

Other Improvements:
* Simplify code
* Fix all CLI scripts to make conversions. This also allow to regenerate tests results
* Replaced getopt by argparse in all CLI scripts
* Added method docstrings
* Move tests under tests directory
* Make one unittest per test
* Use with statement for all file open
* Fix encodings. If the sample file is properly put with right encoding, there is no need to encode/decode. Giving right encoding to file_open is enough
* Improved some samples. Regenerated sample results
* Ensure field order is correct on generated wamas dict (replaced dict by OrderedDict)
* Added FIXME to move some tests to wamas lib
* Dropped unitcode mapping. Directly provide uom wamas_code with unitCodeListAgencyName="WAMAS" instead of unece

Warning:
* The line identifier is given in UBL by the `DespatchLine/ID` and not `DespatchLine/OrderLineReference/ID`. This should match the line DB id.


jbaudoux:
ping @simahawk @sebalix

simahawk:
@jbaudoux pls squash and check pre-commit

base_wamas_ubl/lib/wamas/tests/samples/line_WATEPQ_-_normal.dict
```diff
@@ -1,21 +1,26 @@
...
-             'IvTep_MId_MHD': '20231231',
-             'IvTep_MId_ResNr': '',
-             'IvTep_MId_ThmKz': 'N',
-             'IvTep_MId_WeNr': '12    eCAM  014020',
-             'IvTep_PosNr': '000010',
-             'IvTep_PrimThmKz': 'N',
-             'IvTep_TeId': '376401426411929180',
-             'Mngs_Gew': '000000000000',
-             'Mngs_Mng': '000000030000'}]}
+{'WATEPQ': [OrderedDict([('Telheader_Quelle', 'WAMAS'),
```
simahawk, line 1: can we take the chance to rename these files to .py?

simahawk (approved):

PR #932, opened 2024-02-09
[15.0][IMP] edi_oca: add company_id to edi_exchange_record

ThiagoMForgeFlow:
@ForgeFlow


edi_oca/models/edi_exchange_record.py
```diff
@@ -118,6 +121,14 @@ class EDIExchangeRecord(models.Model):
         ),
     ]
 
+    @api.depends("res_id", "model")
+    def _compute_company_id(self):
+        for rec in self:
+            if rec.record._fields.get("company_id", False):
```
etobella:
```suggestion
            if rec.record and rec.record._fields.get("company_id", False):
```

PR #955, opened 2024-03-20
[16.0][IMP] despatch_advice_import: allow to validate picking with quantity larger than the reserved quantity

tuantrantg:


jbaudoux (approved):

despatch_advice_import/wizard/despatch_advice_import.py
```diff
@@ -186,8 +189,15 @@ def process_data(self, parsed_order_document):
                 lambda x: x.state not in ("cancel", "done")
             )
             moves_qty = sum(stock_moves.mapped("product_qty"))
+            allow_validate_over_qty = self.allow_validate_over_qty or self._context.get(
```
simahawk: do we really need a special ctx key? This can be set by default w/ `default_allow_validate_over_qty` in the ctx when creating the wiz rec.

tuantrantg: this context is used in case we call `process_data` directly without creating wizard, like we did in the tests in [here](https://github.com/OCA/edi/pull/955/files#diff-34a714f35355b8859990138fcb69ad559dc7a75ba60d648548f1e7e31ec21ac4R108)

cyrilmanuel (approved):

despatch_advice_import/wizard/despatch_advice_import.py
simahawk: Either this method is a class method / model method or is an instance method. 
If you rely on on `self` here you should rely always on it.
Otherwise, add a kw arg to `process_data` to pass this option explicitly.

simahawk (changes requested):

tuantrantg:
hi @simahawk please review it again

simahawk (approved):

PR #668, opened 2022-10-02
[15.0][MIG] account_invoice_import

alexis-via:
Up-to-date migration with the latest code from v14, including the "import as empty invoice" feature.

PR #959, opened 2024-03-29
 [15.0][ADD] account_edi_simple_pdf

hbrunn:
this is account_invoice_import_simple_pdf but then plugged into Odoo's existing invoice import mechanism.

Missing:

- [ ] amend invoices by upload
- [ ] cleanup/refactor if we go for 2. below

I've started this as a spiritual successor of account_invoice_import_simple_pdf, that's why I preserved the commits, and prepared for possibly having both - we'd have to move the configuration and the mixin to its own module on which both depend and just apply the mixin. But now I'm unsure, as those functionalities conflict, so coexistence doesn't really seem an option to me.

What do reviewers think?

1. keep history, this will cause problems for a migration of account_invoice_import_simple_pdf
2. remove history, don't try to be compatible with a future account_invoice_import_simple_pdf migration


pedrobaeza:
@hbrunn check this that we are cooking that is very generic for any possible document: https://github.com/Tecnativa/edi/tree/15.0-add-base_import_pdf_simple/base_import_pdf_simple

We have started it for importing sales orders, but it can serve for anything

We'll propose it to OCA soon.

hbrunn:
oh that looks interesting, thanks. The readme is hard to read for me though ;-)

Thing is, @alexis-via's code is time proven and my customer is used to it, they just want to have that in their v15 instance too. And rather now than soon. I don't see a problem having `account_edi_simple_pdf` now and a possible `import_pdf_simple_account_move` (what will this depend on? `account_edi` or `account_invoice_import`?) next to each other here. Maybe a word or two about the differences between them might be helpful when we've come so far.

We do already keep the account_import* stack and the parallel facturx import here, so more choice, the merrier I'd say. Actually for facturx a comparison in the repo's/module's readme would be very helpful too, I can't tell a customer from the top off my head why they should use the OCA version instead of the core one.

@alexis-via what's your take on keeping the history here? And would you prefer me to disentangle this completely from account_invoice_import_simple_pdf whenever this lands in v15, or prepare for sharing most of the code?

pedrobaeza:
The problem with existing `*_simple_import*` module is that it doesn't handle properly multi-lines import with all the needed fields. We have already tackled this in a generic way to serve to import sales order, purchase orders, invoices...

jans23:
Is this used by the ordinary Upload button?

grafik

hbrunn:
yes, watch your logs for messages about recognizing a partner, and further messages

account_edi_simple_pdf/README.rst
```diff
@@ -0,0 +1,212 @@
...
+
+1. `PyMuPDF <https://github.com/pymupdf/PyMuPDF>`_ which is a Python binding for `MuPDF <https://mupdf.com/>`_, a lightweight PDF toolkit/viewer/renderer published under the AGPL licence by the company `Artifex Software <https://artifex.com/>`_.
+#. `pdftotext python library <https://pypi.org/project/pdftotext/>`_, which is a python binding for the pdftotext tool.
+#. `pdftotext command line tool <https://en.wikipedia.org/wiki/Pdftotext>`_, which is based on `poppler <https://poppler.freedesktop.org/>`_, a PDF rendering library used by `xpdf <https://www.xpdfreader.com/>`_ and `Evince <https://wiki.gnome.org/Apps/Evince/FrequentlyAskedQuestions>`_ (the PDF reader of `Gnome <https://www.gnome.org/>`_).
+#. `pypdf <https://github.com/py-pdf/pypdf/>`_, which is one of the most common PDF lib for Python. pypdf is a pure-python solution, so it's very easy to install on all OSes.
+
+PyMuPDF and pdftotext both give a very good text output. So far, I can't say which one is best. pypdf often gives lower-quality text output, but its advantage is that it is a pure-Python librairy, so you will always be able to install it whatever your technical environnement is.
```
jans23: "environment"

account_edi_simple_pdf/README.rst
```diff
@@ -0,0 +1,212 @@
...
+  1. pymupdf
+  #. pdftotext.lib
+  #. pdftotext.cmd
+  #. pypdf
+
+In this configuration, Odoo will only use the selected text extraction method and, if it fails, it will display an error message.
+
+You will find a full demonstration about how to configure each Vendor and import the PDF invoices in this `screencast <https://www.youtube.com/watch?v=edsEuXVyEYE>`_.
+
+Bug Tracker
```
jans23, line 181: I suggest a short section like this:

```
Usage
=====

Go to Invoicing -> Vendors -> Bills, press the button Upload and upload a PDF file. Now the PDF file will be processed by account_edi_simple_pdf. In case your PDF file contains Factur-X data and you have Factur-X activated, that functionality will be executed instead.
```

account_edi_simple_pdf/README.rst
```diff
@@ -0,0 +1,212 @@
...
+
+
+.. code::
+
+  pip3 show regex
+
+Configuration
+=============
+
+By default, for the PDF to text conversion, the module tries the different methods in the order mentioned in the INSTALL section: it will first try to use **PyMuPDF**; if it fails (for example because the lib is not properly installed), then it will try to use the **pdftotext python lib**, if that one also fails, it will try to use **pdftotext command line** and, if it also fails, it will eventually try **pypdf**. If none of the 4 methods work, Odoo will display an error message.
```
jans23: I assume this try-and-error refers to the general availability of the different methods but not to their successful extraction on each individual PDF file. Would be good to clarify this.

pedrobaeza:
Another approach to try is #964 

The fetchmail/import gateway is very simple to be added.

jans23:
When trying the test upload within Contact's tab "Vendor Bill Import" I get the following error while when trying to import the same invoice within Invoicing -> Vendors -> Bills, Upload button I get useful messages such as "Missing invoice number format configuration on partner". For me this makes the test upload functionality useless.

```
RPC_ERROR
Odoo Server Error
  File "/srv/odoo/odoo/parts/odoo/odoo/addons/base/models/ir_http.py", line 242, in _dispatch
    result = request.dispatch()
  File "/srv/odoo/odoo/parts/odoo/odoo/http.py", line 702, in dispatch
    result = self._call_function(**self.params)
  File "/srv/odoo/odoo/parts/odoo/odoo/http.py", line 368, in _call_function
    return checked_call(self.db, *args, **kwargs)
  File "/srv/odoo/odoo/parts/odoo/odoo/service/model.py", line 94, in wrapper
    return f(dbname, *args, **kwargs)
  File "/srv/odoo/odoo/parts/odoo/odoo/http.py", line 357, in checked_call
    result = self.endpoint(*a, **kw)
  File "/srv/odoo/odoo/parts/odoo/odoo/http.py", line 925, in __call__
    return self.method(*args, **kw)
  File "/srv/odoo/odoo/parts/odoo/odoo/http.py", line 546, in response_wrap
    response = f(*args, **kw)
  File "/tmp/addons/web/controllers/main.py", line 1328, in call_button
    action = self._call_kw(model, method, args, kwargs)
  File "/tmp/addons/web/controllers/main.py", line 1316, in _call_kw
    return call_kw(request.env[model], method, args, kwargs)
  File "/srv/odoo/odoo/parts/odoo/odoo/api.py", line 471, in call_kw
    result = _call_kw_multi(method, model, args, kwargs)
  File "/srv/odoo/odoo/parts/odoo/odoo/api.py", line 456, in _call_kw_multi
    result = method(recs, *args, **kwargs)
  File "/tmp/addons/account_edi_simple_pdf/models/res_partner.py", line 194, in pdf_simple_test_run
    aiio = self.env["account.invoice.import"]
  File "/srv/odoo/odoo/parts/odoo/odoo/api.py", line 553, in __getitem__
    return self.registry[model_name]._browse(self, (), ())
  File "/srv/odoo/odoo/parts/odoo/odoo/modules/registry.py", line 182, in __getitem__
    return self.models[model_name]
Exception

The above exception was the direct cause of the following exception:

  File "/srv/odoo/odoo/parts/odoo/odoo/http.py", line 658, in _handle_exception
    return super(JsonRequest, self)._handle_exception(exception)
  File "/srv/odoo/odoo/parts/odoo/odoo/http.py", line 301, in _handle_exception
    raise exception.with_traceback(None) from new_cause
KeyError: 'account.invoice.import'
```

jans23:
account_invoice_import_simple_pdf allows to define Invoice Import Configurations. I don't find such option with this module. Where can I find it or is it obsolete? I want to define the vendor's account.

grafik

account_edi_simple_pdf/models/account_invoice_import_simple_pdf_mixin.py
```diff
@@ -0,0 +1,377 @@
...
+    import pdftotext
+except ImportError:
+    logger.debug("Cannot import pdftotext")
+try:
+    import pypdf
+except ImportError:
+    logger.debug("Cannot import pypdf")
+
+
+class AccountInvoiceImportSimplePdfMixin(models.AbstractModel):
```
jans23, line 33: WARNING odoo odoo.models: The model account.invoice.import.simple.pdf.mixin has no _description

hbrunn:
@jans23 thanks for your input. Note however that this is a draft and not really ready for review, even though it (mostly) works for my customer already the way they need it to. I mainly wanted to start a discussion about the future of invoice import in Odoo/OCA.

Regarding your question about import configurations: That's part of the `account_invoice_import` framework which is migrated in https://github.com/OCA/edi/pull/668 but stalled for quite a while. So won't happen here. I can however imagine adding a product to use for the one imported line per invoice on the partner form, to allow custom accounts/taxes per partner.

account_edi_simple_pdf/models/account_move.py
```diff
@@ -0,0 +1,74 @@
...
+            if result.partner_id.simple_pdf_product_id:
+                with Form(result) as invoice_form:
+                    with invoice_form.invoice_line_ids.edit(0) as line_form:
+                        line_form.product_id = result.partner_id.simple_pdf_product_id
+                        line_form.name = parsed_values.get(
+                            "description", line_form.name
+                        )
+                        line_form.price_unit = amount_untaxed or line_form.price_unit
+        for message in parsed_values.get("chatter_msg", []):
+            result.message_post(body=message)
```
fkantelberg: This makes problems if there was no `"partner"` in the `parsed_values` `result` would be an empty recordset which causes a singleton error in `mail/models/mail_thread.py", line 1854, in message_post`.

I would guess that the indentation might be wrong? What do you think?

hbrunn: thanks, fixed. Note however that I'll probably never finish this PR, as my customer commissioning this moved to v16 for which I made https://github.com/OCA/edi/pull/1091 where I'll implement the missing existing invoice amending (necessary also for mail parsing) some time soon.

PR #960, opened 2024-04-05
[14.0][IMP] edi_oca: Do not retry exchange_send indefinitely

florentx:
Sometimes the backend is badly configured, with invalid URL. Or there might be other conditions.
It happens more frequently on Dev or Test server instances.

In such case we don't want to insist sending Exchange Record indefinitely.

### DIFF
--- a/edi_oca/models/edi_backend.py
+++ b/edi_oca/models/edi_backend.py
@@ -8,6 +8,7 @@
 import base64
 import logging
 import traceback
+from datetime import timedelta
 from io import StringIO
 
 from odoo import _, exceptions, fields, models, tools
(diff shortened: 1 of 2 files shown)


ajaniszewska-dev (approved):

edi_oca/models/edi_backend.py
```diff
@@ -353,6 +364,13 @@ def _send_retryable_exceptions(self):
         # when dealing w/ internal or external systems or filesystems
         return (IOError, OSError)
 
+    def _send_should_retry(self, exchange_record):
+        return (
+            # Safety check not to retry indefinitely
+            exchange_record.create_date >= (fields.Datetime.now() - timedelta(days=1))
+            or self.env.context.get("_edi_send_break_on_error")
```
simahawk, line 371: why should retry when the ctx key is set? They key is there to let if fail directly.

SilvioC2C:
Superseded by #980

Issue #933, opened 2024-02-13
[15.0] edi_oca: support multicompany with edi.exchange.record


ThiagoMForgeFlow:
## Module
edi_oca

## Describe de bug
When you try to access the exchange records menus, a multi-company permissions error appears. If there are elements in the tree view whose `model` field contains record rules associated with multi-companies and the `res_id` field does not belong to the active companies, the error is displayed.

## To Reproduce
**Affected versions**: I see the error in 15.0

1. Have a multi-company record Rule for a model.
2. Generate edi.exchange.record for the record rule model, with different companies.
3. Access the exchange menu without having all the active companies

**Expected behavior**
When accessing the menu, only the elements that belong to the active companies should be displayed.

**Additional context**
To avoid this problem I have tried two solutions, but neither of them seem correct to me:

1. Filter so that only the acces_rule is checked with the records of the active companies `records = self.env[model].browse(rec_ids).with_user(self._uid).filtered(lambda x: x.sudo( ).company_id in self.env.companies.ids)`. This prevents the error from appearing but does not filter the elements that must be seen.
2. Create a company_id field from the res_id and model fields, https://github.com/OCA/edi/pull/932 but I have had problems in a real environment since there may be the case in which the loading order of the modules means that a model does not yet exist when trying to access him.

LoisRForgeFlow:
@simahawk @etobella your ideas would be much appreciated here

etobella:
Well, actually, the problem I see is another. We might need to filter edi records according to the rules of the related record.

Imagine you don't have access to employees and we create an edi exchange for employees, the same problem will arise.

WDYT?

PR #962, opened 2024-04-17
[12.0][FIX] edi_oca: avoid read error in exchange record tree view

MiquelRForgeFlow:
Let's not have errors when accessing tree view. If you later click in a non accessible record, then the error will show anyway. It's just avoid blocking accessing the EDI menu.

The good solution would be filtering the exchange records by related record, as said in https://github.com/OCA/edi/issues/933 discussion. But here we are speaking of v12, no need to be too picky in this version, right?


simahawk (commented):
don't we have `compute_sudo` in v12?

MiquelRForgeFlow:
Replaced by https://github.com/OCA/edi/pull/986.

PR #964, opened 2024-04-24
[15.0][ADD] base_import_pdf_by_template: New module

victoralmau:
New module

Example use case (import a PDF in Contacts)
ejemplo
ejemplo-import

Other use cases: Sale Orders, Purchase Orders, Invoices, etc

TODO:
- [x] Add documentation.
- [x] Add tests

@Tecnativa TT48213

PR #965, opened 2024-04-24
[15.0][WIP] base_import_pdf_simple_pdftotext: New module

victoralmau:
New module

Locked by:
- [ ] `base_import_pdf_simple`: https://github.com/OCA/edi/pull/964

@Tecnativa TT48213


victoralmau:
Finally we will not use it (`base_import_pdf_simple` will enough) although it can serve as an example to add any other extraction mode.

victoralmau:
Finally we will not use it (`base_import_pdf_simple` will enough) although it can serve as an example to add any other extraction mode.

PR #971, opened 2024-04-29
[14.0][FIX] edi_oca: fix read error on edi.exchange.record

ajaniszewska-dev:


simahawk (changes requested):
don't we have compute_sudo in v12?
Also, the original commit should be kept

edi_oca/models/edi_exchange_record.py
```diff
@@ -209,7 +209,7 @@ def record(self):
             return None
         if not self.model and self.parent_id:
             return self.parent_id.record
-        return self.env[self.model].browse(self.res_id).exists()
+        return self.env[self.model].browse(self.res_id).with_user(self._uid).filtered(lambda x: x.sudo().company_id in self.env.companies.ids).exists()
```
florentx: Better use `self.env.uid` than `self._uid`, since the latter form is deprecated.
But I am not even sure that we need `with_user` here.
```suggestion
        record = self.env[self.model].search([("id", "=", self.res_id), ("company_id", "in", self.env.companies.ids)])
        return record.with_user(self.env.uid)
```

or
```suggestion
        return self.env[self.model].search([("id", "=", self.res_id), ("company_id", "in", self.env.companies.ids)])
```

florentx (changes requested):

edi_oca/models/edi_exchange_record.py
```diff
@@ -209,7 +211,9 @@ def record(self):
             return None
         if not self.model and self.parent_id:
             return self.parent_id.record
-        return self.env[self.model].browse(self.res_id).exists()
+        return self.env[self.model].search(
+            [("id", "=", self.res_id), ("company_id", "in", self.env.companies.ids)]
+        )
```
florentx: Proposal to fix method when related model does not have "company_id" field:
```suggestion
        rec = self.env[self.model].browse(self.res_id).exists()
        if rec and "company_id" in rec and rec.sudo().company_id not in self.env.companies.ids:
            # Return empty recordset if company does not match
            rec = rec[:0]
        return rec
```

edi_oca/models/edi_exchange_record.py
```diff
@@ -209,7 +211,15 @@ def record(self):
...
             return self.parent_id.record
-        return self.env[self.model].browse(self.res_id).exists()
+        rec = self.env[self.model].browse(self.res_id).exists()
+        if (
+            rec
+            and "company_id" in rec
+            and rec.sudo().company_id.id not in self.env.companies.ids
+        ):
+            # Return empty recordset if company does not match
+            rec = rec[:0]
```
florentx: I wonder if we could miss another case: when the company_id is not set. Here latest proposal:
```suggestion
        rec = self.env[self.model].browse(self.res_id).exists()
        if rec and "company_id" in rec:
            rec_company = rec.sudo().company_id
            if rec_company and rec_company not in self.env.companies:
                # Return empty recordset if company does not match
                rec = rec[:0]
```

edi_oca/models/edi_exchange_record.py
```diff
@@ -209,7 +211,13 @@ def record(self):
             return None
         if not self.model and self.parent_id:
             return self.parent_id.record
-        return self.env[self.model].browse(self.res_id).exists()
+        rec = self.env[self.model].browse(self.res_id).exists()
+        if rec and "company_id" in rec:
+            rec_company = rec.sudo().company_id
+            if rec_company and rec_company not in self.env.companies:
+                # Return empty recordset if company does not match
+                rec = rec[:0]
```
grindtildeath:
```suggestion
                rec = self.env[self.model].browse()
```
IMO it's cleaner to show you're defining `rec` with an empty recordset.

grindtildeath (approved):
Only a small nitpicking

edi_oca/models/edi_exchange_record.py
```diff
@@ -209,7 +211,13 @@ def record(self):
...
         if not self.model and self.parent_id:
             return self.parent_id.record
-        return self.env[self.model].browse(self.res_id).exists()
+        rec = self.env[self.model].browse(self.res_id).exists()
+        if rec and "company_id" in rec:
+            rec_company = rec.sudo().company_id
+            if rec_company and rec_company not in self.env.companies:
+                # Return empty recordset if company does not match
+                rec = self.env[self.model].browse()
+        return rec
```
SilvioC2C: Just to understand: why not use a `search()` and let Odoo filter out the record if it doesn't match the proper multicompany security rules?
```suggestion
        return self.env[self.model].sudo(False).search([("id", "=", int(self.res_id))])
```

ajaniszewska-dev: @SilvioC2C do you think that there is something seriously breaking with current code? to me current implementation says for itself what we try to prevent here.

SilvioC2C: What I mean is: multicompany rules can be defined in different ways.
For example, we may have different rules that test the record's company against the current user's main company (`self.env.user.company_id`), user's allowed companies (`self.env.user.company_ids`), Env's company (`self.env.company`), session's currently enabled companies (`self.env.companies`), etc...

Using `search()` (with `sudo(False)`) will make Odoo test all these cases, while this current check only tests 1 case.

ajaniszewska-dev: @SilvioC2C  updated.

florentx (approved):

edi_oca/models/edi_exchange_record.py
```diff
@@ -237,7 +239,7 @@ def name_get(self):
         for rec in self:
             rec_name = rec.identifier
             if rec.res_id and rec.model:
-                rec_name = rec.record.display_name
+                rec_name = rec.sudo().record.display_name
```
SilvioC2C: Given the change introduced in how the `record` property is defined, I think `sudo()` should be postponed to `record`:
```suggestion
                rec_name = rec.record.sudo().display_name
```

edi_oca/models/edi_exchange_record.py
```diff
@@ -209,7 +211,7 @@ def record(self):
             return None
         if not self.model and self.parent_id:
             return self.parent_id.record
-        return self.env[self.model].browse(self.res_id).exists()
+        return self.env[self.model].sudo(False).search([("id", "=", int(self.res_id))])
```
simahawk, line 214: why sudo + false?

edi_oca/models/edi_exchange_record.py
```diff
@@ -237,7 +239,7 @@ def name_get(self):
         for rec in self:
             rec_name = rec.identifier
             if rec.res_id and rec.model:
-                rec_name = rec.record.display_name
+                rec_name = rec.record.sudo().display_name
```
simahawk, line 242: pls call sudo on self at the beginning of the loop (eg `for rec in self.sudo()`)

simahawk (changes requested):

edi_oca/models/edi_exchange_record.py
SilvioC2C: The initial purpose was to filter out the retrieved record if multi-company rules apply on the model, therefore I suggested the use of `sudo(False).search()` so that the ORM takes care of applying the security rules while executing the `search()`. `sudo(False)` is needed to ensure those rules are actually applied.

florentx (changes requested):
Please try instead:

- [ ] #974

This is the real cause of the issue.

PR #975, opened 2024-05-13
[14.0][ADD] edi_sending_warning

ajaniszewska-dev:
depends on: https://github.com/OCA/purchase-workflow/pull/2269


edi_sending_warning/models/edi_exchange_record.py
```diff
@@ -0,0 +1,18 @@
+# Copyright 2024 Camptocamp
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo.addons.component.core import Component
+
+
+class EdiExchangeRecordListener(Component):
+    _name = "edi.exchange.listener"
+    _inherit = ["base.connector.listener"]
```
florentx: Try this:
```suggestion
    _inherit = ["base.event.listener"]
```

Because this one is declared in module "component_event" (which is installed as dep of "edi_oca".

The other one (base.connector.listener) is part of "connector" module, which is not required.

florentx (changes requested):

edi_sending_warning/tests/test_edi_sending_warning.py
```diff
@@ -0,0 +1,49 @@
+import mock
+from freezegun import freeze_time
+from requests.exceptions import ConnectionError as ReqConnectionError
+
+from odoo.addons.edi_oca.tests.common import EDIBackendCommonTestCase
+from odoo.addons.queue_job.exception import RetryableJobError
+from odoo.addons.queue_job.tests.common import JobMixin
+
+
+class EdiExchangeSendingWarning(EDIBackendCommonTestCase, JobMixin):
```
ajaniszewska-dev, line 10: this test probably not the most accurate..

PR #978, opened 2024-05-14
[16.0][IMP] base_edifact: return code/barcode product even though not found in map2odoo_product

QuocDuong1306:
Context:
- It will check the corresponding products later https://github.com/OCA/edi/blob/23be3483e846ce16d443a8148c4c8709cd7691ab/sale_order_import/wizard/sale_order_import.py#L268

- And will help to know which code/barcode we can not find the corresponding products in https://github.com/OCA/edi/blob/23be3483e846ce16d443a8148c4c8709cd7691ab/base_business_document_import/models/business_document_import.py#L677


simahawk (approved):

QuocDuong1306:
I rebased

PR #979, opened 2024-05-29
[16.0][IMP] sale_order_import: better line error handling

QuocDuong1306:
Example:

image

### DIFF
--- a/sale_order_import/static/src/scss/error_lines_chatter_msg.scss
+++ b/sale_order_import/static/src/scss/error_lines_chatter_msg.scss
@@ -1,0 +1,6 @@
+div.error-line-message {
+    border-style: groove;
+    border-radius: 8px;
+    padding: 5px;
+    margin-bottom: 10px;
+}
--- a/sale_order_import/static/src/scss/error_lines_chatter_msg.scss
+++ b/sale_order_import/static/src/scss/error_lines_chatter_msg.scss
@@ -1,6 +1,0 @@
-div.error-line-message {
-    border-style: groove;
-    border-radius: 8px;
-    padding: 5px;
-    margin-bottom: 10px;
-}
--- a/sale_order_import/templates/error_lines_chatter_msg.xml
+++ b/sale_order_import/templates/error_lines_chatter_msg.xml
@@ -1,0 +1,42 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<odoo>
+    <template id="template_skip_error_lines_message">
+        <h4 class="text-danger">Errors lines on Import: <t
+                t-out="len(lines)"
+            /> line(s)</h4>
+        <t t-foreach="lines" t-as="line" t-key="line_index">
+            <div class="error-line-message">
+                <span>
+                    <strong>Error Message <t t-out="line_index + 1" /> :</strong>
+                    <span t-out="line['error_msg']" />
+                </span>
+                <br />
+                <span>
+                    <strong>Values <t t-out="line_index + 1" /> :</strong>
+                    <t t-set="values" t-value="line['values']" />
+                    <t
+                        t-call="sale_order_import.template_skip_error_line_value_message"
+                    />
+                </span>
+            </div>
+        </t>
+    </template>
+
+    <template id="template_skip_error_line_value_message">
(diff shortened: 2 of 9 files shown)


simahawk (changes requested):
While this is a good improvement I think we need to make one more step: opt in + better formatting.
IMO we should be able to decide if the lines have to be skipped or not especially because we are introducing this on existing instances. We need a flag that is false by default: `skip_error_lines`.

Then I would add a specific template to render instead of creating HTML ourselves and I would render each key properly w/ a list of `label: value`.

Reminder for commit message: always keep in mind that the main msg is a short summary of what you are changing and you should delegate the explanation to the description of the commit.

In this case it should be something like:

```
[imp] sale_order_import: better line error handling

Show errored lines in the chatter and ignore them on demand when importing.
```

QuocDuong1306:
Hi @simahawk , I updated the code, commit's message, and PR's description as well. Could you have a look?

santostelmo (approved):

sale_order_import/static/src/scss/error_lines_chatter_msg.scss
```diff
@@ -0,0 +1,6 @@
+div.error-line-message {
```
simahawk: we don't need these styles IMO if you use Bootstrap built-in utils

sale_order_import/templates/error_lines_chatter_msg.xml
```diff
@@ -0,0 +1,42 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<odoo>
+    <template id="template_skip_error_lines_message">
+        <h4 class="text-danger">Errors lines on Import: <t
+                t-out="len(lines)"
+            /> line(s)</h4>
+        <t t-foreach="lines" t-as="line" t-key="line_index">
+            <div class="error-line-message">
```
simahawk:
```suggestion
            <div class="alert alert-warning" role="alert">
```
please look for examples among the many in odoo code

sale_order_import/templates/error_lines_chatter_msg.xml
```diff
@@ -0,0 +1,42 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<odoo>
+    <template id="template_skip_error_lines_message">
+        <h4 class="text-danger">Errors lines on Import: <t
+                t-out="len(lines)"
+            /> line(s)</h4>
+        <t t-foreach="lines" t-as="line" t-key="line_index">
```
simahawk: What is `t-key`? Inside a loop you always have `line_index`.
https://www.odoo.com/documentation/17.0/developer/reference/frontend/qweb.html#loops

sale_order_import/templates/error_lines_chatter_msg.xml
```diff
@@ -0,0 +1,42 @@
...
+            <div class="error-line-message">
+                <span>
```
simahawk: use a P instead of a span + br

sale_order_import/templates/error_lines_chatter_msg.xml
```diff
@@ -0,0 +1,42 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<odoo>
+    <template id="template_skip_error_lines_message">
```
simahawk:
```suggestion
    <template id="skip_error_lines_message">
```
this is a template for sure

simahawk (changes requested):

QuocDuong1306:
Thank @simahawk for your suggestions, I added the fixup commit

I am not sure which one you prefer then? Could you please have a look and give me your opinion?

image
And 
image

simahawk:
> Thank @simahawk for your suggestions, I added the fixup commit
> 
> I am not sure which one you prefer then? Could you please have a look and give me your opinion?
> 
> image And image

The second one is better IMO.

cyrilmanuel:
@QuocDuong1306  is it possible to fix this one and redo a pr on the camptocamp fork please ?

thienvh332:
Superseded by new PR https://github.com/OCA/edi/pull/1293, cc @cyrilmanuel

PR #980, opened 2024-05-31
[14.0][IMP] edi_oca: Do not retry exchange_send indefinitely

SilvioC2C:
Sometimes the backend is badly configured, with invalid URL. Or there might be other conditions.
It happens more frequently on Dev or Test server instances.

In such case we don't want to insist sending Exchange Record indefinitely.

Supersedes #960

### DIFF
--- a/edi_oca/models/edi_backend.py
+++ b/edi_oca/models/edi_backend.py
@@ -364,12 +364,9 @@ class EDIBackend(models.Model):
         # when dealing w/ internal or external systems or filesystems
         return (IOError, OSError)
 
-    def _send_should_retry(self, exchange_record):
-        return (
-            # Safety check not to retry indefinitely
-            exchange_record.create_date >= (fields.Datetime.now() - timedelta(days=1))
-            or self.env.context.get("_edi_send_break_on_error")
-        )
+    def _send_should_retry(self, exc_record):
+        # Safety check not to retry indefinitely
+        return exc_record.create_date >= (fields.Datetime.now() - timedelta(days=1))
 
     def _output_check_send(self, exchange_record):
         if exchange_record.direction != "output":
--- a/edi_oca/models/edi_backend.py
+++ b/edi_oca/models/edi_backend.py
@@ -8,6 +8,7 @@
 import base64
 import logging
 import traceback
+from datetime import timedelta
 from io import StringIO
 
 from odoo import _, exceptions, fields, models, tools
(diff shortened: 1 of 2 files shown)


edi_oca/tests/test_backend_jobs.py
```diff
@@ -64,6 +65,38 @@ def test_output_fail_retry(self):
...
+        with mock.patch.object(type(self.backend), "_exchange_send") as mocked:
+            mocked.side_effect = ReqConnectionError("Connection broken")
+            with freeze_time("2024-01-10 11:00:00"):
+                # + 2 hours
+                job = self.backend.with_delay().exchange_send(record)
+                with self.assertRaises(RetryableJobError):
+                    job.perform()
+            with freeze_time("2024-01-11 08:50:00"):
+                # + 23 hours and 50 minutes
+                job = self.backend.with_delay().exchange_send(record)
```
simahawk, line 89: Why not using the same job you've got before?
However, if the identity_key works well, here you should get still the same job but I think is useless to call send again.

edi_oca/tests/test_backend_jobs.py
```diff
@@ -64,6 +65,38 @@ def test_output_fail_retry(self):
...
+                with self.assertRaises(RetryableJobError):
+                    job.perform()
+            self.assertEqual(record.edi_exchange_state, "output_pending")
+            with freeze_time("2024-01-11 09:20:00"):
+                # + 24 hours and 20 minutes
+                job = self.backend.with_delay().exchange_send(record)
+                res = job.perform()
+            self.assertIn("Error", res)
+        job_counter.search_created()
```
simahawk, line 98: this search seems useless as you don't use the result

edi_oca/models/edi_backend.py
```diff
@@ -301,10 +302,20 @@ def exchange_send(self, exchange_record):
             _logger.debug("%s sent", exchange_record.identifier)
         except self._send_retryable_exceptions() as err:
             error = _get_exception_msg()
-            _logger.debug("%s send failed. To be retried.", exchange_record.identifier)
-            raise RetryableJobError(
-                error, **exchange_record._job_retry_params()
-            ) from err
+            if self._send_should_retry(exchange_record):
```
simahawk, line 305: Thinking out loud...
In theory a case where we spawn another job for the same record should not exist thanks to the `identity_key`.
Hence, I assume that jobs won't be retried infinitely if we reach the max retry job count.
If this does not happen it means the identity_key is not working and we end up w/ more than one job to send the same record.
Am I wrong?

edi_oca/tests/test_backend_jobs.py
simahawk: @SilvioC2C ping

edi_oca/models/edi_backend.py
simahawk: @SilvioC2C ping

PR #990, opened 2024-06-10
[16.0][IMP] base_edifact: improve method map2odoo_product to take pia segment into account

trisdoan:
### This change
- Improve `map2odoo_product()` to process pia segment (if provided). The codes is later used to find products in [`business.document.import:_match_product_search() `](https://github.com/OCA/edi/blob/9cd45f58ace990702b5a816f166e51943850f46a/base_business_document_import/models/business_document_import.py#L693)
- Before this change, users cannot import sale order when barcode and product code are both provided in edifact file and product,in Odoo, does not contain barcode.

### Result
1. Import with non-existing code
user_error

3. Import with skipping error
chatter_log

### DIFF
--- a/base_edifact/README.rst
+++ b/base_edifact/README.rst
@@ -104,6 +104,7 @@ Contributors
 * Rafa Morant <rmorant@albasoft.com> (www.albasoft.com)
 * Marc Poch <mpoch@planetatic.com>
 * Duong (Tran Quoc) <duongtq@trobz.com>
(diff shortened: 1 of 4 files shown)


santostelmo (approved):

cyrilmanuel (approved):

cyrilmanuel:
Hiii @simahawk all good for you ?

PR #991, opened 2024-06-10
[ADD] EDIWebserviceSendHTTPException to catch http error form edi_web…

barkat-matthias:
Add a new Exeception EDIWebserviceSendHTTPException inheriting from requests.error.HTTPError.

Catching request.error.HTTPError in EDIWebserviceSend.send to raise EDIWebserviceSendHTTPException instead

### DIFF
--- a/edi_webservice_oca/components/send.py
+++ b/edi_webservice_oca/components/send.py
@@ -2,9 +2,14 @@
 # @author: Simone Orsi <simahawk@gmail.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
+
+from requests.exceptions import HTTPError
+
 from odoo import _, exceptions
 
 from odoo.addons.component.core import Component
+
+from ..exceptions import EDIWebserviceSendHTTPException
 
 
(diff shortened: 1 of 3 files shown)


edi_webservice_oca/components/send.py
```diff
@@ -27,7 +32,10 @@ def __init__(self, work_context):
 
     def send(self):
         method, pargs, kwargs = self._get_call_params()
-        return self.webservice_backend.call(method, *pargs, **kwargs)
+        try:
+            return self.webservice_backend.call(method, *pargs, **kwargs)
+        except HTTPError as err:
+            raise EDIWebserviceSendHTTPException("EDI HTTP Error: %s" % err) from err
```
simahawk, line 38: We already know that this is a specific exception. I would simply use `EDIWebserviceSendHTTPException(str(err))`.
Also, you should be able to find the original exception in the catched exception via `__cause__`

simahawk (changes requested):
Could you tell more about the purpose? I guess you want to catch something specific in the framework?

PR #996, opened 2024-06-14
[12.0][IMP] edi_oca: don't consider exchange records if jobs have been created

MiquelRForgeFlow:
If two `_check_output_exchange_sync` are called consecutively, one after the other, maybe due to `_execute_next_action` method, the following error happens:
Selection_3459
Both `_check_output_exchange_sync` calls detect the same exchange records, and both create same jobs. The first batch of jobs will be processed but the second ones will fail due to check of state = "new".

### DIFF
--- a/edi_oca/models/__init__.py
+++ b/edi_oca/models/__init__.py
@@ -5,3 +5,4 @@ from . import edi_exchange_consumer_mixin
 from . import edi_exchange_type
 from . import edi_exchange_type_rule
 from . import edi_id_mixin
+from . import queue_job
--- a/edi_oca/models/edi_backend.py
+++ b/edi_oca/models/edi_backend.py
@@ -414,6 +414,7 @@ class EDIBackend(models.Model):
             len(new_records),
         )
         for rec in new_records:
+            rec.is_queued = True
             job1 = rec.delayable().action_exchange_generate()
             if not skip_send:
                 # Chain send job.
@@ -447,6 +448,7 @@ class EDIBackend(models.Model):
             ("type_id.direction", "=", "output"),
             ("edi_exchange_state", "=", "new"),
             ("exchange_file", "=", False),
+            ("is_queued", "=", False),
         ]
         if record_ids:
             domain.append(("id", "in", record_ids))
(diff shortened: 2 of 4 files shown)


MiquelRForgeFlow:
@simahawk what's your opinion on this?

simahawk (changes requested):
We should avoid having this kind of relations w/ jobs.
Please see https://github.com/OCA/queue/pull/573

As for the solution on how to prevent the issue to happen... IMO we should rely on
 the id key https://github.com/OCA/edi-framework/blob/16.0/edi_oca/models/edi_exchange_record.py#L591

And we should leverage the job chaining for inbound records too to speed up processing.

FYI there's a way to find jobs for a record https://github.com/OCA/edi-framework/pull/93

Regarding the error itself: an alternative would be to simply let the job fail gracefully (meaning set it as done). 

BTW I was thinking that we probably need a specific state for when a record is waiting for a job... not sure tho.

MiquelRForgeFlow:
@simahawk is it acceptable now? It became simpler.

MiquelRForgeFlow:
@simahawk 

> Regarding the error itself: an alternative would be to simply let the job fail gracefully (meaning set it as done).

Yeah, but silencing possible errors it's bad coding. I prefer avoiding from happening the specific detected ones, and letting the open door to detect other errors.

> BTW I was thinking that we probably need a specific state for when a record is waiting for a job... not sure tho.

A new state between "new" and the others, for example "queued" would be nice, but I am afraid that it's complicated, as we should review all the flows and adapt code. I prefer the boolean solution, it's simpler.

MiquelRForgeFlow:
@simahawk could you take a look?

MiquelRForgeFlow:
@simahawk could you take a look? It already has the stale tag...

simahawk:
@MiquelRForgeFlow I've recently thought about this and I IMO the issue is somewhere else.

If you are using `quick_exec` you should turn off `exchange_file_auto_generate`  and if you do so, the cron will ignore those jobs. 

From my experience the issue comes from there.
Also, I think that nowadays everybody are using `quick_exec` and less and less (if not not at all) relying on the cron to process outgoing records.

I even wonder if it makes sense to keep the cron now that we have  the quick exec + chained jobs generate/send.

PR #1004, opened 2024-07-10
[16.0][IMP] base_edi: introduce a basic edi user group

QuocDuong1306:

PR #1005, opened 2024-07-10
[15.0][IMP] base_edi: introduce a basic edi user group (Backport from 16.0)

QuocDuong1306:
A suggestion from the author (of the module) @simahawk 
- https://github.com/OCA/edi/pull/1004#issuecomment-2219596625

### DIFF
--- a/base_edi/security/edi_groups.xml
+++ b/base_edi/security/edi_groups.xml
@@ -1,5 +1,15 @@
 <?xml version="1.0" encoding="utf-8" ?>
 <odoo>
+    <record id="group_edi_user" model="res.groups">
+        <field name="name">EDI User</field>
+        <field name="category_id" ref="module_category_edi" />
+        <field
+            name="users"
+            eval="[(4, ref('base.user_root')), (4, ref('base.user_admin'))]"
+        />
+        <field name="implied_ids" eval="[(4, ref('base.group_user'))]" />
+    </record>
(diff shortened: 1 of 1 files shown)


pedrobaeza:
`[BKP]` doesn't exit as tag, as why not cherry-picking the commit from 16.0? You are removing the attribution.

QuocDuong1306:
Hi @pedrobaeza , I have a suggestion from @simahawk ([here](https://github.com/OCA/edi/pull/1004#issuecomment-2219596625))

Btw, in many repositories, I saw a `[BKP]` tag is used for this case (Backport), Did I miss sth, could you suggest it?

pedrobaeza:
OK about the attribution. I thought it was from other author. It's good that you link the original pull request for avoiding these misunderstandings.

About the commit message, it's not correct although some use them: https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message

QuocDuong1306:
This commit was merged on 16.0. Do we need to change the commit message? If yes, which message do you suggest for this one (It will be helpful for me to contribute to OCA at the other times also)

pedrobaeza:
I think the only problem is to put `[BKP]` in the PR title. The commit message already contains `[IMP]`.

simahawk (approved):

PR #1023, opened 2024-08-12
 [14.0][BPRT] edi_sale_oca: Backport from v16.0

thienvh332:
### Description:
- Backports the **edi_sale_oca** module to the 14.0 branch.

### Context:
- This backport ensures compatibility with older versions of Odoo while taking advantage of refactoring performed during the migration to version 14.0.

### Changes:

- Backported the **edi_sale_oca** module from version **16.0** to **14.0**.


edi_sale_oca/tests/common.py
```diff
@@ -0,0 +1,50 @@
...
+        vals = dict(commitment_date=fields.Date.today())
+        vals.update(kw)
+        so_vals = model.play_onchanges(vals, [])
+        if "order_line" in so_vals:
+            so_vals["order_line"] = [(0, 0, x) for x in vals["order_line"]]
+        return model.create(so_vals)
+
+    @classmethod
+    def _setup_order(cls, **kw):
+        cls.product_a = cls.env.ref("product.product_product_4")
```
mmequignon, line 25: Please, create products instead of using odoo demo ones, we want to have control about their configuration.
It also help with understanding during reviews, no one can guess what's their `sale_delay` for instance, without having to check odoo demo data.

simahawk: We can do this in a 2nd iteration

simahawk (approved):

PR #1033, opened 2024-09-02
[15.0][FIX] base_import_pdf_by_template: Improve the import process to account for fixed values

victoralmau:
Improve the import process to account for fixed values

When fixed values are defined, _default_x_ was used in the context of Form() but in some cases this was not respected (for example if the value was changed in an onchange, `warehouse_id` field of sale orders).
We changed the way of creating the records to always take into account the fixed values.

Please @pedrobaeza and @pilarvargas-tecnativa can you review it?

@Tecnativa TT50650

### DIFF
--- a/base_import_pdf_by_template/wizards/wizard_base_import_pdf_upload.py
+++ b/base_import_pdf_by_template/wizards/wizard_base_import_pdf_upload.py
@@ -171,7 +171,6 @@ class WizardBaseImportPdfUploadLine(models.TransientModel):
 
     def _process_form(self):
         """Create record with Form() according to text."""
-        # text = "".join(data)
         text = self.data
         template = self.template_id
         model = self.env[template.model]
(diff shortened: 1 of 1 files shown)


pilarvargas-tecnativa (approved):
:+1:

base_import_pdf_by_template/wizards/wizard_base_import_pdf_upload.py
```diff
@@ -202,7 +201,15 @@ def _process_form(self):
...
         try:
-            record = model_form.save()
+            # Prepare vals (similar to .save()) + apply defaults (in case it has changed
+            # in some onchange for example: warehouse_id from sale orders)
+            vals = model_form._values_to_save()
+            for key in list(ctx.keys()):
+                if key.startswith("default_"):
+                    real_key = key.replace("default_", "")
+                    if real_key in vals:
+                        vals.update({real_key: ctx[key]})
```
pedrobaeza: Some improvements for code readability:

```suggestion
            for key in ctx:
                if key.startswith("default_"):
                    field = key.replace("default_", "")
                    if field in vals:
                        vals.update({field: ctx[key]})
```

pedrobaeza (approved):
/ocabot merge patch

PR #1024, opened 2024-08-12
 [14.0][ADD] edi_sale_ubl_oca

thienvh332:
Depends on:
- [x] #1035 
- [x] #1067 

### Description:
- Updates the **edi_sale_ubl_oca** module by adding a necessary dependency in the 14.0 branch.

### Context:
- Previously, **edi_sale_ubl_oca** depended on **edi_sale_oca**, which included the **edi_exchange_type_auto** module. Since **edi_exchange_type_auto** has been removed from **edi_sale_oca**, this dependency is added explicitly to ensure continued functionality.

### Changes:

- Added **edi_exchange_type_auto** as a dependency to the **edi_sale_ubl_oca** module.

PR #1035, opened 2024-09-06
[14.0][IMP] edi_oca: Add new model edi.configuration

thienvh332:
Real life examples:

- [ ] #1067 
- [ ] #1024 
- [ ] purchase soon

# Specs

In edi_oca

Add generic `edi.configuration` model with these characteristics:

**Fields**

    active
    description: describe what the conf is for
    backend_id
    type_id
  
    model: model the conf applies to
  
    trigger_id: m2o field to make easier and more explicit the use of the conf for a specific event (eg: on_post_account_move, on_so_confirm, etc)
        default option: empty
  
    snippet_before_do: validate the state, used to collect records todo
        can be used to do anything before the action that triggered it
      
        eval ctx:
            trigger = create/write/$name_of_the_event
            conf = current conf
  
        the snippet can return:
            todo = True/False
            snippet_do_vars = {}  # set of variables to pass over to the next snippet
            event_only
            tracked_fields
            edi_action
            .. in fact, any other variable that might be needed
  
    snippet_do: used to do something specific here
  
        receives: operation, edi_action, vals, old_vals

This model can then be used by conf consumers.
**Other methods:**

    edi_exec_snippet_before_do(self, record):
      exec before snippet, self=conf
    
    edi_exec_snippet_do(self, record):
      exec snippet, self=conf
    
    edi_get_conf(self, trigger, model_name=None, partners=None, backend=None):
      filter current recordset based on conditions

**Backend / Exc. record**

    In `edi.backend._check_output_exchange_sync` move the loop on new_records to a new method `exchange_generate_send` whereas if the file has to be generated
    2 jobs are chained (generate + send) if only send is needed, delay only one job.
    
Then add edi.exchange.record.action_exchange_generate_send and use that method from backend.
Consumer mixin

Specific fields for configuration (eg: edi_purchase_conf_ids) will be added by glue modules on res.partner.

**Default conf**

    add generic conf for send_via_email
    
    code: send_via_email
    trigger: on_email_send
    snippet: `record._edi_send_via_email()
    
    add generic conf for send_via_edi
    
    code: send_via_edi
    trigger: none
    snippet: `record._edi_send_via_edi(conf.type_id)

**`NOTE:`** sending emails is not really bound to edi
but in real life it's handy to have a default way to send EDI docs via email.
You could even generate an exchange record to keep track of that
but the choice will be up to the implementers.
The key is that w/ this features they can do whatever they want.

**Views**

    add a specific menu item in edi settings to see all confs
    make them searchable by model, active/inactive, partner, exc type, trigger

**On edi_purchase_oca**

    add conf trigger on_button_confirm_purchase_order
    
    add res.partner.edi_purchase_conf_ids
    domain model=purchase.order, default_model=purchase.order, default_res_id=active_id
    
    (document this approach in edi_oca readme section for edi.conf)
    
    add generic conf for send_via_email_rfq
    
    code: send_via_email_rfq
    snippet: `record._edi_send_via_email(ir_action=record.action_rfq_send())

Example of specific listener implementation:

    class EDIConfigPurchaseListener(Component):
        _name = "edi.config.consumer.listener.purchase.order"
        _inherit = "base.event.listener"
    
    def on_button_confirm_purchase_order(self, record):
        trigger = "on_button_confirm_purchase_order"
        confs = record.edi_purchase_conf_ids.edi_get_conf(trigger)
        for conf in confs:
            conf.edi_exec_snippet_do(record)

EDIT: additional model `edi.configuration.trigger` to ease UI mgmt.


etobella:
More information would be awesome in order to understand why you did  this.

simahawk:
@etobella this origins from our discussion on https://github.com/OCA/edi-framework/pull/39
@thienvh332 I'm adding the specs I gave you in the description feel free to adapt / add more things.
IMO we should split the purchase part when ready (let's say before switching from draft to real PR to ease the final review of both parts).

simahawk:
@thienvh332 as it's draft, do you want us to wait for a review?

nilshamerlinck:
hi @simahawk
- yes, please wait before reviewing, we need to finalize the internal review on our side 
- now that you confirmed you want this in v14, we'll complete it asap

thienvh332:
Hello @simahawk ,
This PR is ready for review. Could you please take a look at it?

simahawk:
can you drop the pypdf fix? https://github.com/OCA/edi/pull/1035/commits/15a994eb8a60d01d05202fbd637f04df81b0cef6

edi_oca/models/edi_exchange_consumer_mixin.py
```diff
@@ -291,3 +291,45 @@ def _edi_set_origin(self, exc_record):
...
+        msg = _("EDI auto: output generated.")
+        exchange_record._notify_related_record(msg)
+        exchange_record._trigger_edi_event("generated")
+
+    def _edi_send_via_email(
+        self, ir_action=None, subtype_ref=None, partner_method=None, partners=None
+    ):
+        # Default action if not provided
+        if ir_action is None:
+            # `action_send_email` is just an action name I created
```
simahawk: I don't get what's the need for this ... there's no `action_send_email` anywhere.

edi_oca/models/edi_configuration.py
```diff
@@ -0,0 +1,222 @@
...
+        It allows users to create their own configurations, which can be tailored
+        to meet the specific needs of their business processes.
+    """
+
+    name = fields.Char(string="Name", required=True)
+    active = fields.Boolean(default=True)
+    code = fields.Char(required=True, copy=False, index=True, unique=True)
+    description = fields.Char(help="Describe what the conf is for")
+    backend_id = fields.Many2one(string="Backend", comodel_name="edi.backend")
+    type_id = fields.Many2one(
```
simahawk, line 41: type should be mandatory

edi_oca/models/edi_configuration.py
```diff
@@ -0,0 +1,222 @@
...
+        auto_join=True,
+        index=True,
+    )
+    model = fields.Many2one(
+        "ir.model",
+        string="Model",
+        help="Model the conf applies to. Leave blank to apply for all models",
+    )
+    model_name = fields.Char(related="model.model", store=True)
+    partner_ids = fields.Many2many(
```
simahawk: There's no need for a partner_ids field. The conf should be stored on the partner on a specific field per app or module (eg: `edi_purchase_conf_ids`).
The goal is to have clarity on UI on what confs are used for.
In fact, it's not in the specs.

edi_oca/models/edi_configuration.py
```diff
@@ -0,0 +1,222 @@
...
+                if rec.type_id.backend_type_id != rec.backend_id.backend_type_id:
+                    raise exceptions.ValidationError(
+                        _("Backend type must match with exchange type's backend type!")
+                    )
+
+    @api.model
+    def default_get(self, fields):
+        vals = super().default_get(fields)
+        model = self.env.context.get("default_model_name", False)
+        partners = self.env.context.get("default_partner_ids", False)
```
simahawk: I'd choose which model field you want to be the master. If the model name is only used for searches, we don't need this.
Partners as well, can be trashed here.

edi_oca/models/edi_configuration.py
```diff
@@ -0,0 +1,222 @@
+# Copyright 2024 Camptocamp SA
```
simahawk, line 1: This comes from the auto-type stuff, can you keep the authorship here?

edi_oca/models/edi_configuration.py
```diff
@@ -0,0 +1,222 @@
...
+        vals = super().default_get(fields)
+        model = self.env.context.get("default_model_name", False)
+        partners = self.env.context.get("default_partner_ids", False)
+        if model:
+            vals["model"] = self.env["ir.model"]._get_id(model)
+        if partners and isinstance(partners, list):
+            vals["partner_ids"] = [(6, 0, partners)]
+        return vals
+
+    def _code_snippet_valued(self, snippet):
```
simahawk, line 89: this stuff is copied from edi_exchange_template and the type auto module. Can you please add a TODO to move it to some utils and/or mixin?

edi_oca/models/edi_configuration.py
```diff
@@ -0,0 +1,222 @@
...
+        )
+        return vals
+
+    def edi_exec_snippet_do(self, record, **kwargs):
+        self.ensure_one()
+
+        old_value = kwargs.get("old_vals", {}).get(record.id, {})
+        new_value = kwargs.get("vals", {}).get(record.id, {})
+        vals = {
+            "todo": kwargs.get("todo", True),
```
simahawk: why do you expect to receive a `todo` flag here?

edi_oca/models/edi_configuration.py
```diff
@@ -0,0 +1,222 @@
...
+            before_do_vals = self.edi_exec_snippet_before_do(record, **kwargs)
+            vals.update(before_do_vals)
+        if vals["todo"]:
+            return self._evaluate_code_snippet(self.snippet_do, **vals)
+        return True
+
+    @api.model
+    def edi_get_conf(self, trigger, model_name=None, partners=None, backend=None):
+        domain = [("trigger", "=", trigger)]
+        if model_name:
```
simahawk: Just accept the record and retrieve its name from it.
In fact, you should filter only the configurations that are assigned to the current partner.

I re-write here what has been done for PO to give you an example. 
```
    def on_button_confirm_purchase_order(self, records):
        trigger = "on_button_confirm_purchase_order"
        confs = record.edi_purchase_conf_ids.edi_get_conf(trigger)
        for conf in confs:
             conf.edi_exec_snippet_do(record)
```
You see? You always have the recordset of configurations set on the partner. You just have to filter them by trigger.
In fact, this is what I described in the specs :wink: 
And the docstring I provided is "filter current recordset of conf using trigger".

edi_oca/models/edi_configuration.py
simahawk: The license must be LGPL

edi_oca/components/base_listener_config.py
```diff
@@ -0,0 +1,33 @@
+# Copyright 2024 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo.addons.component.core import Component
+
+
+class EDIBackendListenerComponentConfig(Component):
```
simahawk: this listener is useless IMO. See below.

edi_oca/models/edi_configuration_mixin.py
```diff
@@ -0,0 +1,81 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+import logging
+
+from odoo import api, models
+
+_logger = logging.getLogger(__name__)
+
+
+class EDIConfigurationMixin(models.AbstractModel):
```
simahawk: I don't understand why we need this mixin. Is not in the specs... :/
Looks like a not required copy/past from `*_auto` module.

I think there's a big misunderstanding on WHO is supposed to trigger the specific action.
I wanted to simplify what I've done on `*_auto` module getting rid of all this automation and leave the control to the specific module and to the configuration that you can define.

Concept:
1. the trigger is given by the specific module -> `on_button_confirm_purchase_order` must be an event triggered by the module `edi_purchase_oca` when `button_confirm` is called (which is already the case!)
2. the condition to skip or exec a conf is defined on the conf itself -> that's why I described 2 methods (before_do and do)

edi_oca/tests/fake_models.py
```diff
@@ -7,10 +7,11 @@
 
 class EdiExchangeConsumerTest(models.Model):
     _name = "edi.exchange.consumer.test"
-    _inherit = ["edi.exchange.consumer.mixin"]
+    _inherit = ["edi.exchange.consumer.mixin", "edi.configuration.mixin"]
```
simahawk: you can rollback this

edi_oca/tests/fake_models.py
```diff
@@ -7,10 +7,11 @@
...
     _description = "Model used only for test"
 
     name = fields.Char()
+    partner_id = fields.Many2one("res.partner")
```
simahawk: why a partner_id? If there's no partner on the model we should not run anything or even break the init of the model if you are assigning configurations to a model w/o partners

edi_purchase_oca/components/listener_purchase_order.py
```diff
@@ -0,0 +1,19 @@
...
+
+class EDIConfigPurchaseListener(Component):
+    _name = "edi.listener.config.purchase.order"
+    _inherit = "edi.component.listener.config"
+    _apply_on = ["purchase.order"]
+
+    def on_button_confirm_purchase_order(self, record):
+        trigger = "on_button_confirm_purchase_order"
+        for rec in record:
+            confs = self.env["edi.configuration"].edi_get_conf(
```
simahawk: you are not using `edi_purchase_conf_ids`

simahawk (changes requested):

thienvh332:
Hi @simahawk ,
I have updated the PR but I have 2 things I need your opinion on. Could you give me your views?
> I don't get what's the need for this ... there's no `action_send_email` anywhere.

- We have a common config **`send_via_email`**. There we don't know what **`ir_action`** is used on each model. So I created it as a common path for other models. They can use **`_edi_send_via_email`** without passing **`ir_action`** parameter.
- Of course, if you feel it is inappropriate I will delete it.

> type should be mandatory

- About **`type_id`** must be mandatory:
  - I'm confused about this. There is a requirement to create 2 common confs (**`send_via_email`** and **`send_via_edi`**). If **`type_id`** is required then we will:
      - Create data for **`edi_exchange_type`** and related models in **`edi_oca`** module.
      - Create those **`2 common confs`** in **`another module`** to avoid creating unwanted data when installing the **`edi_oca`** module.

edi_oca/models/edi_backend.py
```diff
@@ -377,6 +377,18 @@ def _cron_check_output_exchange_sync(self, **kw):
         for backend in self:
             backend._check_output_exchange_sync(**kw)
 
+    def exchange_generate_send(self, recordset, skip_generate=False, skip_send=False):
+        for rec in recordset:
+            if skip_generate:
+                job1 = rec
```
simahawk: what would be delayed in this case if we skip_send too? :thinking:

edi_oca/models/edi_configuration.py
```diff
@@ -0,0 +1,193 @@
...
+    description = fields.Char(help="Describe what the conf is for")
+    backend_id = fields.Many2one(string="Backend", comodel_name="edi.backend")
+    type_id = fields.Many2one(
+        string="Exchange Type",
+        comodel_name="edi.exchange.type",
+        ondelete="cascade",
+        auto_join=True,
+        index=True,
+    )
+    model = fields.Many2one(
```
simahawk:
```suggestion
    model_id = fields.Many2one(
```

edi_oca/models/edi_configuration.py
```diff
@@ -0,0 +1,193 @@
...
+        "ir.model",
+        string="Model",
+        help="Model the conf applies to. Leave blank to apply for all models",
+    )
+    model_name = fields.Char(related="model.model", store=True)
+    trigger = fields.Selection(
+        [
+            ("on_record_write", "Update Record"),
+            ("on_record_create", "Create Record"),
+            ("on_email_send", "Send Email"),
```
simahawk: I guess this is to ease the configuration for models like SO, PO, INV, that can be sent by email.
As there's no specific implementation here, I would add a comment that explain why this is here.

Also, I think `trigger` should be mandatory otherwise the conf is useless, no?
Maybe we can add another option like `("disabled", "Disabled")` which would allow to keep the conf visible but disabled.

edi_oca/models/edi_configuration.py
```diff
@@ -0,0 +1,193 @@
...
+        if self.snippet_before_do:
+            before_do_vals = self.edi_exec_snippet_before_do(record, **kwargs)
+            vals.update(before_do_vals)
+        if vals["todo"]:
+            return self._evaluate_code_snippet(self.snippet_do, **vals)
+        return True
+
+    def edi_get_conf(self, trigger, backend=None):
+        domain = [("trigger", "=", trigger)]
+        backend_ids = self.mapped("type_id.backend_id.id")
```
simahawk: what's the point on filtering on all the backends assigned to the current recordset?

simahawk:
> Hi @simahawk , I have updated the PR but I have 2 things I need your opinion on. Could you give me your views?
> 
> > I don't get what's the need for this ... there's no `action_send_email` anywhere.
> 
>     * We have a common config **`send_via_email`**. There we don't know what **`ir_action`** is used on each model. So I created it as a common path for other models. They can use **`_edi_send_via_email`** without passing **`ir_action`** parameter.
> 
>     * Of course, if you feel it is inappropriate I will delete it.

The idea is that everything should be explicit. You must explicitly declare which action you want to use.

> > type should be mandatory
> 
>     * About **`type_id`** must be mandatory:
>       
>       * I'm confused about this. There is a requirement to create 2 common confs (**`send_via_email`** and **`send_via_edi`**). If **`type_id`** is required then we will:
>         
>         * Create data for **`edi_exchange_type`** and related models in **`edi_oca`** module.
>         * Create those **`2 common confs`** in **`another module`** to avoid creating unwanted data when installing the **`edi_oca`** module.

Ok, you are right. Cannot be mandatory. Good point :)

edi_oca/models/edi_configuration.py
simahawk: no, as per our comments.
Please add a comment here explaining why it is not mandatory ;)

edi_oca/models/edi_exchange_consumer_mixin.py
```diff
@@ -291,3 +291,45 @@ def _edi_set_origin(self, exc_record):
     def _edi_get_origin(self):
         self.ensure_one()
         return self.origin_exchange_record_id
+
+    def _edi_send_via_edi(self, exchange_type):
+        exchange_record = self._edi_create_exchange_record(exchange_type)
+        exchange_record.action_exchange_generate_send()
+        msg = _("EDI auto: output generated.")
```
simahawk: This is not only generated but generated and sent.
I wonder if we need this custom message...
The backend will call `notify_action_complete` when is done...

edi_purchase_oca/components/listener_purchase_order.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+class EDIConfigPurchaseListener(Component):
+    _name = "edi.listener.config.purchase.order"
+    _inherit = "base.event.listener"
+    _apply_on = ["purchase.order"]
+
+    def on_button_confirm_purchase_order(self, record):
+        trigger = "on_button_confirm_purchase_order"
+        confs = record.mapped("partner_id.edi_purchase_conf_ids").edi_get_conf(trigger)
```
simahawk: no need to use mapped

edi_purchase_oca/data/edi_configuration.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="edi_conf_button_confirm_purchase_order" model="edi.configuration">
```
simahawk: this is not needed

edi_purchase_oca/models/edi_configuration.py
```diff
@@ -0,0 +1,19 @@
...
+from odoo import fields, models
+
+
+class EdiConfiguration(models.Model):
+    _inherit = "edi.configuration"
+
+    trigger = fields.Selection(
+        selection_add=[
+            ("on_button_confirm_purchase_order", "On Button Confirm Purchase Order"),
+            ("send_via_email_rfq", "Send via Email RFQ"),
```
simahawk:
```suggestion
            ("send_via_email_rfq", "Send via Email RFQ"),
```
To avoid confusion and duplication we should probably have this in the base module but call it `on_send_via_email`.

simahawk:
@etobella would you have time for a review?

edi_purchase_oca/models/res_partner.py
```diff
@@ -0,0 +1,18 @@
...
+from odoo import fields, models
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    edi_purchase_conf_ids = fields.Many2many(
+        string="EDI Purchase Config Ids",
+        comodel_name="edi.configuration",
+        relation="res_partner_edi_configuration_rel",
```
simahawk:
```suggestion
        relation="res_partner_edi_purchase_configuration_rel",
```

edi_purchase_oca/models/res_partner.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+from odoo import fields, models
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    edi_purchase_conf_ids = fields.Many2many(
+        string="EDI Purchase Config Ids",
```
simahawk:
```suggestion
        string="EDI Purchase configuration",
```

simahawk:
Here's a full real life example https://github.com/OCA/edi/pull/1067
@thienvh332 pls have a look.

TODO here:
- [ ] drop the purchase part and move it to its own PR w/ adaptation based on the sale example

simahawk:
@thienvh332 as I want to speed up this and the modules that depend on it I'm taking this over.
I will:
- do some cleanups
- split purchase part
- add a new model edi.configuration.trigger to use m2o and be able to filter in the UI for only relevant triggers, otherwise you'll see ALL the triggers for all specific modules everytime you configure something

thienvh332:
Hi @simahawk ,
Ok for me, thanks for the help

simahawk:
@etobella would you have time for a review?

edi_oca/models/edi_backend.py
```diff
@@ -398,6 +398,19 @@ def _cron_check_output_exchange_sync(self, **kw):
...
+            else:
+                job1 = rec.delayable().action_exchange_generate()
+            if hasattr(job1, "on_done"):
```
etobella: Why do you need the on_done check?

thienvh332: I want to check if my record is not a job then it will skip delay

simahawk: maybe add a comment?

simahawk: I will :)

simahawk: I refactored here [`66797c2` (#1035)](https://github.com/OCA/edi/pull/1035/commits/66797c2a1c0607da7a60fe05173237fd80ec02f5) in a way that does not require any comment or check on attributes

simahawk:
@etobella good for you?

etobella (approved):

simahawk (approved):

PR #1051, opened 2024-10-07
[17.0][IMP] base_import_pdf_by_template: Add dd/MM/YY date_format compatibility

victoralmau:
Changes done:
- [x] Add dd/MM/YY `date_format` compatibility
- [x] Add dd/MM/yy `date_format` compatibility
- [x] Add B d, YY `date_format` compatibility 
- [x] Add file_name to view to display the file correctly
- [x]  Improve the display of line data in the form view
- [x] Add message_log in the correct format (html)

Please @CarlosRoca13 and @pilarvargas-tecnativa can you review it?

@Tecnativa TT50003

### DIFF
--- a/base_import_pdf_by_template/views/base_import_pdf_template_line_views.xml
+++ b/base_import_pdf_by_template/views/base_import_pdf_template_line_views.xml
@@ -19,68 +19,69 @@
             <form>
                 <sheet>
                     <group name="main_group">
-                        <group name="left">
-                            <field name="model" invisible="1" />
-                            <field name="related_model" widget="radio" />
-                            <field
-                                name="field_id"
-                                invisible="not related_model"
-                                options="{'no_create': True}"
-                            />
-                            <field name="field_relation" invisible="1" />
-                            <field name="column" invisible="related_model != 'lines'" />
-                            <field
-                                name="pattern"
-                                invisible="value_type == 'fixed'"
-                                required="value_type != 'fixed'"
-                            />
-                            <field
-                                name="value_type"
-                                widget="radio"
-                                invisible="not field_id"
-                            />
-                            <field
-                                name="search_field_id"
-                                invisible="not field_id or value_type == 'fixed'"
-                            />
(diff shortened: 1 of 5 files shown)


pilarvargas-tecnativa (approved):
LGTM, thanks!

base_import_pdf_by_template/views/base_import_pdf_template_line_views.xml
```diff
@@ -19,68 +19,69 @@
             <form>
                 <sheet>
                     <group name="main_group">
-                        <group name="left">
```
pedrobaeza, line 22: Why removing groups? At first sight, I prefer to have it into 2 columns.

base_import_pdf_by_template/wizards/wizard_base_import_pdf_upload.py
```diff
@@ -143,28 +145,30 @@ def _process_set_value_form(self, _form, field_name, value):
...
-                    %(old_value)s</p>"""
+                text = _(
+                    """%(item_name)s has been set with %(new_value)s instead of
+                    %(old_value)s"""
                 ) % {
                     "item_name": getattr(_form, "name"),  # noqa: B009
                     "old_value": old_value_data,
                     "new_value": new_value_data,
                 }
+                if not self.log_text:
```
pedrobaeza: Don't repeat this all the times. Extract it to a method, and don't repeat also the Markup creation.

pedrobaeza (commented):
I think you can squash all the commits into one. Just enumerate all the changes in the expanded commit message.

base_import_pdf_by_template/views/base_import_pdf_template_line_views.xml
victoralmau: I think an example may help to understand it

**Before**
antes

**After**
despues

pedrobaeza (approved):
/ocabot merge minor

PR #1056, opened 2024-10-24
[17.0][ADD] base_import_pdf_by_template_account: New module

victoralmau:
New module

Locked by:
- [x] `pdf_helper`: https://github.com/OCA/edi/pull/1057

Please @pilarvargas-tecnativa can you review it?

@Tecnativa TT50003


victoralmau:
Ping @pedrobaeza What do you think about this approach?

base_import_pdf_by_template_account/models/account_move.py
```diff
@@ -0,0 +1,14 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from odoo import models
+
+
+class AccountMove(models.Model):
+    _inherit = "account.move"
+
+    def _get_edi_decoder(self, file_data, new=False):
+        if file_data["type"] == "pdf":
+            return self.env[
```
pedrobaeza: This mustn't return, but try to import, and if not, call super.

test_base_import_pdf_by_template/demo/base_import_pdf_template.xml
```diff
@@ -113,140 +113,4 @@
         <field name="value_type">variable</field>
         <field name="log_distinct_value" eval="True" />
     </record>
-    <!-- Account invoice Template Data !-->
```
pedrobaeza, line 123: Why removing all of this?

base_import_pdf_by_template/models/baseimportpdfbytemplate_mixin.py
```diff
@@ -0,0 +1,19 @@
+# Copyright 2024 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from odoo import models
+
+
+class BaseimportpdfbytemplatetMixin(models.AbstractModel):
+    _name = "baseimportpdfbytemplate.mixin"
+    _description = "Functions for baseimportpdfbytemplate mixin documents"
+
+    def _import_record_base_import_pdf_by_template(self, record, attachment):
```
pedrobaeza: I don't think this little code is worth for a whole mixin

victoralmau: The idea of this was that it could be used in another context, for example, when processing fetchmail and creating a purchase/sale order.

test_base_import_pdf_by_template/demo/base_import_pdf_template.xml
victoralmau: Previously the account template and the test were in `test_base_import_pdf_by_template`. Now, with the new module (which already depends on `account`) I have moved that to this new module and used that template to create a test of the process of creating an invoice from the PDF that is processed when receiving an email.

base_import_pdf_by_template/models/baseimportpdfbytemplate_mixin.py
pedrobaeza: That's not a real case right now (there are no such mechanisms) and it's very few code to be reused, so for now, please don't do it.

test_base_import_pdf_by_template/demo/base_import_pdf_template.xml
pedrobaeza: But then, do we need the `test_*` module?

victoralmau: For now, it is still necessary for a purchase order template (for example).

pedrobaeza: OK

base_import_pdf_by_template_account/models/account_move.py
victoralmau: This was made similar to `account_edi_ubl_cii` (https://github.com/odoo/odoo/blob/17.0/addons/account_edi_ubl_cii/models/account_move.py#L52) because `account` processes the response in a “_special_” way https://github.com/odoo/odoo/blob/17.0/addons/account/models/account_move.py#L3203 but **ok**, I change it to what you comment.

victoralmau:
It's ready to review.

pilarvargas-tecnativa:
Remove WIP tag from PR title

base_import_pdf_by_template_account/i18n/es.po
```diff
@@ -0,0 +1,62 @@
...
+msgstr ""
+
+#. module: base_import_pdf_by_template_account
+#: model:product.template,name:base_import_pdf_by_template_account.product_leds_product_template
+msgid "Test Leds"
+msgstr ""
+
+#. module: base_import_pdf_by_template_account
+#: model:product.template,name:base_import_pdf_by_template_account.product_laminas_product_template
+msgid "Test Lmaninas"
```
pilarvargas-tecnativa: Maybe you meant to say ‘Láminas’?

base_import_pdf_by_template_account/demo/base_import_pdf_template.xml
```diff
@@ -152,4 +159,18 @@
...
+    >
+        <field name="template_id" ref="invoice_tecnativa" />
+        <field name="related_model">lines</field>
+        <field
+            name="field_id"
+            ref="account.field_account_move_line__analytic_distribution"
+        />
+        <field name="column">3</field>
+        <field name="value_type">fixed</field>
+        <!--<field name="fixed_value_text">{"cls.env.ref('base_import_pdf_by_template_account.analytic_account_tecnativa').id": 100.0}</field>!-->
```
pilarvargas-tecnativa, line 174: It is not necessary?

victoralmau: Yes, it is necessary but it is commented because it is not possible to do that directly. If you know the way to do it, it will be welcome! (this is currently set correctly in tests).

pilarvargas-tecnativa (approved):

pedrobaeza:
Better to not use demo data on tests, as already stated in other PRs.

victoralmau:
So in all tests we create everything related to a template again? I do not agree that this is useful in the OCA modules (I am not sure it is an “_official_” guideline either.).

pedrobaeza:
You have to test the specific gestures of this module, not an entire template, and yes, what's needed for that tests should be in the test itself.

victoralmau:
Sorry, but how is the `_import_base_import_pdf_by_template()` method supposed to be tested without a full template (that template already existed in `test_base_import_pdf_by_template` so that user installing the module can test it). The template is also useful for testing analytic distributions.

As a personal reflection, i think that all this "_bureaucracy_" is not good and will only demotivate some contributors.

pedrobaeza:
This is not bureaucracy. It's doing tests the best way. I already stated the reasons to not use demo data in other PR. Do you remember?

I'm not saying here to not use templates. What I'm saying is that you don't need a whole template, just the minimal data to test the features. And all of that definitions inside the test class, not as demo data.

victoralmau:
> This is not bureaucracy. It's doing tests the best way. I already stated the reasons to not use demo data in other PR. Do you remember?

Yes, and I explained then that I disagreed on the arguments. I explained earlier that you moved the demo data from `test_base_import_pdf_by_template` (and the tests) to `base_import_pdf_by_template_account` but never mind, I'll make the change so you don't block the PR. I'll get on it.

pedrobaeza:
> Yes, and I explained then that I disagreed on the arguments

You said that you don't agree, and then I'll explain the reasons, but you didn't counter-argument my explanations to say why it's better to use demo data. Odoo itself has moved to be demo-data independent because of these reasons.

victoralmau:
It was not worth counter-argument, just like now, it is simply better to make the changes you suggest and that's it.

pedrobaeza:
OK, we have debated about this internally. Just for others reading this, there's a guideline about this:

https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#616avoid-relying-on-demo-data

Let's change it for avoiding problems in the future.

victoralmau:
Extra commit added (and many lines) to define data in tests (without using demo data). Do you want that change also made to `base_import_pdf_by_template` and `test_base_import_pdf_by_template`?

pedrobaeza:
Do you really need the supplierinfo for all the products?
Do you really need so many products?

When I talked about doing the minimal template in tests, I was referring to that, to have the minimal test unit, not a PDF template with lots of lines that are the same (but with another product).

I don't want to bother you more in this one, but it's for you to take that into consideration. Always design the tests with the minimum use case, not a one with a lot of things repeated, but that are not adding real value to the test, just boilerplate creating more data, but testing the same feature.

victoralmau:
> Do you really need the supplierinfo for all the products? Do you really need so many products?
> 
> When I talked about doing the minimal template in tests, I was referring to that, to have the minimal test unit, not a PDF template with lots of lines that are the same (but with another product).
> 
This is importing an existing .pdf file created specifically for that template (it was created some time ago), you can create another one specifically for the tests, but it would not make sense (or at least not to me), precisely for this reason I did not see sense to make this change.

pedrobaeza:
Well, even if using this not so ideal template, I think you can create just one product with full data, and another "fake one" that you put as fallback in case the product is not found for the rest of the lines. Or is there any exception in that case?

victoralmau:
The test was intended to validate a multi-line use case with different quantities, prices, etc, but if you want we can change it... (I don't understand that these comments appear on unrelated things, i.e. the test already existed and was validated).

pedrobaeza (approved):
/ocabot merge patch

PR #1063, opened 2024-11-07
[15.0][MIG] account_invoice_import: Migration to 15.0

V3st1g3:


bosd:
Can you force push?

bosd:
Looks like a dependency is missing `base_business_document_import`. Can you add it to the test-requirements.txt

bosd:
oopsie, the module `base_business_document_import` is not available on v15 yet. this was the most recent/complete migration attempt: https://github.com/OCA/edi/pull/737 
But it's better to start a new one, like I indicated there in the commits.
Will do it, whn I find some time.

PR #1065, opened 2024-11-14
Fix ubl_template picking, always return DeliveredQuantity

cyrilmanuel:
The block for `t-if="line.Mngs_Mng"` return false in case of the value is 0.0 that mean you will not see the block as result for values set to 0.0. 

adding test with entry set to 0.0 as qty


jbaudoux (approved):

base_wamas_ubl/lib/wamas/ubl_template/picking.xml
```diff
@@ -138,15 +138,8 @@
 			</cac:GrossWeightMeasure>
 		</cac:Shipment>
 		<cbc:ID><t t-raw="line.IvAusp_HostPosNr" /></cbc:ID>
+		<t t-raw="line.Mngs_Mng or line.IvAusp_LiefMngsWamas_Mng" />
```
jbaudoux: oh, this line needs to move inside the DeliveredQuantity block instead of the other

jbaudoux (changes requested):

jbaudoux (approved):

santostelmo (approved):

base_wamas_ubl/lib/wamas/ubl_template/picking.xml
```diff
@@ -139,14 +139,14 @@
 		</cac:Shipment>
 		<cbc:ID><t t-raw="line.IvAusp_HostPosNr" /></cbc:ID>
 		<cbc:DeliveredQuantity
-            t-if="line.Mngs_Mng"
+            t-if="line.Mngs_Mng or line.Mngs_Mng == 0"
```
Ricardoalso: Not a big issue but this if condition and the one below are kind of useless as they will always be True

Ricardoalso (approved):

base_wamas_ubl/lib/wamas/ubl_template/picking.xml
santostelmo: I agree, this t-if could be dropped

jbaudoux: @cyrilmanuel Why you went back from your initial change where you kept only one `DeliveredQuantity` block? It looked correct. It doesn't make sense to have twice the block and it should be always there, so `t-if` should be dropped

cyrilmanuel: Hi, with only one block its not working the block contain only value set to none. Same for the if.

simahawk: I agree that the comparison to zero seem useless

jbaudoux: Repeating the block with a condition is useless as the block is mandatory and can be there only once. i'm convinced it should be just a matter of expressing properly the value extraction inside the unique block

cyrilmanuel: @jbaudoux  @simahawk  i have made the modification could you explain me what i've made wrong ?

base_wamas_ubl/lib/wamas/ubl_template/picking.xml
```diff
@@ -139,18 +139,10 @@
...
         >
-		<t t-raw="line.Mngs_Mng" />
-		</cbc:DeliveredQuantity>
-		<cbc:DeliveredQuantity
-            t-if="line.IvAusp_LiefMngsWamas_Mng"
-            t-att-unitCode="line.IvArt_Bestand_Einheit"
-            unitCodeListAgencyName="WAMAS"
-        >
-		<t t-raw="line.IvAusp_LiefMngsWamas_Mng" />
+		<t t-raw="line.Mngs_Mng or line.IvAusp_LiefMngsWamas_Mng" />
```
jbaudoux: If the expression is wrong, it will silently return "None". You need to check if the attribute is part of the object before accessing it

```suggestion
        <t t-raw="hasattr(line,'Mngs_Mng') and line.Mngs_Mng or (hasattr(line, 'IvAusp_LiefMngsWamas_Mng') and line.IvAusp_LiefMngsWamas_Mng) or 0.0" />
```

jbaudoux (changes requested):
You will also need to remove the blank line in `wamas/tests/samples/WAMAS2UBL-SAMPLE_AUSKQ_WATEKQ_WATEPQ-DESPATCH_ADVICE.xml`

```
  		</cbc:DeliveredQuantity>
- 		
  		<cac:OrderLineReference>
```

base_wamas_ubl/lib/wamas/ubl_template/picking.xml
jbaudoux: sure, see below

jbaudoux (approved):
Yes, like that :)

PR #1067, opened 2024-11-14
[14.0] edi_sale_oca: use edi.configuration 

simahawk:
Depends on
- [x] #1035

* provides full support and examples on how to use edi.conf in real life
* after we merge this we can get rid of exc type auto in #1024

### DIFF
--- a/edi_sale_oca/demo/edi_configuration.xml
+++ b/edi_sale_oca/demo/edi_configuration.xml
@@ -1,0 +1,38 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<odoo>
+    <record id="demo_edi_configuration_confirmed" model="edi.configuration">
+        <field name="name">Demo Sale OrderResponse - order confirmed</field>
+        <field
+            name="description"
+        >Show case how you can send out an order response automatically</field>
+        <field name="type_id" ref="demo_edi_exc_type_order_out" />
+        <field name="backend_id" ref="demo_edi_backend" />
+        <field name="model_id" ref="sale.model_sale_order" />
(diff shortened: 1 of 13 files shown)


simahawk:
@JordiMForgeFlow would you mind to have a look at this?
I'll work on sale ubl and purchase ubl as well in the same fashion (using new edi.configuration model) and you'll likely benefit from it in v17 ;)

edi_sale_oca/demo/edi_configuration.xml
```diff
@@ -0,0 +1,38 @@
...
+        <field name="trigger_id" ref="edi_conf_trigger_sale_order_state_change" />
+        <field name="snippet_do">
+# STATES
+# ('draft', 'Quotation'),
+# ('sent', 'Quotation Sent'),
+# ('sale', 'Sales Order'),
+# ('done', 'Locked'),
+# ('cancel', 'Cancelled'),
+if record.state == 'sale':
+  record._edi_send_via_edi(conf.type_id)
```
SilvioC2C, line 20: Could be useful to add help text explaining which variables are available to use in the snippet (like Odoo does by default for server actions or crons)

SilvioC2C (approved):
Overall LGTM, one small imrpovement suggested

edi_sale_oca/demo/edi_configuration.xml
simahawk: variables come from edi_oca. I'll check there.

simahawk:
Build on 14.0 is broken https://github.com/OCA/edi/issues/1084

PR #1068, opened 2024-11-16
[16.0] base_wamas_ubl: run lib tests

jbaudoux:
* Add missing freezetime
* Fix html escaping
* Move lib tests to lib (some tests were inside the lib but not run by the CI, some other pure lib tests were inside the module)
* Run all lib tests as module unittests


jbaudoux:
ping @cyrilmanuel @santostelmo @simahawk 
Some lib tests were failing (due to missing freezetime) but it was not visible as they were not run by the CI

jbaudoux:
@simahawk Can you review and merge this if good for you?

jbaudoux:
cc @nilshamerlinck

base_wamas_ubl/lib/wamas/dict2wamas.py
```diff
@@ -60,6 +62,7 @@ def dict2wamas(dict_input, msg_type):
     return wamas
 
 
+@freeze_time("2024-02-11 22:14:22")
```
simahawk, line 65: why freezing this?

jbaudoux: That's the datetime in the xml, so that the test succeeds

jbaudoux:
rebased

PR #1076, opened 2024-11-29
[14.0][IMP] product_import: reorganize for inheritance to import by batch

florentx:
Some catalogs contain 10'000 references and import does not complete due to other tasks running at the same time, and locking table.

Splitting product import into individual jobs should mitigate this issue.

**[ EDIT ]**
This PR reorganizes code in order to easily be able to split import by batch with method `split()` of `queue_job` add-on, or any similar machinery.

### DIFF
--- a/product_import/tests/test_product_import.py
+++ b/product_import/tests/test_product_import.py
@@ -50,6 +50,7 @@ PARSED_CATALOG = {
         },
     ],
     "ref": "1387",
+    "company": {"name": "Customer ABC"},
     "seller": {
         "contact": False,
         "email": False,
@@ -88,9 +89,15 @@ class TestProductImport(TestCommon):
 
     def test_product_import(self):
         # product.product
-        products = self.wiz_model._create_products(
-            self.parsed_catalog, seller=self.supplier
+        product_obj = self.env["product.product"].with_context(active_test=False)
+        existing = product_obj.search([], order="id")
+
+        wiz = self.wiz_model.create(
+            {"product_file": b"====", "product_filename": "test_import.xml"}
         )
+        with self._mock("parse_product_catalogue", return_value=self.parsed_catalog):
+            wiz.import_button()
+        products = product_obj.search([], order="id") - existing
         self.assertEqual(len(products), 3)
         for product, parsed in zip(products, PARSED_CATALOG["products"]):
 
@@ -140,7 +147,7 @@ class TestProductImport(TestCommon):
             self.assertEqual(product.seller_ids, product_tmpl.seller_ids)
             self.assertEqual(
                 product.seller_ids.mapped("delay")[0], parsed.get("sale_delay", 0)
-            ),
+            )
 
     def test_import_button(self):
         form = self.wiz_form
--- a/product_import/wizard/product_import.py
+++ b/product_import/wizard/product_import.py
@@ -150,7 +150,7 @@ class ProductImport(models.TransientModel):
         return result
 
     @api.model
-    def _prepare_product(self, parsed_product, chatter_msg, seller=None):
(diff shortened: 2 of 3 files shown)


product_import/wizard/product_import.py
```diff
@@ -211,23 +212,29 @@ def create_product(self, parsed_product, chatter_msg, seller=None):
...
-            )
-            if record:
-                products |= record
+    def _create_update_product(self, parsed_product, seller_id):
+        """Create / Update a product.
+
+        This method is called from a queue job.
+        """
+        messgs = []
+        product = self.create_update_product(parsed_product, messgs, seller_id)
```
simahawk: Sounds weird to have a public method called by the private one. It should be the other way around.
And why passing empty `messgs`?

product_import/tests/test_product_import.py
```diff
@@ -86,8 +86,11 @@ def test_get_company_id(self):
 
     def test_product_import(self):
         # product.product
-        products = self.wiz_model._create_products(
-            self.parsed_catalog, seller=self.supplier
+        self.wiz_model._import_products(self.parsed_catalog, seller=self.supplier)
+        products = (
```
simahawk: existing = ...search()
new = ...search() - existing

product_import/wizard/product_import.py
```diff
@@ -211,23 +212,29 @@ def create_product(self, parsed_product, chatter_msg, seller=None):
...
+        log_msg = f"Product created/updated {product.id}\n" + "\n".join(messgs)
+        return log_msg
+
+    @api.model
+    def _import_products(self, catalogue, seller, filename=None):
+        for product_vals in catalogue["products"]:
+            # One job per product
+            self.with_delay()._create_update_product(product_vals, seller.id)
         self._bdimport.post_create_or_update(catalogue, seller, doc_filename=filename)
```
simahawk: Is it correct to call the post update here?
Also, now you have 3 methods to the import... why do you need them all?

simahawk (changes requested):

product_import/wizard/product_import.py
florentx: Refactored.

twalter-c2c (approved):
LGTM

product_import/__manifest__.py
```diff
@@ -13,9 +13,12 @@
         "stock",
         # OCA/edi
         "base_business_document_import",
+        # OCA/queue
+        "queue_job",
```
simahawk: I've thought again about this... We should remove this dependency IMO.
In the simplest case we won't need it, in the more complex case we'll probably use the framework. Then, we can leverage the framework to the split. 
Additionally, there's now a new feature in queue_job which is called `split` and allows to.... split items in batches.
See https://github.com/OCA/queue/pull/658

product_import/wizard/product_import.py
```diff
@@ -213,33 +248,33 @@ def create_product(self, parsed_product, chatter_msg, seller=None):
...
-        self.with_context(product_company_id=company_id)._create_products(
-            catalogue, seller, filename=self.product_filename
+        wiz = self.with_context(product_company_id=company_id)
+        # Create products asynchronously
+        iterators = [iter(catalogue["products"])] * chunk_size
+        for products in zip_longest(*iterators):
+            if products[-1] is None:
+                products = [product for product in products if product]
+            # One job for x products (chunk of 40)
+            wiz.with_delay()._create_update_products(products, seller.id)
```
simahawk: here I would simply call `wiz.create_update_products` that will call `_create_update_products` which can be overridden or you can rely on https://github.com/OCA/queue/blob/16.0/queue_job/models/base.py#L145

twalter-c2c (approved):
LGTM

PR #1098, opened 2024-12-20
[14.0] edi_purchase: use edi.configuration + add edi_purchase_ubl_output_oca

simahawk:
Replaces #669

### DIFF
--- a/edi_purchase_oca/__init__.py
+++ b/edi_purchase_oca/__init__.py
@@ -1,1 +1,2 @@
 from . import models
+from . import components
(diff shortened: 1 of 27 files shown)


ivantodorovich (approved):

JordiMForgeFlow:
@simahawk in this case for example, shouldn't we create in the module data the equivalent triggers for PO confirm and PO cancel? At least to keep the original behavior

simahawk:
> @simahawk in this case for example, shouldn't we create in the module data the equivalent triggers for PO confirm and PO cancel? At least to keep the original behavior

The original behavior is kept because I'm not removing the old events' triggers. It would be too risky.
However, I added a generic config to hook to the state change.

Am I missing anything?

JordiMForgeFlow:
@simahawk oh yes, I see 👍🏼

JordiMForgeFlow (approved):

PR #1112, opened 2025-01-27
[17.0][FIX] base_import_pdf_by_template_account: Avoid error when adding a pdf attachment to an invoice if no template applies

victoralmau:
Avoid error when adding a pdf attachment to an invoice if no template applies

@Tecnativa TT54717

### DIFF
--- a/base_import_pdf_by_template/wizards/wizard_base_import_pdf_upload.py
+++ b/base_import_pdf_by_template/wizards/wizard_base_import_pdf_upload.py
@@ -60,7 +60,9 @@ class WizardBaseImportPdfUpload(models.TransientModel):
         self.line_ids = lines
         # Error si corresponde
         lines_without_template = self.line_ids.filtered(lambda x: not x.template_id)
-        if lines_without_template:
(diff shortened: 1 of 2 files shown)


pedrobaeza (changes requested):
I don't think this should go this way. What should be is that there's no error if there's no template to apply to the PDF.

victoralmau:
Change to another approach, what do you think?

base_import_pdf_by_template/wizards/wizard_base_import_pdf_upload.py
```diff
@@ -70,7 +72,7 @@ def action_process(self):
             )
         # Process + return records
         records = self.env[self.model]
-        for line in self.line_ids:
+        for line in self.line_ids.filtered(lambda x: x.template_id):
```
pedrobaeza:
```suggestion
        for line in self.line_ids.filtered("template_id"):
```

pedrobaeza (approved):

carlos-lopez-tecnativa (approved):
Technical review: LGTM.

PR #1123, opened 2025-02-05
[14.0] edi_exchange_template: declare tmpl to use explicitly 

simahawk:
Relying on the code by convention was fragile for many reasons.
Most important:

1. users have to remember about this convention
2. is not clear which template is going to be used by a specific type

With this change we make it more clear and deprecate the old behavior.

A long awaited improvement....

### DIFF
--- a/edi_exchange_template_oca/models/edi_backend.py
+++ b/edi_exchange_template_oca/models/edi_backend.py
@@ -34,9 +34,12 @@ class EDIBackend(models.Model):
         # TODO: maybe we can add a m2o to output templates
         # but then we would need another for input templates if they are introduced.
         tmpl = None
+        code = code or exchange_record.type_id.code
         if code:
             domain = [("code", "=", code)]
-            return search(domain, limit=1)
+            tmpl = search(domain, limit=1)
+            if tmpl:
+                return tmpl
         for domain in self._get_output_template_domains(exchange_record):
             tmpl = search(domain, limit=1)
             if tmpl:
--- a/edi_exchange_template_oca/models/edi_backend.py
+++ b/edi_exchange_template_oca/models/edi_backend.py
@@ -1,8 +1,13 @@
 # Copyright 2020 ACSONE SA
+# Copyright 2025 Camptocamp SA
 # @author Simone Orsi <simahawk@gmail.com>
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
 
-from odoo import models
+import logging
+
+from odoo import fields, models
+
+_logger = logging.getLogger(__name__)
 
 
 class EDIBackend(models.Model):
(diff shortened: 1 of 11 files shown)


edi_exchange_template_oca/models/edi_backend.py
```diff
@@ -26,27 +31,48 @@ def output_template_model(self):
...
+            "|",
+            ("allowed_type_ids", "=", exchange_record.type_id.id),
+            ("allowed_type_ids", "=", False),
+        ]
+        candidates = self.output_template_model.search(base_domain)
+        for rec in candidates:
+            if rec.type_id == exchange_record.type_id:
+                return rec
+        # Take the 1st one having allowed_type_ids set
+        return fields.first(candidates.sorted(lambda x: 0 if x.allowed_type_ids else 1))
```
mmequignon, line 78:
```suggestion
        return fields.first(candidates.sorted(lambda x: x.allowed_type_ids))
```

edi_exchange_template_oca/models/edi_backend.py
```diff
@@ -26,27 +31,48 @@ def output_template_model(self):
...
+        # Match by backend and allowed types
+        base_domain = [
+            ("backend_type_id", "=", self.backend_type_id.id),
+            "|",
+            ("allowed_type_ids", "=", exchange_record.type_id.id),
+            ("allowed_type_ids", "=", False),
+        ]
+        candidates = self.output_template_model.search(base_domain)
+        for rec in candidates:
+            if rec.type_id == exchange_record.type_id:
```
mmequignon, line 75: Doesn't look like smth we want to do in a getter method.
Can this be extracted?

edi_exchange_template_oca/tests/test_backend_and_type.py
```diff
@@ -0,0 +1,157 @@
...
+        cls.record1 = cls.backend.create_record("test_type_out1", vals)
+        vals = {
+            # doesn't matter what model we use
+            "model": cls.env.user._name,
+            "res_id": cls.env.user.id,
+            "type_id": cls.type_out2.id,
+        }
+        cls.record2 = cls.backend.create_record("test_type_out2", vals)
+
+    # TODO: getting a template via code or relying on a fallback is deprecated
```
mmequignon, line 79: Any test regarding testing that a msg is logged when using code/fallback?

mmequignon (approved):
Code lg, don't have the full picture though.
Tiny remarks.

edi_exchange_template_oca/models/edi_backend.py
```diff
@@ -26,27 +31,48 @@ def output_template_model(self):
...
-        backend_type_leaf = [("backend_type_id", "=", self.backend_type_id.id)]
-        exchange_type_leaf = [("type_id", "=", exchange_record.type_id.id)]
-        full_match_domain = backend_type_leaf + exchange_type_leaf
-        partial_match_domain = backend_type_leaf
-        return full_match_domain, partial_match_domain
+        # Match by backend and allowed types
+        base_domain = [
+            ("backend_type_id", "=", self.backend_type_id.id),
+            "|",
+            ("allowed_type_ids", "=", exchange_record.type_id.id),
```
yankinmax:
```suggestion
            ("allowed_type_ids", "in", exchange_record.type_id.ids),
```

yankinmax (approved):
Thanks

yankinmax:
@simahawk @mmequignon 
can you pls drop a review also on the forwarded PR to v18:
- [ ] https://github.com/OCA/edi-framework/pull/138

edi_exchange_template_oca/models/edi_backend.py
yankinmax: The proposed change isn't equal and results in an error.

simahawk: exactly :)
Items w/o allowed types would be sorted 1st.

edi_exchange_template_oca/models/edi_backend.py
simahawk: I want to trash this code. I won't spend time on this :wink:

edi_exchange_template_oca/tests/test_backend_and_type.py
simahawk: I'm testing the fallback itself... and yeah ok, but I want to trash this code... so I won't spend time now.

edi_exchange_template_oca/models/edi_backend.py
simahawk: done

PR #1129, opened 2025-02-18
[17.0][IMP] base_import_pdf_by_template: Add *d-*b-*Y date_format

victoralmau:
Add *d-*b-*Y date_format

Please @pedrobaeza can you review it?

@Tecnativa TT51980

### DIFF
--- a/base_import_pdf_by_template/README.rst
+++ b/base_import_pdf_by_template/README.rst
@@ -52,58 +52,59 @@ is to have the document defined with a specific structure.
 
    Fields to consider completing on template:
 
-      - Main Model: model on which the record will be generated.
-        Example: purchase.order
-      - Child field: One2many field that will create records from
-        selected template. Example: Order Lines (purchase.order)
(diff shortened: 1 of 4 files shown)


pedrobaeza (commented):
Why not having a custom format where you introduce in a text field the markers?

victoralmau:
Because a “basic” user who defines the templates does not know the date formats to be defined, for this reason I have added to ROADMAP an improvement in this regard (a model that has the formats defined in text and avoids having to create new ones here).

pedrobaeza:
But that's only an advanced option. You still have the other options, but in case of not being a "usual" format, you can do it without changing every time the code. At the end, if you have a lot of elements in the list, it's going to be the same confusing.

victoralmau:
Can we make that improvement in the future? In this case the format needed is clear.

pedrobaeza (approved):
OK

/ocabot merge patch

PR #1133, opened 2025-03-14
[15.0][IMP] edi_voxel_stock_picking_oca: Select the desired XML template in the partner company.

juancarlosonate-tecnativa:
We added the voxel_report_id field in res.partner to allow companies to select an XML report template. Additionally, we set its visibility to be restricted to companies only, ensuring it remains hidden for individual contacts.

@Tecnativa TT55445
@sergio-teruel @pedrobaeza

### DIFF
--- a/edi_voxel_stock_picking_oca/models/res_partner.py
+++ b/edi_voxel_stock_picking_oca/models/res_partner.py
@@ -1,0 +1,14 @@
+from odoo import api, fields, models
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    voxel_picking_report_id = fields.Many2one(
+        comodel_name="ir.actions.report",
+        domain=[("report_type", "=", "qweb-xml"), ("model", "=", "stock.picking")],
+    )
+
+    @api.model
+    def _commercial_fields(self):
+        return super()._commercial_fields() + ["voxel_picking_report_id"]
--- a/edi_voxel_stock_picking_oca/models/stock_picking.py
+++ b/edi_voxel_stock_picking_oca/models/stock_picking.py
@@ -75,9 +75,12 @@ class Picking(models.Model):
 
         pickings = self.filtered(able_to_voxel)
         if pickings:
-            pickings.enqueue_voxel_report(
-                "edi_voxel_stock_picking_oca.report_voxel_picking"
-            )
+            for picking in pickings:
+                report = picking.partner_id.voxel_picking_report_id or self.env.ref(
+                    "edi_voxel_stock_picking_oca.report_voxel_picking"
+                )
+                if report:
+                    picking.enqueue_voxel_report(report)
 
     def get_document_type(self):
         """Document type name to be used in the name of the Voxel report"""
--- a/edi_voxel_stock_picking_oca/README.rst
+++ b/edi_voxel_stock_picking_oca/README.rst
@@ -114,6 +114,7 @@ Contributors
   * Ernesto Tejeda
   * Pedro M. Baeza
   * Sergio Teruel
+  * Juan Carlos Oñate
 
 Maintainers
 ~~~~~~~~~~~
--- a/edi_voxel_stock_picking_oca/__manifest__.py
+++ b/edi_voxel_stock_picking_oca/__manifest__.py
@@ -15,6 +15,7 @@
         "views/stock_picking_views.xml",
         "views/res_company_view.xml",
         "views/res_config_settings_views.xml",
+        "views/res_partner_view.xml",
     ],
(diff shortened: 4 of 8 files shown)


edi_voxel_stock_picking_oca/models/res_partner.py
```diff
@@ -0,0 +1,22 @@
...
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    voxel_report_id = fields.Many2one(
+        comodel_name="ir.actions.report",
+        domain=[("report_type", "=", "qweb-xml"), ("model", "=", "stock.picking")],
+    )
+
+    def write(self, vals):
```
pedrobaeza: Instead of this, overwrite the method `commercial_fields` and add the field name. Example:

https://github.com/odoo/odoo/blob/b9abe46c1492b09e369434e76ec8196c6b02dd19/addons/account/models/partner.py#L658-L661

edi_voxel_stock_picking_oca/models/stock_picking.py
```diff
@@ -75,9 +75,14 @@ def able_to_voxel(record):
...
-            pickings.enqueue_voxel_report(
-                "edi_voxel_stock_picking_oca.report_voxel_picking"
-            )
+            for picking in pickings:
+                report_id = picking.partner_id.voxel_report_id
+                report = (
+                    report_id.get_external_id().get(report_id.id)
+                    if report_id
+                    else "edi_voxel_stock_picking_oca.report_voxel_picking"
+                )
```
pedrobaeza: - Name properly the variables. It should be `report`, as it contains a recordset, not the integer ID.
- Doing an `or`, you also cover the case where there's no XML-ID associated with the selected report.

```suggestion
                report = picking.partner_id.voxel_report_id
                report = ( 
                    report.get_external_id().get(report_id.id)
                    or
                    "edi_voxel_stock_picking_oca.report_voxel_picking"
                )
```

Anyway, this has a problem: it's not accepting reports manually created without XML-ID. You can annotate that problem to be taken into account.

pedrobaeza (changes requested):

juancarlosonate-tecnativa:
ping @sergio-teruel

edi_voxel_stock_picking_oca/models/res_partner.py
```diff
@@ -0,0 +1,14 @@
+from odoo import api, fields, models
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    voxel_report_id = fields.Many2one(
```
sergio-teruel: This name is a lot generic...
Whats happend if I want a specific report for voxel xml report invoices??
So I think that the name could be voxel_picking_report_id...

edi_voxel_stock_picking_oca/models/stock_picking.py
```diff
@@ -75,9 +75,14 @@ def able_to_voxel(record):
 
         pickings = self.filtered(able_to_voxel)
         if pickings:
-            pickings.enqueue_voxel_report(
-                "edi_voxel_stock_picking_oca.report_voxel_picking"
```
sergio-teruel, line 79: Why do not  you fix the method 'enqueue_voxel_report' in edi_voxel_oca to work properly with a report name and with a report object??

https://github.com/OCA/edi/blob/ec9df99605bff89235f3b0cc344fb7eab0a0299e/edi_voxel_oca/models/voxel_mixin.py#L53

https://github.com/OCA/edi/blob/ec9df99605bff89235f3b0cc344fb7eab0a0299e/edi_voxel_oca/models/voxel_mixin.py#L73-L75

sergio-teruel (changes requested):

edi_voxel_stock_picking_oca/models/stock_picking.py
juancarlosonate-tecnativa: Because then we will need to review all the dependencies on it, but i don't think it's too much, im doing

juancarlosonate-tecnativa: https://github.com/OCA/edi/pull/1135

pedrobaeza:
Please adjust this according the other PR.

juancarlosonate-tecnativa:
> Please adjust this according the other PR.

there's nothing to change, its already sending the object

pedrobaeza (approved):
If you don't ping back when doing the changes, we are not aware of such changes.

juancarlosonate-tecnativa:
ping @sergio-teruel

sergio-teruel (approved):

PR #1134, opened 2025-03-17
[17.0][FIX] base_import_pdf_by_template: Don't fail on fetchmail

pedrobaeza:
If you have a bad formatted PDF that comes from the mail, the cron process may fail trying to parse it, and block the whole fetching indefinitely.

Let's avoid the crash masking it behind the existing context key.

@Tecnativa TT55481

### DIFF
--- a/base_import_pdf_by_template/wizards/wizard_base_import_pdf_upload.py
+++ b/base_import_pdf_by_template/wizards/wizard_base_import_pdf_upload.py
@@ -73,7 +73,11 @@ class WizardBaseImportPdfUpload(models.TransientModel):
         # Process + return records
         records = self.env[self.model]
         for line in self.line_ids.filtered("template_id"):
-            records += line.action_process()
+            try:
+                records += line.action_process()
+            except Exception:
+                if not self.env.context.get("skip_template_not_found_error"):
+                    raise
         action = {
             "type": "ir.actions.act_window",
             "res_model": records._name,
@@ -119,13 +123,16 @@ class WizardBaseImportPdfUploadLine(models.TransientModel):
     def _compute_template_id(self):
         self.template_id = False
         for item in self.filtered("attachment_id"):
-            data = item._parse_pdf_grouped()
-            text = ""
-            for key in list(data.keys()):
-                text += "".join(data[key])
-            item.template_id = (
-                item.parent_id.allowed_template_ids._auto_detect_from_text(text)
-            )
+            try:
+                data = item._parse_pdf_grouped()
+                text = ""
+                for key in list(data.keys()):
+                    text += "".join(data[key])
+                item.template_id = (
+                    item.parent_id.allowed_template_ids._auto_detect_from_text(text)
+                )
+            except Exception:  # pylint: disable=W8138 - never fail
+                pass
 
     def action_process(self):
         """Parse the file again, this time with the corresponding extraction mode."""


victoralmau (approved):

base_import_pdf_by_template/wizards/wizard_base_import_pdf_upload.py
```diff
@@ -73,7 +73,11 @@ def action_process(self):
         # Process + return records
         records = self.env[self.model]
         for line in self.line_ids.filtered("template_id"):
-            records += line.action_process()
+            try:
+                records += line.action_process()
+            except Exception:
```
carlos-lopez-tecnativa, line 78: Code review, LGTM.

However, I think it's good practice to wrap the `try/except` block inside a `savepoint` to prevent cache issues.

```suggestion
            try:
                with self.env.cr.savepoint():
                    records += line.action_process()
            except Exception:
```

base_import_pdf_by_template/wizards/wizard_base_import_pdf_upload.py
```diff
@@ -119,13 +123,16 @@ class WizardBaseImportPdfUploadLine(models.TransientModel):
...
-            )
+            try:
+                data = item._parse_pdf_grouped()
+                text = ""
+                for key in list(data.keys()):
+                    text += "".join(data[key])
+                item.template_id = (
+                    item.parent_id.allowed_template_ids._auto_detect_from_text(text)
+                )
+            except Exception:  # pylint: disable=W8138 - never fail
```
carlos-lopez-tecnativa, line 134: Same applies here.

carlos-lopez-tecnativa (commented):
@pedrobaeza @victoralmau  What do you think?

base_import_pdf_by_template/wizards/wizard_base_import_pdf_upload.py
pedrobaeza: There's no records involved here yet, as it only creates the move, and if failed, no record is created.

base_import_pdf_by_template/wizards/wizard_base_import_pdf_upload.py
pedrobaeza: There's no record here, as this is only a computed field.

base_import_pdf_by_template/wizards/wizard_base_import_pdf_upload.py
carlos-lopez-tecnativa: > There's no records involved here yet, as it only creates the move, and if failed, no record is created.

Ok, I was just considering a scenario where a record has multiple lines:

- L1 → OK
- L2 → Error
- L3 → OK

Before the commit, if L2 failed, the transaction was fully reverted, and execution stopped. Now, execution continues, but L2 is created and not reverted. If this is the expected behavior, then it's fine.

carlos-lopez-tecnativa (approved):

base_import_pdf_by_template/wizards/wizard_base_import_pdf_upload.py
pedrobaeza: No, that's the question. The transaction is not reverted, as the Python crash is intercepted and no exception is raised upstream. The only possible fail is if there's a PG crash (like DB constraint or similar), but this is not the case.

PR #1135, opened 2025-03-21
[15.0][IMP] edi_voxel_oca: Allow enqueue_voxel_report to accept report objects

juancarlosonate-tecnativa:
@Tecnativa TT55445
@sergio-teruel @pedrobaeza

### DIFF
--- a/edi_voxel_oca/models/voxel_mixin.py
+++ b/edi_voxel_oca/models/voxel_mixin.py
@@ -50,7 +50,7 @@ class VoxelMixin(models.AbstractModel):
 
     # Export methods
     # --------------
-    def enqueue_voxel_report(self, report_name):
+    def enqueue_voxel_report(self, report):
         eta = self.company_id._get_voxel_report_eta()
         queue_obj = self.env["queue.job"].sudo()
         for record in self.sudo():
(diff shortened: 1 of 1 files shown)


pedrobaeza (changes requested):
I'm not sure about this change because 2 things:

- We pass the XML-ID for serializing the arguments in queue.job in an easy way. I don't know how the report object is being serialized there.
- Doing this change, all the existing pending jobs will fail due to the change of the argument.

juancarlosonate-tecnativa:
> I'm not sure about this change because 2 things:
> 
> * We pass the XML-ID for serializing the arguments in queue.job in an easy way. I don't know how the report object is being serialized there.
> * Doing this change, all the existing pending jobs will fail due to the change of the argument.

You’re right that the change would break existing pending jobs, but there shouldn’t be any issue with serialization since it was already being transformed into an object before. Using isinstance, it should now accept both formats.

edi_voxel_oca/models/voxel_mixin.py
```diff
@@ -65,14 +65,15 @@ def enqueue_voxel_report(self, report_name):
...
+                ._get_and_send_voxel_report(report)
             )
             job = queue_obj.search([("uuid", "=", new_delay.uuid)], limit=1)
             record.voxel_job_ids |= job
 
-    def _get_and_send_voxel_report(self, report_name):
+    def _get_and_send_voxel_report(self, report):
         self.ensure_one()
-        report = self.env.ref(report_name)
+        if isinstance(report, str):
```
pedrobaeza, line 76: Put a comment saying that this is for legacy support for old enqueued jobs.

pedrobaeza (approved):
/ocabot merge patch

PR #1138, opened 2025-03-27
[15.0][IMP] edi_voxel_stock_picking_oca: Remove HTML tags from picking note for Voxel compatibility

juancarlosonate-tecnativa:
The change is made because the picking.note field is now of type HTML, and even if the HTML is empty, `"Markup('<p><br></p>')"` is added. Voxel requires the note to be in str, and previously, if the field was empty, HTML tags were being sent. Now, the tags are stripped to ensure compatibility.

@Tecnativa TT55445
@sergio-teruel @pedrobaeza

### DIFF
--- a/edi_voxel_stock_picking_oca/reports/report_voxel_picking.py
+++ b/edi_voxel_stock_picking_oca/reports/report_voxel_picking.py
@@ -2,6 +2,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 from odoo import api, models
+from odoo.tools import html2plaintext
 
 
 class ReportVoxelPicking(models.AbstractModel):
@@ -77,7 +78,9 @@ class ReportVoxelPicking(models.AbstractModel):
         ]
 
     def _get_comments_data(self, picking):
-        return picking.note and [{"Msg": picking.note}] or []
+        """Ensure the note is in str and strips HTML tags"""
+        note = picking.note and html2plaintext(picking.note)
+        return [{"Msg": note}] if note else []
 
     def _get_references_data(self, picking):
         so = picking.sale_id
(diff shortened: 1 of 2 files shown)


edi_voxel_stock_picking_oca/reports/report_voxel_picking.py
```diff
@@ -77,7 +77,8 @@ def _get_customers_data(self, picking):
         ]
 
     def _get_comments_data(self, picking):
-        return picking.note and [{"Msg": picking.note}] or []
```
sergio-teruel, line 80: Please... explain the reason of this change!!!

juancarlosonate-tecnativa: done

pedrobaeza (commented):
Shouldn't you use `tools.html2text` or `html2plaintext`?

juancarlosonate-tecnativa:
> Shouldn't you use `tools.html2text` or `html2plaintext`?

But html2plaintext or html2text leaves the HTML tags in plain text, what we actually want is to remove all the tags, right?

pedrobaeza:
AFAIK, it's removing the tags and leaving only plain text.

juancarlosonate-tecnativa:
> AFAIK, it's removing the tags and leaving only plain text.

can't find the `html2text` method inside tools

image

pedrobaeza:
Sorry, the name is `html2plaintext` as you have shown, and it's the one doing a correct conversion from the line break to `\n`, on contrary than striptags, which is putting `estiloheading` all together, so you should switch to this other method.

juancarlosonate-tecnativa:
ping @sergio-teruel @pedrobaeza

pedrobaeza (approved):

sergio-teruel (approved):

PR #1140, opened 2025-04-02
[16.0][IMP] account_invoice_factux: stop pinning factur-x lib

alexis-via:


alexis-via:
@simahawk I don't know the root cause for https://github.com/OCA/edi/pull/1089 but now tests are green without pinning an old version of the factur-x lib, so I created this PR to remove the pinning in v16.
I guess the pinning must be kept in v14 because the tests in v14 use python 3.6.
Pinning could be removed in v15 because the tests in v15 use python 3.8.

alexis-via:
FYI, pinning is not present in v17 and v18, so no need to forward port.

rvalyi (approved):

PR #1144, opened 2025-04-14
[14.0][MIG] edi purchase edifact oca

mathiasfrancke:


mathiasfrancke:
For me locally the unit test `test_action_confirm()` is failing in last assertion. The problem there is that the generated file is using an ASCII coded umlaut (\xf6=ö) but the expected result is using the umlaut.

I'm not sure whether the ASCII coded umlaut is a valid case or not.

edi_purchase_edifact_oca/tests/test_edifact_purchase.py
```diff
@@ -0,0 +1,276 @@
...
+        self.product_3 = self.env.ref("product.product_product_5")
+        self.product_3.default_code = "FURN_667777"
+        self.product_3.type = "product"
+        partner_id_number = self.env["res.partner.id_number"]
+        self.partner_1 = self.env.ref("base.res_partner_1")
+        self.partner_1.edifact_purchase_order_out = True
+        self.partner_2 = self.env.ref("base.res_partner_12")
+        self.exc_type_input = self.env.ref(
+            "edi_purchase_edifact_oca.edi_exchange_type_purchase_order_input"
+        )
```
thienvh332, line 34:
```suggestion
        )
        self.exc_type_output = self.env.ref("edi_purchase_edifact_oca.edi_exchange_type_purchase_order_out")

```

edi_purchase_edifact_oca/tests/test_edifact_purchase.py
```diff
@@ -0,0 +1,276 @@
...
+        # Add edi_exchange_record.id to UNT segment in expected_data
+        pattern = r"(UNT\+[^']*\+)[^']*(?=(?:'|$))"
+        expected_data = re.sub(
+            pattern,
+            lambda match: f"{match.group(1)}{exchange_record.id}",
+            expected_data,
+        )
+        self.assertEqual(
+            exchange_record.exchange_file, b64encode(expected_data.encode())
+        )
```
thienvh332, line 152:
```suggestion
        self.assertEqual(
            exchange_record.exchange_file, b64encode(expected_data.encode(self.exc_type_output.encoding)))

thienvh332 (commented):
> For me locally the unit test test_action_confirm() is failing in last assertion.

Hi @mathiasfrancke,
This issue was caused by my mistake, as I only tested with `ASCII` characters, while the encoding used by edi.exchange.type is `ISO-8859-1`.
You can fix this by applying the changes I suggested. The same fix should be applied in a few other similar places as well.

If you need anything, feel free to tag me :wink:

PR #1146, opened 2025-04-16
[17.0][FIX] edi icon

Gerloav:
New edi icon


simahawk:
Thanks for you contrib and sorry for being late on review!
We now have a new icon more inline w/ the new styles.
See https://github.com/OCA/edi/pull/1161

PR #1147, opened 2025-04-22
[15.0][FIX] edi_oca: fix search logic

ThiagoMForgeFlow:
Maintaining this part of code can occur a visual issue. If the `result` variable has 80 records (default by Odoo) and the `orig_ids` is less than result, Odoo only shows the `orig_ids`, and we lost the previous work finding next records.

@ForgeFlow

### DIFF
--- a/edi_oca/models/edi_exchange_record.py
+++ b/edi_oca/models/edi_exchange_record.py
@@ -524,8 +524,6 @@ class EDIExchangeRecord(models.Model):
                     access_rights_uid=access_rights_uid,
                 )[: limit - len(result)]
             )
-        # Restore original ordering
-        result = [x for x in orig_ids if x in result]
         return len(result) if count else list(result)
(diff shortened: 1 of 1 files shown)


LoisRForgeFlow:
Hi @simahawk 

We saw that you specifically added this change here https://github.com/OCA/edi/commit/2324d212a23249f809fcffa0f346e3470e6f8a1e

do you have some more insights?

LoisRForgeFlow:
@simahawk a kind reminder here

@etobella maybe you have some comments?

LoisRForgeFlow:
I'm not sure if the issues is clear, but it is quite problamatic. In a multicompany environment, if you only enable one company you won't see all the records that you should. Real example:

1. With only one company active you don't see all the records of that company.

image

2. You can only access all the records activiating all the companies.

image

PR #1153, opened 2025-05-08
[18.0][MIG] edi_voxel_sale_order_import_oca: Migration to version 18.0

juancarlosonate-tecnativa:
@tecnativa TT54159
@pedrobaeza @sergio-teruel


edi_voxel_sale_order_import_oca/models/sale_order.py
```diff
@@ -283,13 +285,16 @@ def _parse_product_voxel(self, line_vals, line_element):
                     self.env["sale.order"].browse(line_vals["order_id"]).partner_id
                 )
                 customerinfo = self.env["product.customerinfo"].search(
```
pedrobaeza, line 290: This makes this module depends on `product_customerinfo`, so please add it to the manifest.

edi_voxel_sale_order_import_oca/views/res_config_settings_views.xml
```diff
@@ -6,18 +6,17 @@
         <field name="name">res.config.settings.view.form.voxel.invoice</field>
         <field name="model">res.config.settings</field>
         <field name="inherit_id" ref="edi_voxel_oca.res_config_settings_view_form" />
-        <field
-            name="groups_id"
-            eval="[(4, ref('edi_voxel_oca.group_voxel_manager'))]"
-        />
         <field name="arch" type="xml">
```
pedrobaeza, line 9: Please transform this to the new setting tags.

pedrobaeza (changes requested):
You can't use f-strings on strings to translate, so please undo that part.

edi_voxel_sale_order_import_oca/models/sale_order.py
juancarlosonate-tecnativa: `edi_voxel_oca` depends on `product_customerinfo`

pedrobaeza: OK then, although it seems a weird dependency in the base module.

edi_voxel_sale_order_import_oca/views/res_config_settings_views.xml
juancarlosonate-tecnativa: All done, thx!

edi_voxel_sale_order_import_oca/models/sale_order.py
```diff
@@ -53,18 +53,18 @@ def create_document_from_xml(self, xml_content, voxel_filename, company):
...
             for error_msg in error_msgs:
-                str_error_msgs += "<li>%s</li>" % error_msg
-            create_msg += (
-                _(
-                    "<br/><span style='font-weight: bold;'>"
-                    "The following errors were found:</span><br/>"
-                    "<ul>%s</ul>"
-                )
-                % str_error_msgs
+                str_error_msgs += "<li>%s</li>", error_msg
```
pedrobaeza: This is not being translated, so you can use f-string here.

edi_voxel_sale_order_import_oca/models/sale_order.py
```diff
@@ -77,15 +77,20 @@ def _parse_voxel_order(self, order_file, order_filename, error_msgs, company):
...
+                raise UserError(
+                    self.env._("This XML file is not XML-compliant")
+                ) from exc
         else:
-            raise UserError(_("'%s' is not recognised as an XML file") % order_filename)
-        _logger.debug("Starting to import:%s" % (order_filename))
+            raise UserError(
+                self.env._("'%s' is not recognised as an XML file", order_filename)
+            )
+        _logger.debug("Starting to import:%s", order_filename)
```
pedrobaeza: The same here, and put an space after `:`

edi_voxel_sale_order_import_oca/models/sale_order.py
```diff
@@ -77,15 +77,20 @@ def _parse_voxel_order(self, order_file, order_filename, error_msgs, company):
...
         try:
             return self._parse_xml_order(xml_root, error_msgs, company)
         except Exception as exc:
             raise UserError(
-                _("Error creating the order") + f"\n\nFile contents:\n{order_file}"
+                self.env._("Error creating the order") + "\n\nFile contents:\n%s",
```
pedrobaeza: You can use f-string in the non-translatable part.

edi_voxel_sale_order_import_oca/models/sale_order.py
```diff
@@ -236,7 +241,7 @@ def _parse_partner_data_voxel(self, data, raise_error=True):
             return partner  # return the unique partner matching
         if raise_error:
             raise UserError(
-                _(
+                self.env._(
                     "Can't find a suitable partner for this data:\n\n%(data)s"
                     "\nResults: %(partner_count)s",
                     data=data,
```
pedrobaeza, line 247: Weird indentation in this block

juancarlosonate-tecnativa: i don't see what's wrong with it, anyway that's how the formatter does it

pedrobaeza: I mean this:

imagen

juancarlosonate-tecnativa: it looks like a visualization issue on GitHub 
image

pedrobaeza: Ah,OK, sorry for the noise.

pedrobaeza:
Moved to https://github.com/OCA/edi-voxel/pull/4

PR #1157, opened 2025-05-13
[15.0][IMP] base_import_pdf_by_template: Allow process PDF files attached in emails from mail alias

sergio-teruel:
cc @Tecnativa TT56271

ping @victoralmau @pilarvargas-tecnativa

### DIFF
--- a/base_import_pdf_by_template/models/mail_thread.py
+++ b/base_import_pdf_by_template/models/mail_thread.py
@@ -1,0 +1,50 @@
+# Copyright 2025 Tecnativa - Sergio Teruel
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+import logging
+
+from odoo import models
+
+_logger = logging.getLogger(__name__)
+
+
+class MailThread(models.AbstractModel):
+    _inherit = "mail.thread"
+
+    def message_new(self, msg_dict, custom_values=None):
+        process_pdf_template = custom_values.pop("process_pdf_template", None)
+        if process_pdf_template:
+            self = self.with_context(process_pdf_template=process_pdf_template)
+        return super().message_new(msg_dict, custom_values)
+
+    def _message_post_after_hook(self, new_message, message_values):
+        res = super()._message_post_after_hook(new_message, message_values)
+        if self.env.context.get("process_pdf_template", False):
+            # Process PDF attachments only if they come from an alias with
+            # ‘process_pdf_template’ in default values.
+            # One record is already created automatically from the mail alias, all
+            # other attachments are processed without the ‘record_ref’ key so that a
+            # new record is created for each attachment.
+            attachments = new_message.attachment_ids.filtered(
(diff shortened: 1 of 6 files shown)


base_import_pdf_by_template/models/mail_thread.py
```diff
@@ -0,0 +1,49 @@
...
+                pdf_upload_wiz = (
+                    self.env["wizard.base.import.pdf.upload"]
+                    .with_context(**ctx)
+                    .create(wiz_vals)
+                )
+                try:
+                    pdf_upload_wiz.action_process()
+                except (Exception) as err:
+                    _logger.error(err)
+        return res
```
victoralmau:
```suggestion
def _import_base_import_pdf_by_template(self, record, file_data, new=False):
        """Method to process the PDF if any template is available."""
        company = self.env.company
        if "company_id" in record._fields:
            company = record.company_id
        template_model = self.env["base.import.pdf.template"].with_company(
            company.id
        )
        total_templates = template_model.search_count([("model", "=", record._name)])
        if total_templates == 0:
            return False        
        wizard = self.env["wizard.base.import.pdf.upload"].create(
            {
                "model": record._name,
                "record_ref": f"{record._name},{record.id}",
                "attachment_ids": [(6, 0, file_data["attachment"].ids)],
            }
        )
        wizard.with_context(skip_template_not_found_error=True).action_process()
        return True
    
    def _message_post_after_hook(self, new_message, message_values):
        res = super()._message_post_after_hook(new_message, message_values)
        if self.env.context.get("process_pdf_template", False):
            # Process PDF attachments only if they come from an alias with
            # ‘process_pdf_template’ in default values.
            # One record is already created automatically from the mail alias, all
            # other attachments are processed without the ‘record_ref’ key so that a
            # new record is created for each attachment.
            attachments = new_message.attachment_ids.filtered(
                lambda x: x.mimetype == "application/pdf"
            )
            ctx = self.env.context.copy()
            ctx.pop("default_fetchmail_server_id", None)
            for _index, attachment in enumerate(attachments):
                self._import_base_import_pdf_by_template(
                    self, {"attachment": attachment}
                )
        return res
```

Something similar already exists in `base_import_pdf_by_template_account` ([v17](https://github.com/OCA/edi/blob/17.0/base_import_pdf_by_template_account/models/account_move.py#L9)), I think this change and the generic `_import_base_import_pdf_by_template()` method would help later to reduce the `base_import_pdf_by_template_account` code.

victoralmau (changes requested):

sergio-teruel:
I have seen that in v15 has code not prepared for your proposal.. So, I prefer merge this code in 15 and in the fw-port adapt to v17... Ok??

victoralmau (approved):

pilarvargas-tecnativa (approved):

pedrobaeza:
/ocabot merge patch

Please fw-port it if needed

PR #1159, opened 2025-06-03
[18.0][MIG] pdf_helper: Migration to 18.0

DantePereyra:
Standard migration to v18


pdf_helper/tests/test_helper.py
```diff
@@ -17,7 +15,7 @@ def read_test_file(filename, mode="r"):
         return thefile.read()
 
 
-class TestPDFHelperUtils(TestCase):
+class TestPDFHelperUtils(BaseCommon):
```
simahawk, line 18: I know we started adopting BaseCommon everywhere but in fact it does too many things and adds quite some overhead to test cases setup even when not needed. Also, in this particular case, there was a reason to use `TestCase`. This should be restored.

For the other cases, can you please follow this approach?
https://github.com/OCA/stock-logistics-orderpoint/pull/41

simahawk (changes requested):
Thanks for your work!

pdf_helper/tests/test_helper.py
simahawk: there's no need to disable tracking tho. I'm cleaning this up.

simahawk:
@DantePereyra thanks for your work. I'm moving on here https://github.com/OCA/edi/pull/1170. Please check my fix for test classes :wink:

PR #1160, opened 2025-06-03
Add many2many type when using a fixed value

pau-lafoz:
### Error description
When a line field is defined as "many2many" (for example tax_ids from invoice line) and has a fixed value, the following error is triggered during template use (doesn't happen during preview):
 `line 354, in _get_fixed_value
    f_name = self._get_fixed_field_name_ttype_mapped()[self.field_ttype]
             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^
KeyError: 'many2many'`

### Patch
Adding the `"many2many" : "fixed_value"` line avoids the error and writes the selected fixed value.

### Drawback
Only one value can be selected. For example, I haven't been able to select multiple tax ids in my case.

### DIFF
--- a/base_import_pdf_by_template/models/base_import_pdf_template.py
+++ b/base_import_pdf_by_template/models/base_import_pdf_template.py
@@ -348,6 +348,7 @@ class BaseImportPdfTemplateLine(models.Model):
             "text": "fixed_value_text",
             "json": "fixed_value_text",
             "many2one": "fixed_value",
+            "many2many": "fixed_value",
         }
 
     def _get_fixed_value(self):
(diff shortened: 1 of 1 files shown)


base_import_pdf_by_template/models/base_import_pdf_template.py
```diff
@@ -348,6 +349,7 @@ def _get_fixed_field_name_ttype_mapped(self):
             "text": "fixed_value_text",
             "json": "fixed_value_text",
             "many2one": "fixed_value",
+            "many2many": "fixed_value",
```
victoralmau, line 352: This probably solves the use case now, but I think it is not correct (for that reason this type of field is pending), you should create a `fixed_m2m_value_ids` field (or similar) related to the model that corresponds in each case (it will not always be the same), it would be a field m2mReference (I think it does not exist).

The approach is clear, to be able to set several values (tags for example) and that these values are applied in the registry that is created.

victoralmau (changes requested):

victoralmau:
The title of the PR should be: _[17.0][IMP] base_import_pdf_by_template: Add many2many compatibility_ (or similar)

PR #1163, opened 2025-06-06
[17.0][ADD] ediversa

manuelregidor:
This PR includes 3 modules:
- edi_ediversa_oca: base module with the API calls
- edi_ediversa_oca_sale_order_import: module to import sale order files from ediversa, process them and generate sale orders.
- edi_ediversa_oca_account_invoice_oca: module to generate EDI files to send to Ediversa when an invoice in validated.

T-5771


manuelregidor:
@ValentinVinagre

ValentinVinagre (changes requested):
@manuelregidor rename the following modules
- edi_ediversa_invoice_send -> edi_ediversa_**oca**_invoice_send
- edi_ediversa_sale_order_import -> edi_ediversa_**oca**_sale_order_import

manuelregidor:
@ValentinVinagre I have followed the same rule applied to the voxel modules. The general one has the "oca" element but the ones that depend on the general one do not contain that bit.

ValentinVinagre:
> @ValentinVinagre I have followed the same rule applied to the voxel modules. The general one has the "oca" element but the ones that depend on the general one do not contain that bit.

I prefer to have it by module order etc. Could you make the change?

manuelregidor:
@ValentinVinagre Changes applied

manuelregidor:
@ValentinVinagre Changes applied

javierizaca (approved):
LGTM

ValentinVinagre:
> * Requested changes with read-only permissions

Could you explain what tests you have performed?

ValentinVinagre:
> LGTM

Could you explain what tests you have performed?

javierizaca:
> > LGTM
> 
> Could you explain what tests you have performed?


The modules have been tested on an Odoo test server against a test environment provided by Ediversa. Within this framework, the following test scenarios have been executed:

###  Purchase Order Transmission

Sending purchase orders from the Ediversa platform using the same **Supplier Code** and varying:

- Invoice recipients  
- Delivery warehouses  
- Payment issuers  
- Buyers  
- Product codes  

###  Invoice Reception

Receiving invoices on the Ediversa platform under the same conditions, with the same **Supplier Code** and different:

- Invoice recipients  
- Delivery warehouses  
- Payment issuers  
- Buyers  
- Product codes  

###  Field Lengths

Invoice records were generated using values **exceeding character limits** accepted by Ediversa in the following fields:

- Legal buyer name  
- Delivery warehouse name  
- Invoice recipient name  
- Legal supplier name  
- Supplier name  
- Buyer  
- Product name  
- Order number  
- Invoice number  
- Delivery note number  

###  Error Scenarios

- No tax correlation configured between platforms  
- Ordering products **not present in Odoo** (by archiving products included in orders)

###  Adjustments and Improvements

The module set was progressively adjusted to meet the technical requirements of the Ediversa platform:

- Enforced character length limits  
- Logging of detailed error messages indicating processing failures (e.g., tax mismatch, missing product)

###  Final Validation

After all adjustments, approximately **175 distinct purchase orders**, based on real data from the platform, were successfully tested between an Odoo test instance and Ediversa’s production server.

ValentinVinagre (approved):
LGTM 👍🏻

ValentinVinagre:
@pedrobaeza @simahawk How do you see the module?

pedrobaeza:
I don't know about this, but I would say we should create a repository `edi-ediversa` to handle these modules separately from this main one for not being a crazy repository with hundred of modules, and then you can be maintainer of that modules. What do you think? If so, you can make the request in https://github.com/OCA/repo-maintainer-conf

ValentinVinagre:
> I don't know about this, but I would say we should create a repository `edi-ediversa` to handle these modules separately from this main one for not being a crazy repository with hundred of modules, and then you can be maintainer of that modules. What do you think? If so, you can make the request in https://github.com/OCA/repo-maintainer-conf

Sorry, let me explain better.
Ediversa (https://www.ediversa.com) is a company that connects systems (Corteingles, Leroy Merlin, etc.).
We've created these modules in this repository because we've seen other modules with similar functionality, such as 'voxel,' being added there. I don't think creating a new repository is necessary, as it won't have much more.
It's possible it could expand a bit more if it requires more functionality; I don't think it will be more than 1-2 modules. But this section would be the basic one.

pedrobaeza:
I would extract also Voxel modules (they are ours) to `edi-voxel`. Now that there's no problem and it's very easy to create new repositories (previously the administrative work was a pain and the critical thing for not doing that), I prefer to fragment more the scope of each repository, even if there will be 2 or 3 modules (or even 1 if it's very specific) - look for example mis-builder -.

ValentinVinagre:
> I would extract also Voxel modules (they are ours) to `edi-voxel`. Now that there's no problem and it's very easy to create new repositories (previously the administrative work was a pain and the critical thing for not doing that), I prefer to fragment more the scope of each repository, even if there will be 2 or 3 modules (or even 1 if it's very specific) - look for example mis-builder -.

I don't see any problem with it then. Should I move it this way then?

pedrobaeza:
If @simahawk and @etobella, which are the EDI masters, agree, then let's go!

etobella:
It makes sense to split it in different repos and leave this only for the generic set of addons

PR #1167, opened 2025-06-09
[18.0][MIG] base_business_document_import

simahawk:
And finally removed pdf_helper dep (see atomic commit).


base_business_document_import/models/business_document_import.py
```diff
@@ -0,0 +1,1391 @@
...
+                    "_match_currency",
+                    currency_dict,
+                    _(
+                        "The analysis of the business document returned '%s' "
+                        "as the country code to find the related currency. "
+                        "But there is no country with that code in Odoo."
+                    )
+                    % country_code,
+                )
+        if self._context.get("force_company"):
```
ivantodorovich: `force_company` is used here (and many other places in this module's code), but it's been deprecated by Odoo:

https://github.com/odoo/odoo/blob/68cdabc3217e557485fba7eceafbaff18c7f1a30/odoo/models.py#L6366-L6371

It's very likely dead code anyways, because in all cases it seems to fallback to self.env.company, which is the way to go.. but IMO it should be removed everywhere

base_business_document_import/models/business_document_import.py
```diff
@@ -0,0 +1,1391 @@
...
+            elif raise_exception:
+                raise UserError(
+                    _(
+                        "The record '{record}' is an instance of '{record_model}', "
+                        "not of '{target_model}'."
+                    ).format(
+                        record=record.display_name,
+                        record_model=record._name,
+                        target_model=model._name,
+                    )
```
ivantodorovich:
```suggestion
                    this.env._(
                        "The record '%(record)s' is an instance of '%(record_model)s', "
                        "not of '%(target_model)s'.",
                        record=record.display_name,
                        record_model=record._name,
                        target_model=model._name,
                    )
```

Two things here (and everywhere else):
- Use of `this.env._`
- Use the builting interpolation mechanism which lets odoo recover gracefully from bad translation texts

ivantodorovich (changes requested):

base_business_document_import/models/business_document_import.py
grindtildeath: :+1: 
But it's actually `self.env._`, you're doing too much JS :stuck_out_tongue_winking_eye:

base_business_document_import/models/business_document_import.py
```diff
@@ -0,0 +1,1391 @@
...
+                )
+            if record:
+                return record
+        if data_dict.get("xmlid"):
+            xmlid = data_dict["xmlid"]
+            record = False
+            try:
+                record = self.env.ref(xmlid, raise_if_not_found=True)
+            except Exception as e:
```
grindtildeath:
```suggestion
            except ValueError as e:
```

base_business_document_import/models/business_document_import.py
```diff
@@ -0,0 +1,1391 @@
...
+            return False
+        company_id = self._context.get("force_company") or self.env.company.id
+        bankaccount = rpbo.search(
+            [
+                "|",
+                ("company_id", "=", False),
+                ("company_id", "=", company_id),
+                ("sanitized_acc_number", "=", iban),
+                ("partner_id", "=", partner.id),
+            ],
```
grindtildeath: IMO would be nicer to use `odoo.osv.expression` helpers to build this domain as by reading it (without much Odoo knowledge) I wouldn't be sure if we want:
- `op1 OR op2 AND op3 AND op4`
- `op1 AND op2 OR op3 AND op4`
- `op1 AND op2 AND op3 OR op4`

By using helpers we would have:
```
>>> AND([OR([[("company_id", "=", False)], [("company_id", "=", 1)]]), [("sanitized_acc_number", "=", ""), ("partner_id", "=", 2)]])
['&', '|', ('company_id', '=', False), ('company_id', '=', 1), '&', ('sanitized_acc_number', '=', ''), ('partner_id', '=', 2)]
```
Which IMO is definitely more explicit :shrug:

base_business_document_import/models/business_document_import.py
```diff
@@ -0,0 +1,1391 @@
...
+            return product
+        product = self._match_product_search(product_dict)
+        if product:
+            return product
+        elif seller:
+            # WARNING: Won't work for multi-variant products
+            # because product.supplierinfo is attached to product template
+            sinfo = self.env["product.supplierinfo"].search(
+                self._match_company_domain()
+                + [
```
grindtildeath: `+` is not a safe operator to combine domains, better use those from `odoo.osv.expression`:
`AND([self._match_company_domain(), [("partner_id", "=", seller.id),("product_code", "=", product_dict["code"])]])`

base_business_document_import/models/business_document_import.py
```diff
@@ -0,0 +1,1391 @@
...
+                    ("partner_id", "=", seller.id),
+                    ("product_code", "=", product_dict["code"]),
+                ],
+                limit=1,
+            )
+            if (
+                sinfo
+                and sinfo.product_tmpl_id.product_variant_ids
+                and len(sinfo.product_tmpl_id.product_variant_ids) == 1
```
grindtildeath: This is a bit redundant, you could keep only `len(...)`

base_business_document_import/models/business_document_import.py
```diff
@@ -0,0 +1,1391 @@
...
+            ):
+                return sinfo.product_tmpl_id.product_variant_ids[0]
```
grindtildeath:
```suggestion
                return sinfo.product_tmpl_id.product_variant_id
```

base_business_document_import/models/business_document_import.py
```diff
@@ -0,0 +1,1391 @@
...
+            )
+        )
+        return self.env.ref("uom.product_uom_unit")
+
+    @api.model
+    def _match_taxes(
+        self, taxes_list, chatter_msg, type_tax_use="purchase", price_include=False
+    ):
+        """taxes_list must be a list of tax_dict"""
+        taxes_recordset = self.env["account.tax"].browse(False)
```
grindtildeath:
```suggestion
        taxes_recordset = self.env["account.tax"].browse()
```

base_business_document_import/models/business_document_import.py
```diff
@@ -0,0 +1,1391 @@
...
+        if tax_dict.get("unece_type_code"):
+            domain.append(("unece_type_code", "=", tax_dict["unece_type_code"]))
+        if tax_dict.get("unece_categ_code"):
+            domain.append(("unece_categ_code", "=", tax_dict["unece_categ_code"]))
+        if tax_dict.get("unece_due_date_code"):
+            tax_exigibility = ato._get_tax_exigibility_from_unece_code(
+                tax_dict["unece_due_date_code"]
+            )
+            if tax_exigibility:
+                domain = expression.AND(
```
grindtildeath, line 930: Looks weird to use `AND` after appending domain elements all the time before (which is fine if you want `AND` operator and you control all the domain operands as is the case here)

grindtildeath (approved):
Looks good, mostly some nitpicking on things unrelated to the migration itself. Feel free to ignore

base_business_document_import/models/business_document_import.py
ivantodorovich: 😆 you got me

base_business_document_import/models/business_document_import.py
simahawk: done

base_business_document_import/models/business_document_import.py
simahawk: The OR is just for the company. I wouldn't touch this right now.
But in other places is done... I'll review them.

simahawk:
@ivantodorovich @grindtildeath updated

twalter-c2c (approved):
Looks good, as far as I can tell.

base_business_document_import/models/business_document_import.py
```diff
@@ -0,0 +1,1379 @@
...
+            return product
+        elif seller:
+            # WARNING: Won't work for multi-variant products
+            # because product.supplierinfo is attached to product template
+            sinfo = self.env["product.supplierinfo"].search(
+                expression.AND(
+                    [
+                        self._match_company_domain(),
+                        [
+                            ("product_tmpl_id", "=", seller.product_tmpl_id.id),
```
twalter-c2c: Tests are failing on this one
` AttributeError: 'res.partner' object has no attribute 'product_tmpl_id'`

simahawk: it's fixed.

jbaudoux (approved):
lg

PR #1170, opened 2025-06-10
[18.0][MIG] pdf_helper: Migration to 18.0 

simahawk:
Replaces #1159 + fix test case perf


pedrobaeza:
I think this module should go to other repository or change its name. `pdf_helper` is too generic and not saying what is doing. I also don't understand too much the need of this for most of EDI processes.

simahawk:
In the ctx of EDI is used to get XML from PDF or add XML to PDF.
It could be renamed `pdf_xml_helper` or be moved to `server-tools` or `reporting-engine`.

jbaudoux:
> In the ctx of EDI is used to get XML from PDF or add XML to PDF. It could be renamed `pdf_xml_helper` or be moved to `server-tools` or `reporting-engine`.

pdf_xml_attachment in reporting-engine

simahawk:
Please review this one https://github.com/OCA/reporting-engine/pull/1031

PR #1176, opened 2025-06-15
[18.0][MIG] partner_identification_import

simahawk:


partner_identification_import/models/business_document_import.py
```diff
@@ -0,0 +1,66 @@
...
+                        ("is_company", "=", False),
+                    ]
+                    contact = self._match_partner_contact(
+                        partner_dict, chatter_msg, domain, order
+                    )
+                    if contact:
+                        return contact
+                    return id_number.partner_id
+                unmatched.append(
+                    _(
```
twalter-c2c:
```suggestion
                    self.env._(
```

partner_identification_import/models/business_document_import.py
```diff
@@ -0,0 +1,66 @@
...
+                        "ID Number: %(value)s\nID Number Category: %(schemeID)s\n\n",
+                        **ident,
+                    )
+                )
+            if unmatched:
+                raise self.user_error_wrap(
+                    "_hook_match_partner",
+                    partner_dict,
+                    _(
```
twalter-c2c:
```suggestion
                    self.env._(
```

rlimaeco (approved):

partner_identification_import/models/business_document_import.py
```diff
@@ -0,0 +1,66 @@
...
+                        "Odoo couldn't find a partner corresponding to the "
+                        "following information extracted from the business document:\n"
+                        "{}"
+                    )
+                    .format("or\n")
```
twalter-c2c, line 63: Am I right that `format` is not the preferred way to format strings (in favour of `%s` instead)?

twalter-c2c (approved):
LGTM, tiny comments.

partner_identification_import/models/business_document_import.py
simahawk: that's not for the translation, it's to format what comes after. The main problem here is that "or" is not translated but I won't fix it now.

PR #1181, opened 2025-06-25
[17.0][IMP+FIX] base_import_pdf_by_template_account: Suitable value of sort_weight in _unwrap_edi_attachments() method if applicable

victoralmau:
Suitable value of `sort_weight` in `_unwrap_edi_attachments()` method if applicable

**Example of use case**:
- Create a pdf template linked to invoices.
- Define an email alias to automatically create the invoices and process the attachments
- Send an email to the alias with 2 attachments: example.xml and example.pdf

**Before**:
The `_unwrap_edi_attachments()` method indirectly applied a lower sort_weight value (10) to the example.xml file than to the example.pdf file (20), which meant that the https://github.com/odoo/odoo/blob/399da69818d22860a808f5416829863f002ad38a/addons/account/models/account_move.py#L3202 condition was met for the example.pdf file and the pdf was not processed correctly (https://github.com/odoo/odoo/blob/399da69818d22860a808f5416829863f002ad38a/addons/account/models/account_move.py#L3215).

**After**:
The `_unwrap_edi_attachments()` method modifies the value of sort_weight in .pdf files if there is +1 file and if there is any pdf import template that applies to the corresponding record.

**Future**:
In v19 (master for now) the process has been improved using the `account.document.import.mixin` and https://github.com/odoo/odoo/blob/cf4e5b95c34357e5129adf9c78429f7c1a4ea33e/addons/account/models/account_document_import_mixin.py#L318 template.

@Tecnativa TT56274

### DIFF
--- a/base_import_pdf_by_template/wizards/wizard_base_import_pdf_upload.py
+++ b/base_import_pdf_by_template/wizards/wizard_base_import_pdf_upload.py
@@ -75,7 +75,8 @@ class WizardBaseImportPdfUpload(models.TransientModel):
         for line in self.line_ids.filtered("template_id"):
             try:
                 records += line.action_process()
-            except Exception:
+            except Exception as e:
+                logger.info(e)
                 if not self.env.context.get("skip_template_not_found_error"):
                     raise
         action = {
@@ -190,7 +191,8 @@ class WizardBaseImportPdfUploadLine(models.TransientModel):
         else:
             try:
                 setattr(_form, field_name, value)
-            except Exception:
+            except Exception as e:
+                logger.info(e)
                 self._add_log_error_text(field_name, value)
 
     def _process_form(self):
@@ -217,7 +219,8 @@ class WizardBaseImportPdfUploadLine(models.TransientModel):
                     field_value = model[field_name].browse(field_value)
                 try:
                     setattr(model_form, field_name, field_value)
-                except Exception:
+                except Exception as e:
+                    logger.info(e)
                     extra_vals[field_name] = ctx[key]
         # Set the values of the header in Form
         header_values = template._get_field_header_values(text)
@@ -240,7 +243,8 @@ class WizardBaseImportPdfUploadLine(models.TransientModel):
                     child_field_value = child_fixed_values[field_name]
                     try:
                         setattr(line_form, field_name, child_field_value)
-                    except Exception:
+                    except Exception as e:
+                        logger.info(e)
                         self._add_log_error_text(field_name, child_field_value)
                 # set the values of any line
                 for field_name in list(line.keys()):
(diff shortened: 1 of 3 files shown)


victoralmau:
Please @pedrobaeza and @sergio-teruel can you review it?

victoralmau:
Ping @pedrobaeza and @sergio-teruel

base_import_pdf_by_template_account/demo/base_import_pdf_template.xml
```diff
@@ -100,7 +100,7 @@
             name="child_field_id"
             ref="account.field_account_move__invoice_line_ids"
         />
-        <field name="auto_detect_pattern">(B 8 7 5 3 0 4 3 2)</field>
+        <field name="auto_detect_pattern">(B87530432)</field>
```
pedrobaeza, line 103: Shouldn't we pin the PyPDF version according Odoo upstream? Or is it aligned?

victoralmau: Odoo only has the PyPDF version fixed in some cases.: https://github.com/odoo/odoo/blob/032f6179dea0d6ec56cbb0d10d397d2d67a042ff/requirements.txt#L61

In any case, it hasn't been set yet, so I don't think it's a good idea to do it now with regard to very old versions.

pedrobaeza: I see all pinned, but they have changed between `PyPDF2` and `PyPDF` depending on the Python version.

victoralmau: pypdf only has python_version >= '3.13 set.

In addition to that, I think it would be a mistake to set the version in the base_import_pdf_by_template module now: What happens to those users who are using base_import_pdf_by_template without setting a version and have corrections applied (without knowing it) in versions higher than 5.4.0? (for example, errors in tests such as these)

pedrobaeza: I mean that the `import pypdf` clause in the module is for the library `PyPDF2` for all the Python versions except Python 3.13, or am I wrong?

victoralmau: I'm sorry, but I don't understand what you mean. 
The `base_import_pdf_by_template` module imports `pypdf` directly (https://github.com/OCA/edi/blob/b8832851a875f395a91596f7f21c0c2dfeec19aa/base_import_pdf_by_template/wizards/wizard_base_import_pdf_mixin. py#L7) and currently does not have the version set in the manifest; setting the version (previous to the current one) in the manifest now would be a mistake and would cause a multitude of problems for users who are using this module and regex (something I have already pointed out previously).

pedrobaeza: OK, I was wrong thinking that PyPDF2 was used in this module.

pedrobaeza (approved):

victoralmau:
Ping @sergio-teruel

victoralmau:
Please @cristina-hidalgo-tecnativa , can you review it?

cristina-hidalgo-tecnativa (approved):

PR #1182, opened 2025-06-30
[17.0][FIX] base_import_pdf_by_template: Apply sudo() to avoid access errors to obtain e.g. fixed values

victoralmau:
Apply `sudo()` to avoid access errors to obtain e.g. fixed values

Please @pedrobaeza can you review it?

@Tecnativa TT56930

### DIFF
--- a/base_import_pdf_by_template/models/base_import_pdf_template.py
+++ b/base_import_pdf_by_template/models/base_import_pdf_template.py
@@ -231,6 +231,9 @@ class BaseImportPdfTemplateLine(models.Model):
         domain="[('model_id.model', '=', search_field_relation)]",
         string="Search subfield",
     )
+    search_subfield_name = fields.Char(
+        related="search_subfield_id.name", string="Search subfield name"
+    )
     default_value = fields.Reference(
         selection="_selection_reference_value",
         string="Default value",
@@ -353,7 +356,10 @@ class BaseImportPdfTemplateLine(models.Model):
     def _get_fixed_value(self):
         self.ensure_one()
         f_name = self._get_fixed_field_name_ttype_mapped()[self.field_ttype]
-        f_value = self[f_name]
+        # Apply .sudo() because it is possible that the user executing code
+        # does not have read permissions on the model (for example:
+        # ir.module.fields.selection).
+        f_value = self.sudo()[f_name]
         if self.field_ttype == "selection":
             f_value = f_value.value
         elif self.field_ttype == "json":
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):
Clear patch, so fast-tracking it:

/ocabot merge patch

base_import_pdf_by_template/models/base_import_pdf_template.py
```diff
@@ -353,7 +353,10 @@ def _get_fixed_field_name_ttype_mapped(self):
     def _get_fixed_value(self):
         self.ensure_one()
         f_name = self._get_fixed_field_name_ttype_mapped()[self.field_ttype]
-        f_value = self[f_name]
+        # Apply .sudo() because it is possible that the user executing code
+        # does not have read permissions on the model (for example:
+        # ir.module.fields.selection).
+        f_value = self[f_name].sudo()
```
pedrobaeza: Overlooked...

```suggestion
        f_value = self.sudo()[f_name]
```

base_import_pdf_by_template_account/tests/test_base_import_pdf_by_template_account.py
```diff
@@ -13,6 +15,11 @@ class TestBaseImportPdfByTemplateAccount(BaseCommon):
     @classmethod
     def setUpClass(cls):
         super().setUpClass()
+        new_test_user(
+            cls.env,
+            login="test-account-user",
+            groups="account.group_account_invoice,analytic.group_analytic_accounting",
```
pedrobaeza, line 21: AFAIK, this doesn't depend on `account`, so you can't do this.

base_import_pdf_by_template/models/base_import_pdf_template.py
```diff
@@ -231,6 +231,9 @@ class BaseImportPdfTemplateLine(models.Model):
         domain="[('model_id.model', '=', search_field_relation)]",
         string="Search subfield",
     )
+    search_subfield_name = fields.Char(
+        related="search_subfield_id.name", string="Search subfield name"
```
pedrobaeza, line 235: Why this related field with no `related_sudo`?

base_import_pdf_by_template_account/tests/test_base_import_pdf_by_template_account.py
victoralmau: Yes, this module `base_import_pdf_by_template_account` depends on account (https://github.com/OCA/edi/blob/afafe7635377af88322541004a4b57c620530950/base_import_pdf_by_template_account/__manifest__.py#L9) and I have done this to test a process with a user without the Administration > Settings permission.

Precisely this has shown that https://github.com/OCA/edi/pull/1182#discussion_r2174892097 is also necessary, although it has indeed caught my attention that `related_sudo=True` is not necessary (it is true that it should also be necessary in many other fields and concepts).

This also verifies that https://github.com/OCA/edi/pull/1182#discussion_r2174847002 is working properly.

pedrobaeza: OK, I thought it was in the main module, not here.

pedrobaeza (approved):
/ocabot merge patch

PR #1187, opened 2025-07-15
[18.0] sale_order_import_ubl: tests disable tracking

simahawk:
and update missing contributors since a while


simahawk:
This will also make PRs depending on this module on v18 work (the nobump merge does not push to pypi).
/ocabot merge patch

PR #1197, opened 2025-07-24
[18.0] base_business_document_import: fix supplier product match

simahawk:
When wrapping the error was still using the seller.name relation which does not exists anymore.

Left over from v18 migration.


simahawk:
Fast tracking as it's broken now
/ocabot merge patch

PR #1201, opened 2025-07-30
[IMP] edi_oca: Add company field

PolComas:
The `edi.backend` model in edi_oca was missing a `company_id` Many2one field, which caused the contact form to crash in multi‑company environments.

### DIFF
--- a/edi_oca/models/edi_backend.py
+++ b/edi_oca/models/edi_backend.py
@@ -56,6 +56,7 @@ class EDIBackend(models.Model):
     """
     )
     active = fields.Boolean(default=True)
+    company_id = fields.Many2one("res.company", string="Company")
 
     def _get_component(self, exchange_record, key):
         record_conf = self._get_component_conf_for_record(exchange_record, key)


etobella:
@PolComas where does it fail? can you provide an example from the runboat?

etobella (changes requested):
Also, please provide a correct commit name like `[IMP] edi_oca: Add company field`.

PolComas:
Hi @etobella, thanks for looking at this. It turns out the failure only occurs in our migrated database (where the old rule_edi_backend_multi_company was still present). On a clean v15 install or on the runbot, I can’t reproduce the error at all—everything passes and no Invalid field edi.backend.company_id ever shows up. In other words, it wasn’t a bug in the core OCA code but a leftover migration artifact on our side, sorry for the confusion.

I’m going to withdraw the PR, since it isn’t needed.

For reference, the only error was:
ValueError: Invalid field edi.backend.company_id in leaf ('company_id','in',[…])

PR #1205, opened 2025-08-15
[16.0][IMP] account_invoice_ubl: add support to self-billing

jbaudoux:
Allow to generate self-billing invoice and credit note in UBL.

https://docs.peppol.eu/poacc/self-billing/3.0/bis-sb/

cc @lmignon @rousseldenis @sbidoul 

Also fixed in:
* account_invoice_ubl:
  * due date : only declare if there is an amount_residual

### DIFF
--- a/account_invoice_ubl/models/account_move.py
+++ b/account_invoice_ubl/models/account_move.py
@@ -307,6 +307,9 @@ class AccountMove(models.Model):
             line_root,
             ns,
             type_="sale",
+            seller=None,
+            customer=self.partner_id,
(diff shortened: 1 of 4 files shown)


lmignon:
IMO, this addon should depend on [account_invoice_supplier_self_invoice](https://github.com/OCA/account-invoicing/tree/16.0/account_invoice_supplier_self_invoice). Personally, I would have preferred the term `self_billing` to be used in the name of this existing module and for the fields defined in it instead of `self_invoice`, but despite this, it would seem logical to me to depend on this module.

jbaudoux:
Includes https://github.com/OCA/edi/pull/1287
To rebase after that other PR is merged

PR #1206, opened 2025-08-18
[17.0][FIX] edi_voxel_oca: _render_qweb_xml require report ref

christian-ramos-tecnativa:
Since https://github.com/OCA/reporting-engine/commit/3dfa12cea0caee22cfea332b7968803789e55be3 the function `_render_qweb_xml` is `api.model` and requires the ref string.

### DIFF
--- a/edi_voxel_oca/models/voxel_mixin.py
+++ b/edi_voxel_oca/models/voxel_mixin.py
@@ -72,8 +72,9 @@ class VoxelMixin(models.AbstractModel):
 
     def _get_and_send_voxel_report(self, report_name):
         self.ensure_one()
-        report = self.env.ref(report_name)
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Does it make sense to move voxel modules to https://github.com/OCA/edi-voxel also in this version? We have made it for 18.0.

christian-ramos-tecnativa:
I saw it, based on this fix We can assume that no one is using the module in v17.0 as is not functional. I don't mind moving it to https://github.com/OCA/edi-voxel, but it will require creating the `17.0` branch there

christian-ramos-tecnativa:
Branch creation in https://github.com/OCA/repo-maintainer-conf/pull/108

pedrobaeza:
Created, so you can propose the module with the whole history + this patch.

PR #1213, opened 2025-09-15
[18.0][IMP] acc_einvoice_generate: make move types selection overridable

bealdav:
Allow to manage specific case

### DIFF
--- a/account_einvoice_generate/models/account_move.py
+++ b/account_einvoice_generate/models/account_move.py
@@ -23,7 +23,7 @@ class AccountMove(models.Model):
         if (
             xml_format
             and xml_format != "none"
-            and self.move_type in ("out_invoice", "out_refund")
+            and self.move_type in self._select_move_type_for_xml_in_pdf()
             and self.partner_id
             and self.state != "cancel"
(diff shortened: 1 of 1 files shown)


account_einvoice_generate/models/account_move.py
```diff
@@ -23,7 +23,7 @@ def _xml_format_in_pdf_invoice(self):
         if (
             xml_format
             and xml_format != "none"
-            and self.move_type in ("out_invoice", "out_refund")
+            and self.move_type in self._select_moves_type_for_xml_in_pdf()
```
hparfr:
```suggestion
            and self.move_type in self._select_move_type_for_xml_in_pdf()
```

To be more constistent with the field

hparfr (changes requested):

account_einvoice_generate/models/account_move.py
bealdav: Thanks, done

hparfr (approved):
LGTM
thanks

TB-Ph35 (approved):
LGTM

PR #1224, opened 2025-10-06
[14.0][IMP] account_invoice_facturx: add support for negative price

alexis-via:
This is a backport of PR #1139 
It includes a fix for LineTotalAmount when AllowanceTotalAmount is not null


jbaudoux (commented):
There are pre-commit issues :/

PR #753, opened 2023-05-04
[16.0][MIG] account_invoice_import

alexis-via:
This is a draft PR. Many more changes are planned.

PR #1238, opened 2025-11-17
[16.0][MIG] account_invoice_download

alexis-via:
Migration from 14 to 16. Updated to newer version of account_invoice_import, as proposed in https://github.com/OCA/edi/pull/753


bguillot (approved):
Used in production, it works well

PR #1252, opened 2025-11-27
[18.0][IMP] account_invoice_export: add timeout config in UI

TDu:


sebalix (approved):

account_invoice_export/models/transmit_method.py
```diff
@@ -15,6 +15,11 @@ class TransmitMethod(models.Model):
     destination_url = fields.Char(string="Url")
     destination_user = fields.Char(string="User", copy=False)
     destination_pwd = fields.Char(string="Password", copy=False)
+    export_connection_timeout = fields.Integer(
```
simahawk, line 18: this should go in the base transmit method module, no?

twalter-c2c (commented):
LGTM. I don't know if it should go to another module as suggested by Simone.

account_invoice_export/models/transmit_method.py
TDu: Well the base module does not handle the communication so I would say no.

twalter-c2c (approved):

PR #1253, opened 2025-12-02
[18.0][FIX] account_invoice_facturx: fix support for non-VAT taxes

alexis-via:
Remove configuration parameter facturx_refund_type, to simplify the code

FYI, I plan to also remove the configuration parameter "facturx_level", and generate only the extended profile. If you disagree with this change, please yell now!


alexis-via:
I also removed the config parameter for the Factur-X profile. The goal was to simplify the code.

rlemeur (approved):

PR #1267, opened 2025-12-22
[16.0][ADD] account_edi_ubl_cii_purchase_match

sbejaoui:
Extend UBL vendor bill import to automatically match and link bill lines to purchase order lines using the OrderReference and product label


sbejaoui:
cc/ @phschmidt , @jbaudoux , @rousseldenis

phschmidt (approved):
Functional review

account_edi_ubl_cii_purchase_match/models/account_edi_xml_ubl_20.py
```diff
@@ -0,0 +1,111 @@
...
+        order_ref = invoice_origin_node.text
+        self._match_invoice_to_purchase_order(invoice, order_ref)
+        return res
+
+    def _match_invoice_to_purchase_order(self, invoice, order_ref):
+        purchase_order = self.env["purchase.order"].search(
+            [
+                "|",
+                ("name", "=", order_ref),
+                ("partner_ref", "=", order_ref),
```
jbaudoux: `partner_ref` should compare to `OrderReference/SalesOrderID` and not  `OrderReference/ID`

account_edi_ubl_cii_purchase_match/models/account_edi_xml_ubl_20.py
```diff
@@ -0,0 +1,111 @@
...
+class AccountEdiXmlUBL20(models.AbstractModel):
+
+    _inherit = "account.edi.xml.ubl_20"
+
+    def _import_fill_invoice_form(self, journal, tree, invoice, qty_factor):
+        invoice = invoice.with_context(no_purchase_set=True)
+        res = super()._import_fill_invoice_form(journal, tree, invoice, qty_factor)
+        if journal.type != "purchase":
+            return res
+        invoice_origin_node = tree.find("./{*}OrderReference/{*}ID")
```
jbaudoux: Maybe you should only search for the parent here as ID is mandatory when the parent is there and then parse ID and SalesOrderID

account_edi_ubl_cii_purchase_match/models/account_edi_xml_ubl_20.py
```diff
@@ -0,0 +1,111 @@
...
+        if invoice_origin_node is None:
+            return res
+        order_ref = invoice_origin_node.text
```
jbaudoux: then handover this to the called method below

account_edi_ubl_cii_purchase_match/models/account_edi_xml_ubl_20.py
```diff
@@ -0,0 +1,111 @@
...
+            self._match_invoice_line_to_purchase_order_line(
+                invoice_line, purchase_order.order_line
+            )
+        return True
+
+    def _match_invoice_line_to_purchase_order_line(self, invoice_line, purchase_lines):
+        if invoice_line.purchase_line_id:
+            return False
+        product = self._get_matching_product(invoice_line)
+        purchase_line = purchase_lines.filtered(lambda line: line.product_id == product)
```
jbaudoux: What if no product is found and there are PO lines without product?
```suggestion
        purchase_line = purchase_lines.filtered(lambda line: line.product_id and line.product_id == product)
```

account_edi_ubl_cii_purchase_match/models/account_edi_xml_ubl_20.py
```diff
@@ -0,0 +1,111 @@
...
+    def _import_fill_invoice_line_form(
+        self, journal, tree, invoice, invoice_line, qty_factor
+    ):
+        res = super()._import_fill_invoice_line_form(
+            journal, tree, invoice, invoice_line, qty_factor
+        )
+        supplier_product_code = self._find_value(
+            "./cac:Item/cac:SellersItemIdentification/cbc:ID", tree
+        )
+        if invoice_line.product_id or not supplier_product_code:
```
jbaudoux: There is I think a bug in the call to super that retrieves the product. It's searching for a product with default_code based on the SellersItemIdentification while it should be BuyersItemIdentification
https://github.com/odoo/odoo/blob/f07949e4ccdb5d7a17a6678b008fa6785ec3fb12/addons/account_edi_ubl_cii/models/account_edi_xml_ubl_20.py#L727
I would force a new call to the method `_retrieve_product` with the right argument

account_edi_ubl_cii_purchase_match/models/account_edi_xml_ubl_20.py
```diff
@@ -0,0 +1,111 @@
...
+    def _match_invoice_line_to_purchase_order_line(self, invoice_line, purchase_lines):
+        if invoice_line.purchase_line_id:
+            return False
+        product = self._get_matching_product(invoice_line)
+        purchase_line = purchase_lines.filtered(lambda line: line.product_id == product)
+        invoice_line._set_product(product)
+        invoice_line.purchase_line_id = purchase_line
+        return True
+
+    def _get_matching_product(self, invoice_line):
```
jbaudoux: What's the purpose of this? The product matching should already be resolved by `_retrieve_product` and what you improved in `_import_fill_invoice_line_form`?

What could be usefull is to match a PO line name based on the invoice line name, in case the seller or buyer does not manage the line with a product

jbaudoux:
@sbejaoui Can you attend comments?

account_edi_ubl_cii_purchase_match/models/account_edi_xml_ubl_20.py
sbejaoui: I will use what base module stored in invoice_origin

sbejaoui:
pipeline is red due to another module #1313

lmignon (approved):
code review + functional tests

jbaudoux (approved):

PR #1274, opened 2026-01-08
[16.0][FIX] account_invoice_import_simple_pdf: Update test years to current range

MarwanBHL:
Fix `test_invoice_number_parsing` test that fails since January 1, 2026.


alexis-via (approved):

alexis-via:
FYI, the cause of this problem is here:

https://github.com/OCA/edi/blob/16.0/account_invoice_import_simple_pdf/models/account_invoice_import_simple_pdf_invoice_number.py#L12

PR #1275, opened 2026-01-09
[18.0][FIX] account_invoice_import_simple_pdf: Update test years to current range

alexis-via:


alexis-via:
I'll make another implementation

PR #1281, opened 2026-01-18
[18.0][IMP] account_invoice_import: Very simple fix of tests to be independent of demo data

arnaudlayec:
Before : tests were failing because `demo` data are not loaded in tests anymore and of the reference to `base.res_partner_1` and `base.res_partner_2`
Now : removed update of `supplier_rank` or `customer_rank`, since tests are running OK without it


flotho (approved):
easy code review
LGTM

alexis-via (approved):

PR #1283, opened 2026-01-23
[16.0][FIX] account_invoice_download: fix access right issue when accessing partner form view

alexis-via:
fix access right issue when accessing partner form view partner form view without rights on account.invoice.download.config


arnaudlayec (approved):
Looks good to me

PR #1284, opened 2026-01-23
[18.0][FIX] account_invoice_download: fix access right issue when accessing partner form view

alexis-via:
Fix access right issue when accessing partner form view without rights on account.invoice.download.config

Fw port of https://github.com/OCA/edi/pull/1283


arnaudlayec (approved):
Looks good to me

PR #1287, opened 2026-01-30
[16.0][FIX] base_ubl: TaxExemptionReason - use odoo VATEX selection

jbaudoux:
Allow to use odoo standard VATEX exemption reason that should be part of any declared exempt tax, otherwise provide the corresponding exemption reason

Do not provide VAT Party identifiers when the invoice is outside scope of VAT. Add the company registry.

Remove VAT name, it's not part of peppol

In case of non-VAT tax converted to AllowanceCharge, provide the tax description instead of "Miscellaneous"

cc @nicolas-delbovier-acsone @sbejaoui @phschmidt

### DIFF
--- a/account_invoice_ubl/models/account_move.py
+++ b/account_invoice_ubl/models/account_move.py
@@ -307,6 +307,9 @@ class AccountMove(models.Model):
             line_root,
             ns,
             type_="sale",
+            seller=None,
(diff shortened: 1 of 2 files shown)


lmignon (approved):
All this makes sense and works in production

PR #1290, opened 2026-02-10
[17.0][IMP] base_import_pdf_by_template: Use finditer() + groups() to define the value from text

victoralmau:
FWP from 15.0: https://github.com/OCA/edi/pull/1183

Use finditer() + groups() to define the value from text

Please @pedrobaeza and @sergio-teruel can you review it?

@Tecnativa TT56971


pedrobaeza (approved):
The same as 15, so fast-tracking it:

/ocabot merge patch

victoralmau:
Can you try the command again to merge it?

PR #1296, opened 2026-02-17
[17.0][FIX] base_import_pdf_by_template: Avoid the error caused by the wizard's default_model value

victoralmau:
Avoid the error caused by the wizard's default_model value

Example use case:
- Go to Sales > Import records (with template)
- Select a .pdf in the wizard
- Click the Import button

```
if model._fields[field_name].type == "many2one":
    KeyError: 'model'
```

It's already done in v18: https://github.com/OCA/edi/commit/613734ead5405dfb3535c96987217a74478b9532#diff-8bc11e59f7d505f9e63bcb8f45dcbda86d4cf06ff596456bd174697ee8934298R197-R226

Please @pedrobaeza and @cristina-hidalgo-tecnativa can you review it?

@Tecnativa TT61034

### DIFF
--- a/base_import_pdf_by_template/wizards/wizard_base_import_pdf_upload.py
+++ b/base_import_pdf_by_template/wizards/wizard_base_import_pdf_upload.py
@@ -215,7 +215,10 @@ class WizardBaseImportPdfUploadLine(models.TransientModel):
             if key.startswith("default_") and key != "default_fetchmail_server_id":
                 field_name = key.replace("default_", "")
                 field_value = ctx[key]
-                if model._fields[field_name].type == "many2one":
+                if (
+                    field_name in model._fields
+                    and model._fields[field_name].type == "many2one"
+                ):
                     field_value = model[field_name].browse(field_value)
                 try:
                     setattr(model_form, field_name, field_value)


pedrobaeza (approved):
As preventive measure, but the template won't work correctly until fixed. Shouldn't a `UserError` should be raised instead telling about that?

victoralmau:
I don't quite understand the UserError you want to display. 
It's something that shouldn't happen regardless of how the template is set. In this case, the template is correct, but when you open the wizard to add files, default_model=xxx (sale.order, for example) is added to context, and we most likely won't need that default (because the model field doesn't exist in sale.order). That's why this change is added.

pedrobaeza:
Ah, OK, I misunderstood the patch. Then what you have to do is to clean context (there's a method for that), not to ignore possible errors in the definition of the template.

victoralmau:
> Ah, OK, I misunderstood the patch. Then what you have to do is to clean context (there's a method for that), not to ignore possible errors in the definition of the template.

I don't think it's right (quite the opposite, in fact) to clear the context, because there will be situations where you want to use the context (for example, when using the wizard from the Customer Invoices or Supplier Invoices menu).

pedrobaeza:
OK then. I was thinking in migrations where a field put in the template has disappeared, and a way to notify the user, but that can be other improvement.

victoralmau:
> OK then. I was thinking in migrations where a field put in the template has disappeared, and a way to notify the user, but that can be other improvement.

Exactly, that would be another improvement. In any case, it is important to remember https://github.com/OCA/edi/pull/1296#issuecomment-3914107998 (even if no error is visible, it is reflected internally).

cristina-hidalgo-tecnativa (approved):

PR #1300, opened 2026-02-23
[16.0][FIX] account_invoice_import: do not use `taxes` and `account` of `import_config` over ones in `lines`

arnaudlayec:
**Issue description**
When passing the `parsed_inv` of an invoice with `taxes` in defined `lines`, the value of line's `taxes` are replaced by the ones of `import_config`.
This may even happens if `import_config` is empty but the partner already has an invoice in Odoo: `import_config` is generated from the previous invoice of the partner.

**Suggested fix**
We want the real-data of `taxes` and `account` passed in `lines` to take the precedence on any passed `import_config` or guessed from partner's previous invoice.

This commit defaults to `import_config` taxes and account as *last* possible values.

Forward-port in v18.0 : https://github.com/OCA/edi/pull/1288

### DIFF
--- a/account_invoice_import/wizard/account_invoice_import.py
+++ b/account_invoice_import/wizard/account_invoice_import.py
@@ -442,6 +442,7 @@ class AccountInvoiceImport(models.TransientModel):
                 )
             if not product and import_config.get("product"):
                 product = import_config["product"]
+            account, taxes = None, None
             if product:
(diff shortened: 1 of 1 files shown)


arnaudlayec:
Hello @alexis-via , might you have a quick look on this simple PR?

PR #1302, opened 2026-02-27
[18.0][FIX] account_invoice_import: Use correct fiscal position

mt-software-de:
Right only the fiscal position is used if it is directly set on the partner. 
In a normal odoo environment it isn't the case that the fiscal position needs to be set,
because it is selected automatically. 

This change uses the method `_get_fiscal_position` which either returns the manually set fiscal position or the automatically detected on.

### DIFF
--- a/account_invoice_import/readme/CONTRIBUTORS.md
+++ b/account_invoice_import/readme/CONTRIBUTORS.md
@@ -4,3 +4,4 @@
 - Yannick Vaucher \<<yannick.vaucher@camptocamp.com>\>
 - Ronald Portier \<<ronald@therp.nl>\>
 - Simone Orsi \<<simone.orsi@camptocamp.com>\>
+- Michael Tietz (MT Software) \<<mtietz@mt-software.de>\>
--- a/account_invoice_import/wizard/account_invoice_import.py
+++ b/account_invoice_import/wizard/account_invoice_import.py
@@ -1,5 +1,6 @@
 # Copyright 2015-2021 Akretion France (http://www.akretion.com/)
 # Copyright 2020-2021 Therp BV (https://therp.nl)
+# Copyright 2026 Michael Tietz (MT Software) <mtietz@mt-software.de>
 # @author: Alexis de Lattre <alexis.delattre@akretion.com>
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
@@ -364,6 +365,24 @@ class AccountInvoiceImport(models.TransientModel):
             vals["check_total"] = parsed_inv["amount_total"]
         return vals
 
+    def _get_fiscal_position(self, import_config, partner, partner_shipping=None):
+        return (
+            self.env["account.fiscal.position"]
+            .with_company(import_config["company"])
+            ._get_fiscal_position(partner, partner_shipping)
+        )
+
+    def _map_account_and_taxes(self, import_config, partner, account, taxes):
+        res = account, taxes
+        if not partner:
+            return res
+        fiscal_position = self._get_fiscal_position(import_config, partner)
+        if not fiscal_position:
+            return res
+        new_account = fiscal_position.map_account(account)
+        new_taxes = fiscal_position.map_tax(taxes)
+        return new_account, new_taxes
+
     @api.model
     def _prepare_line_vals_1line(self, parsed_inv, import_config, vals, partner):
         il_vals = {
@@ -394,10 +413,9 @@ class AccountInvoiceImport(models.TransientModel):
                 account = import_config["account"]
             if import_config.get("taxes"):
                 taxes = import_config["taxes"]
-        fp = partner and partner.property_account_position_id or False
-        if fp:
-            account = fp.map_account(account)
-            taxes = fp.map_tax(taxes)
+        account, taxes = self._map_account_and_taxes(
+            import_config, partner, account, taxes
+        )
         il_vals.update(
             {
                 "account_id": account.id,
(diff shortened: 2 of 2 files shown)


account_invoice_import/wizard/account_invoice_import.py
```diff
@@ -1,5 +1,6 @@
 # Copyright 2015-2021 Akretion France (http://www.akretion.com/)
 # Copyright 2020-2021 Therp BV (https://therp.nl)
+# Copyright 2026 Michael Tietz (MT Software) <mtietz@mt-software.de>
```
arnaudlayec, line 3: I'm not sure if such a modification involves copyrighting the module?

account_invoice_import/readme/CONTRIBUTORS.md
```diff
@@ -4,3 +4,4 @@
 - Yannick Vaucher \<<yannick.vaucher@camptocamp.com>\>
 - Ronald Portier \<<ronald@therp.nl>\>
 - Simone Orsi \<<simone.orsi@camptocamp.com>\>
+- Michael Tietz (MT Software) \<<mtietz@mt-software.de>\>
```
arnaudlayec, line 7: I'm not sure if such a modification involves being in contributors list?

account_invoice_import/wizard/account_invoice_import.py
```diff
@@ -364,6 +365,24 @@ def _prepare_create_invoice_vals(self, parsed_inv, import_config):
...
+        )
+
+    def _map_account_and_taxes(self, fiscal_position, account, taxes):
+        if not fiscal_position:
+            return account, taxes
+        account = fiscal_position.map_account(account)
+        taxes = fiscal_position.map_tax(taxes)
+        return account, taxes
+
+    def map_account_and_taxes(self, import_config, partner, account, taxes):
```
arnaudlayec, line 382: This method should be private (start with a `_`)
I don't think it is necessary to split in 3 methods the tasks handle here. For code readability, I would suggest to merge `_get_fiscal_position`, `_map_account_and_taxes` and `map_account_and_taxes` in a single method called `_map_account_and_taxes`

account_invoice_import/wizard/account_invoice_import.py
```diff
@@ -364,6 +365,24 @@ def _prepare_create_invoice_vals(self, parsed_inv, import_config):
...
+        fiscal_position = self._get_fiscal_position(import_config, partner)
```
arnaudlayec, line 383: If `partner` is empty, there should be a shortcut before calling `_get_fiscal_position` or , like:
```
if not partner:
     return account, taxes
```

The case where `partner` is empty may occur and is handled by the module (field `account_move.import_partner_data`)

arnaudlayec (changes requested):
Hello,
Thank you for this change. I agree it is really needed because in previous code, if `partner.property_account_position_id` is empty, the fiscal position of the invoice can be set wrongly.
A call to the native method `res_partner._get_fiscal_position` should indeed replace the `partner.property_account_position_id`

I have a few suggestions to let the code simple to read

account_invoice_import/readme/CONTRIBUTORS.md
mt-software-de: This is common practice.

account_invoice_import/wizard/account_invoice_import.py
mt-software-de: This is common practice.

account_invoice_import/wizard/account_invoice_import.py
mt-software-de: IMO not. Each method is dedicated tow one separate mechanism.

account_invoice_import/wizard/account_invoice_import.py
mt-software-de: Yes that can be done. I will remove the check in `_map_account_and_taxes` and add it here.

account_invoice_import/wizard/account_invoice_import.py
```diff
@@ -364,6 +365,24 @@ def _prepare_create_invoice_vals(self, parsed_inv, import_config):
             vals["check_total"] = parsed_inv["amount_total"]
         return vals
 
+    def _get_fiscal_position(self, import_config, partner):
+        return (
+            self.env["account.fiscal.position"]
+            .with_company(import_config["company"])
+            ._get_fiscal_position(partner)
```
arnaudlayec, line 372: Last though: managing the case where the delivery address' country is different than the partner's country would complete this new feature. In such case, the fiscal position guessing considers the delivery country instead of the main partner's country (already in native `account_fiscal_position._get_fiscal_position()`).

Knowing this module does not handle the import of `partner_shipping_id` anyway, this proposal could just be skipped now and planned for another PR.
But I think it's cheap to add in this `account_invoice_import._get_fiscal_position()` method an optional `partner_shipping` arg defaulting to None, used in call to the native method like `._get_fiscal_position(partner, partner_shipping)`, allowing for further inheritance

account_invoice_import/wizard/account_invoice_import.py
arnaudlayec: Ok, why not
The `_` at beginning of method's name remains needed

arnaudlayec (approved):
All good for me, I think it is very helpful

mt-software-de:
@alexis-via could you maybe give this PR a second review?

account_invoice_import/wizard/account_invoice_import.py
mt-software-de: `_` is not needed. It depends if you want to declare a method as private or not. 

But for now i merged everything into two methods to get your approval, and i don't need a good "hook ability" for now.

mt-software-de:
@OCA/edi-maintainers can someon merge this one?

PR #1310, opened 2026-03-16
[ADD] account_edi_retrieve_partner_from_purchase_order

sbejaoui:
Retrieves the partner from the related purchase order if set in UBL document


jbaudoux (commented):
Is it necessary to have the same partner than the PO?
When ordering, you have the partner to who you place the order, and you have the invoicing partner. They can be different. It's like that on an SO. When you receive a vendor bill, you can received it from the invoicing department that is different from the contact of the PO.
Does that causes an issue?

jbaudoux:
> Is it necessary to have the same partner than the PO? When ordering, you have the partner to who you place the order, and you have the invoicing partner. They can be different. It's like that on an SO. When you receive a vendor bill, you can received it from the invoicing department that is different from the contact of the PO. Does that causes an issue?

@sbejaoui Can you check this question?

sbejaoui:
> Is it necessary to have the same partner than the PO?
> When ordering, you have the partner to who you place the order, and you have the invoicing partner. They can be different. It's like that on an SO. When you receive a vendor bill, you can received it from the invoicing department that is different from the contact of the PO.
> Does that causes an issue?

In general, no, we do not need to have the same partner as the PO.

The issue is more practical than functional. In some messy databases, with duplicate or misconfigured partners (e.g. VAT configured on the wrong contact, inconsistent emails/names, etc.), we don't fully control the partner matching process during invoice import. The system searches by VAT first, then name, email, and other criteria, and simply takes the first matching partner.

One way to make the matching more reliable is to retrieve the partner directly from the purchase order using its unique reference. So this module is not strictly necessary in a clean database, but it becomes useful in some real-life cases to avoid incorrect partner matching.

phschmidt (approved):
Functional tests

rousseldenis (approved):

PR #1311, opened 2026-03-18
[16.0][ADD] account_edi_ubl_move_line_uom_and_packaging_unece

sbejaoui:
Adds support for UNECE unit codes during UBL invoice import to automatically detect and assign the appropriate unit of measure or product packaging on invoice lines.

When a UNECE code matches a product packaging, the packaging is set with priority. Otherwise, the corresponding unit of measure is applied based on the UNECE code.


jbaudoux (changes requested):
There are plenty of unece codes for a pallet. You need m2m between packaging level and unece codes to be able to match a level based on the given unece code in the bill.

jbaudoux (approved):
LGTM

account_edi_ubl_move_line_uom_and_packaging_unece/models/account_edi_xml_ubl_20.py
```diff
@@ -0,0 +1,44 @@
...
+        if uom:
+            invoice_line.product_uom_id = uom
+            return True
+        return False
+
+    def _import_fill_invoice_line_packaging(self, invoice_line, unit_code):
+        product_packaging = self.env["product.packaging"].search(
+            [
+                ("product_id", "=", invoice_line.product_id.id),
+                ("packaging_level_id.unece_type_ids.code", "=", unit_code),
```
jbaudoux:
```suggestion
                ("packaging_level_id.unece_type_ids.code", "=", f"X{unit_code}"),
```

jbaudoux (changes requested):
As discussed. When packaging codes are used as unit code, they are prefixed by "X"

account_edi_ubl_move_line_uom_and_packaging_unece/models/account_edi_xml_ubl_20.py
```diff
@@ -0,0 +1,49 @@
...
+                # we therefore check both formats (e.g. "8A" and "X8A")
+                ("packaging_level_id.unece_type_ids.code", "=", unit_code),
+                ("packaging_level_id.unece_type_ids.code", "=", f"X{unit_code}"),
+            ],
+            limit=1,
+        )
+        if product_packaging:
+            invoice_line.product_packaging_id = product_packaging
+            invoice_line.product_packaging_qty = invoice_line.quantity
+            return True
```
lmignon: @sbejaoui the `unit_code` into the file is prefixed with 'X'...... if the unit of measure is a packaging.. 
Therefore, you should look for a packaging **only if** the unit_code starts with 'X' for a packaging code without X....

```suggestion
        if not self._import_fill_invoice_line_packaging(invoice_line, unit_code) and unit_code.startswith("X"):
            # the unitCode attr is based on UNECE Rec 20 (UoM)
            # when it's prefixed with "X" Rec 21 'packaging' is used as unit of measure
            # we therefore check for packaging with code without the 'X' prefix
            self._import_fill_invoice_line_uom(invoice_line, unit_code.lstrip("X"))
        return res

    def _import_fill_invoice_line_uom(self, invoice_line, unit_code):
        uom = self.env["uom.uom"].search([("unece_code", "=", unit_code)], limit=1)
        if uom:
            invoice_line.product_uom_id = uom
            return True
        return False

    def _import_fill_invoice_line_packaging(self, invoice_line, unit_code):
        product_packaging = self.env["product.packaging"].search(
            [
                ("product_id", "=", invoice_line.product_id.id),
                ("packaging_level_id.unece_type_ids.code", "=", unit_code),
            ],
            limit=1,
        )
        if product_packaging:
            invoice_line.product_packaging_id = product_packaging
            invoice_line.product_packaging_qty = invoice_line.quantity
            return True
```

lmignon (changes requested):
@sbejaoui Is-it a misunderstanding on my side or the way you detect the packaging is not the right one?

lmignon:
https://github.com/OCA/community-data-files/pull/268 is merged

account_edi_ubl_move_line_uom_and_packaging_unece/models/account_edi_xml_ubl_20.py
sbejaoui: good catch, it should be `unit_code.lstrip("X")`, not `f"X{unit_code}"`

That said, we can't assume that codes starting with X are packaging codes only, REC 20 also contains codes starting with X, so X is not a reliable packaging marker by itself

my understanding is that, in this case, the X prefix is a UBL/Peppol convention used when Rec 21 packaging codes are carried in the unitCode field

plz check now


image

account_edi_ubl_move_line_uom_and_packaging_unece/models/account_edi_xml_ubl_20.py
```diff
@@ -0,0 +1,49 @@
...
+        self, journal, tree, invoice, invoice_line, qty_factor
+    ):
+        res = super()._import_fill_invoice_line_form(
+            journal, tree, invoice, invoice_line, qty_factor
+        )
+        billed_quantity_node = tree.find("./{*}InvoicedQuantity")
+        unit_code = billed_quantity_node.attrib.get("unitCode")
+        if not unit_code:
+            return res
+        if not self._import_fill_invoice_line_packaging(invoice_line, unit_code):
```
lmignon: @sbejaoui Line packaging is a less common case than a simple uom unit. We should first look for uom and then for packaging

lmignon (changes requested):
see my last comment

account_edi_ubl_move_line_uom_and_packaging_unece/models/account_edi_xml_ubl_20.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def _import_fill_invoice_line_uom(self, invoice_line, unit_code):
+        uom = self.env["uom.uom"].search([("unece_code", "=", unit_code)], limit=1)
+        if uom:
+            invoice_line.product_uom_id = uom
+            return True
+        return False
+
+    def _import_fill_invoice_line_packaging(self, invoice_line, unit_code):
+        product_packaging = self.env["product.packaging"].search(
```
lmignon, line 61: on a performance point of view, these look up by unit_code on `uom.uom` and `product.packaging` should use the ormcache...

```python

@api.model
@ormcache()
def _get_uom_id_by_unece_code(self): -> dict[str, int]
    uoms = self.env["uom.uom"].search([("unece_code", "!=", False)])
    return {u.unece_code: u.id for u in uoms}

def _get_uom_for_unece_code(self, unece_code):
    id = self.__get_uom_id_by_unece_code().get(unece_code, None)
    result = self.env["uom.uom"]
    if id:
        result = result.browse(id)
    return result

....


```

account_edi_ubl_move_line_uom_and_packaging_unece/models/account_edi_xml_ubl_20.py
```diff
@@ -0,0 +1,56 @@
...
+                ("product_id", "=", invoice_line.product_id.id),
+                "|",
+                # the unitCode attr is based on UNECE Rec 20 (UoM)
+                # when Rec 21 (packaging) codes are used, they may be prefixed with "X"
+                # we therefore check both formats (e.g. "8A" and "X8A")
+                ("packaging_level_id.unece_type_ids.code", "=", unit_code),
+                ("packaging_level_id.unece_type_ids.code", "=", unit_code.lstrip("X")),
+            ],
+            limit=1,
+        )
```
lmignon: @sbejaoui IMO, we should also have a cache pour level_id by unece_type_ids code....otherwise the request generated by odoo (at least in 16.0) is ugly.

account_edi_ubl_move_line_uom_and_packaging_unece/models/account_edi_xml_ubl_20.py
```diff
@@ -0,0 +1,56 @@
...
+            return res
+        if not self._import_fill_invoice_line_packaging(invoice_line, unit_code):
+            self._import_fill_invoice_line_uom(invoice_line, unit_code)
+        return res
+
+    @api.model
+    @ormcache()
+    def _get_uom_by_unece_code(self):
+        uoms = self.env["uom.uom"].search([("unece_code", "!=", False)])
+        return {u.unece_code: u.id for u in uoms}
```
lmignon: A good practice is to put this kind of method into the model itself. You must also pay attention to invalidate the cache on 'create', 'write' and 'unlink' on the model.

lmignon (changes requested):

account_edi_ubl_move_line_uom_and_packaging_unece/models/account_edi_xml_ubl_20.py
```diff
@@ -0,0 +1,53 @@
...
+        ) + pl_model._get_packaging_level_ids_by_unece_code(unit_code.lstrip("X"))
+        if not packaging_level_ids:
+            return False
+        product_packaging = self.env["product.packaging"].search(
+            [
+                ("product_id", "=", invoice_line.product_id.id),
+                ("packaging_level_id", "in", packaging_level_ids),
+            ],
+            limit=1,
+        )
```
lmignon:
```suggestion
        packaging_level_id = pl_model._get_packaging_level_ids_by_unece_code(
            unit_code
        ) or pl_model._get_packaging_level_ids_by_unece_code(unit_code.lstrip("X"))
        if not packaging_level_id:
            return False
        product_packaging = self.env["product.packaging"].search(
            [
                ("product_id", "=", invoice_line.product_id.id),
                ("packaging_level_id", "=", packaging_level_id),
            ],
            limit=1,
        )
```

lmignon (changes requested):
2 files seem missing 😏

account_edi_ubl_move_line_uom_and_packaging_unece/models/account_edi_xml_ubl_20.py
sbejaoui: the helper is done in base modules
https://github.com/OCA/community-data-files/pull/270
https://github.com/OCA/community-data-files/pull/271

sbejaoui: The relation between unece codes and packaging levels is m2m, so multiple packaging levels can share the same UNECE code, the filtering is then done at the product level

account_edi_ubl_move_line_uom_and_packaging_unece/models/account_edi_xml_ubl_20.py
```diff
@@ -0,0 +1,53 @@
...
+        return False
+
+    def _import_fill_invoice_line_packaging(self, invoice_line, unit_code):
+        # the unitCode attr is based on UNECE Rec 20 (UoM)
+        # when Rec 21 (packaging) codes are used, they may be prefixed with "X"
+        # we therefore check both formats (e.g. "8A" and "X8A")
+        pl_model = self.env["product.packaging.level"]
+        packaging_level_ids = pl_model._get_packaging_level_ids_by_unece_code(
+            unit_code
+        ) + pl_model._get_packaging_level_ids_by_unece_code(unit_code.lstrip("X"))
```
jbaudoux:
```suggestion
        packaging_level_ids = pl_model._get_packaging_level_ids_by_unece_code(
            unit_code
        ) if not unit_code.startswith("X") else pl_model._get_packaging_level_ids_by_unece_code(unit_code.lstrip("X"))
```

jbaudoux: or simply :)
```suggestion
        packaging_level_ids = pl_model._get_packaging_level_ids_by_unece_code(unit_code.lstrip("X"))
```

sbejaoui: We should search for both, since some packaging codes already start with X. The X prefix is a convention, not a rule, so we can't assume all suppliers will follow it

image

check now

https://github.com/acsone/edi/blob/b010f6eddedb3a772e2021a810406c016f8038cf/account_edi_ubl_move_line_uom_and_packaging_unece/models/account_edi_xml_ubl_20.py#L36-L40

lmignon (approved):

jbaudoux (approved):

account_edi_ubl_move_line_uom_and_packaging_unece/models/account_edi_xml_ubl_20.py
jbaudoux: Yes but the unit code never starts with X. So it should work, like you did with this final version

PR #1312, opened 2026-03-23
[16.0][ADD] account_edi_ubl_cii_no_embed_facturx

jbaudoux:
This addon prevents to embed a factur-x attachment inside every invoice pdf
by deleting the definition of factur-x as edi format.

See https://github.com/odoo/odoo/blob/9fd78da68bc6167c99932fbc5ec7b1d8e264d5ea/addons/account_edi_ubl_cii/models/ir_actions_report.py#L116

cc @sbejaoui


sbejaoui (approved):

account_edi_ubl_cii_no_embed_facturx/data/account_edi_format.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo noupdate="1">
+    <delete model="account.edi.format" id="account_edi_ubl_cii.edi_facturx_1_0_05" />
```
lmignon, line 3: To prevent any error in case of the xml id no more exists, you should use the search parameter to look for the record to delete...
```suggestion
    <delete model="account.edi.format" search="[('id', '=', ref('account_edi_ubl_cii.edi_facturx_1_0_05')]" />
```

lmignon (changes requested):

account_edi_ubl_cii_no_embed_facturx/data/account_edi_format.xml
jbaudoux: That's why it's in noupdate and odoo is not raising an error but a warning:
WARNING: Skipping deletion for missing XML ID `'account_edi_ubl_cii.edi_facturx_1_0_05'`

jbaudoux: I tested your proposal and it's adding an additional traceback with
```
ValueError: <class 'ValueError'>: "External ID not found in the system: account_edi_ubl_cii.edi_facturx_1_0_05" while evaluating
"[('id', '=', ref('account_edi_ubl_cii.edi_facturx_1_0_05'))]"
```

So I keep my approach :)

jbaudoux:
This module will certainly make other modules in this repo using factur-x failing

lmignon:
> This module will certainly make other modules in this repo using factur-x failing

You should add an exclude rule and modify the CI to isolate the tests

PR #1320, opened 2026-04-08
[17.0][FIX] account_invoice_facturx: fix Factur-X 4.x schematron errors

tendil:
With factur-x 4.x, schematron validation became stricter and the module
started generating EN16931 XML that no longer passed validation for
credit transfer invoices.

Two concrete problems caused that failure:

- the seller contact email was exported with a schemeID attribute on
  URIID, although this attribute is not allowed in that context;
- when the creditor account was not an IBAN, the module did not export
  a fallback account identifier, so the payment means block failed
  BR-CO-27.

As a result, invoices that were previously generated successfully with
factur-x 3.x started failing during Factur-X XML generation with
factur-x 4.x.

---

This patch fixes the XML generation logic by:
- removing the invalid schemeID attribute from the seller email URIID;
- exporting ProprietaryID for non-IBAN creditor accounts;
- raising a clear UserError when no account identifier is available,
  instead of generating XML that is known to be invalid.

The test suite is updated accordingly:
- the default invoice fixture now includes a valid creditor account;
- a regression test checks that URIID is generated without schemeID;
- a regression test checks that non-IBAN accounts are exported as
  ProprietaryID;
- a regression test checks that missing account identifiers are rejected
  explicitly.

Task: 5347

### DIFF
--- a/account_invoice_facturx/models/account_move.py
+++ b/account_invoice_facturx/models/account_move.py
@@ -268,6 +268,15 @@ class AccountMove(models.Model):
         So it's difficult to have a common datamodel for it"""
         return False
 
+    def _cii_get_delivery_date(self):
+        """Return the delivery/service date to export in Factur-X XML.
+
+        Designed to be inherited by modules that store a dedicated delivery
+        or service date on invoices.
+        """
+        self.ensure_one()
+        return self.invoice_date
+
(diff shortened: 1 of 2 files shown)


alexis-via:
FYI, we stop using account_invoice_facturx and we now use account_invoice_en16931 (https://github.com/akretion/fr-einvoicing)

PR #1327, opened 2026-04-23
[19.0][MIG] purchase_order_ubl: Migration to 19.0

yankinmax:
Depends on:
- [ ] https://github.com/OCA/reporting-engine/pull/1157


simahawk:
I'd recommend using https://github.com/OCA/reporting-engine/tree/18.0/pdf_xml_attachment (used by other modules in EDI).

yankinmax:
> I'd recommend using https://github.com/OCA/reporting-engine/tree/18.0/pdf_xml_attachment (used by other modules in EDI).

With the help of Codex I've found same approach in core:
https://github.com/odoo/odoo/blob/6da43a511444441724733141708e245cc390ae7b/addons/sale/models/ir_actions_report.py#L4

Ricardoalso (approved):
Technically LGTM

purchase_order_ubl/tests/common.py
```diff
@@ -0,0 +1,180 @@
...
+                "name": "VAT purchase 18.0%",
+                "description": "VAT-buy-18.0",
+                "type_tax_use": "purchase",
+                "amount": 18,
+                "amount_type": "percent",
+                "tax_exigibility": "on_invoice",
+                "unece_type_id": cls.vat_tax_type.id,
+                "unece_categ_id": cls.s_tax_categ.id,
+            }
+        )
```
vvrossem: `cls.tax_21` but the values indicate 18

purchase_order_ubl/tests/common.py
```diff
@@ -0,0 +1,180 @@
...
+
+class PurchaseOrderUblCommon(HttpCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
+        cls.uom_unit = cls.env.ref("uom.product_uom_unit")
+        cls.uom_hour = cls.env.ref("uom.product_uom_hour")
+        cls.vat_tax_type = cls.env.ref("account_tax_unece.tax_type_vat")
+        cls.s_tax_categ = cls.env.ref("account_tax_unece.tax_categ_s")
```
vvrossem, line 19: `account_tax_unece` is not in the `__manifest__.py` dependencies.
But instead of adding a dependency, you could create those record from scratch

purchase_order_ubl/tests/common.py
yankinmax: Good catch. Thank you for the review

purchase_order_ubl/tests/common.py
yankinmax: I've double checked and we need that dependency for the `account_tax_unece::unece_type_id` and `account_tax_unece::unece_categ_id` fields.
Thanks @vvrossem

yankinmax:
> I'd recommend using https://github.com/OCA/reporting-engine/tree/18.0/pdf_xml_attachment (used by other modules in EDI).

Applied your suggestion and migrated `pdf_xml_attachment`:
- [ ] https://github.com/OCA/reporting-engine/pull/1157

vvrossem (approved):

simahawk (approved):

PR #1330, opened 2026-04-23
[19.0][MIG] account_edi_no_autocreate_partner

sbejaoui:


phschmidt (approved):
Functional tests and review

account_edi_no_autocreate_partner/models/res_partner.py
```diff
@@ -0,0 +1,49 @@
...
+    def _retrieve_partner(
+        self, name=None, phone=None, email=None, vat=None, domain=None, company=None
+    ):
+        partner = super()._retrieve_partner(
+            name=name, phone=phone, email=email, vat=vat, domain=domain, company=company
+        )
+        if not partner:
+            partner = self.env.ref(
+                "account_edi_no_autocreate_partner.partner_not_found"
+            )
```
lmignon, line 48: @sbejaoui Since you modify the default behavior, a config parameter should activate this way of doing.

lmignon (changes requested):
Useful!

PR #879, opened 2023-11-29
[16.0][MIG] despatch_advice_import

thienvh332:
Migrate from:
- https://github.com/OCA/edi/pull/280

Depends on:
- https://github.com/OCA/edi/pull/884

PR #1332, opened 2026-04-24
[19.0][MIG] despatch_advice_import: Migration to 19.0

yankinmax:
Depends on:
- [ ] https://github.com/OCA/reporting-engine/pull/1157


ivantodorovich (approved):

despatch_advice_import/wizard/despatch_advice_import.py
```diff
@@ -211,102 +283,183 @@ def _process_picking_done(self, move):
...
+        self,
+        stock_moves: models.Model,
+        parsed_order_document: dict,
+        forced_qty=False,
+    ):
+        """Mark moves as delivered, optionally with an over-delivered quantity.
+
+        :param stock_moves: The open stock moves linked to the purchase order line.
+        :param parsed_order_document: The normalized despatch advice data being applied.
+        :param forced_qty: The delivered quantity to force on the first move, if any.
```
vvrossem: why only on the first move?

vvrossem (approved):
Code LGTM

despatch_advice_import/wizard/despatch_advice_import.py
yankinmax: This is a mistake indeed.
I've updated the code and a comment describing the process.
You're really a good reviewer, thanks.

test-requirements.txt
```diff
@@ -0,0 +1 @@
+odoo-addon-pdf_xml_attachment @ git+https://github.com/OCA/reporting-engine.git@refs/pull/1157/head#subdirectory=pdf_xml_attachment
```
cyrilmanuel: @yankinmax  can be removed as merged and rebase the PR :)

cyrilmanuel (approved):

yankinmax:
Hello @OCA/edi-maintainers  can you pls take a look and possibly trigger merge?

despatch_advice_import/readme/CREDITS.md
```diff
@@ -1,2 +0,0 @@
-The migration of this module from 10.0 to 16.0 was financially supported
```
simahawk: do not remove this

simahawk (approved):
LG overall. 
@jbaudoux ping

despatch_advice_import/wizard/despatch_advice_import.py
```diff
@@ -0,0 +1,453 @@
...
+        :param move: A stock move belonging to the picking that should be validated.
+        :return: True when the picking only contains canceled moves, else None.
+        """
+        picking = move.picking_id
+        if all(line.state == "cancel" for line in picking.move_ids):
+            return True
+        # skip backorder wizard
+        picking.with_context(
+            skip_immediate=True, skip_backorder=True, skip_sms=True, skip_expired=True
+        ).button_validate()
```
grindtildeath: ⚠️ 
We do not want to validate the incoming picking on the reception of despatch advice. I'm not sure if this must be configurable in the module or if it just shouldn't be here at all.

If that is really needed, IMO we should have a "post process hook" in this module that can be redefined in another module to process the picking as done.

jbaudoux: There have been changes in the migration commit to v16 that are out of the scope of the migration and that should be reverted. Validing the picking makes no sense and some other features like having one picking per reception date seems to have been lost

yankinmax: Hello @jbaudoux I see your point.
Can you probably specify the workflow of this module as I just did a technical migration and completely not familiar with the module itself?
That way I could revert to the main ideas of the module.

Ricardoalso: For the context, it was added in https://github.com/OCA/edi/pull/879/changes/f9a4c0f39164b1e89c3f014bad6140f5570dbf53#diff-ba317b47ca5c669f7eb8f157d21581ec7a424e672eccfdf785c0cf2dc97517aaR191

Ricardoalso: I was asked to take over this open point and just commited a proposition 
f90cf9c to handle this concern. 
Picking validation becomes configurable via context, with backward-compatible fallback. Default flow does not apply implicit validation, and custom modules can preserve legacy behavior by overriding process_data and enabling the context flag.

Open to suggestions

jbaudoux: @Ricardoalso Fine for me.

You also need to skip the quantity done in the 2 methods
https://github.com/OCA/edi/commit/f9a4c0f39164b1e89c3f014bad6140f5570dbf53#diff-ba317b47ca5c669f7eb8f157d21581ec7a424e672eccfdf785c0cf2dc97517aaR220

`move.quantity_done = move.product_qty`

Ricardoalso: Thanks for the prompt response @jbaudoux 
I will continue with the refactoring then

jbaudoux:
Not linked to the migration, but I see no mention/WARNING in the readme regarding the fact that the parsed uom is not evaluated. So it assumes that the despatch advice will have lines in the same uom as the PO. Be carefull with that!

From experience, the unece code put as uom in the parsed result could relate to a packaging. As well as the gtin (product identifier) that could relate to a packaging with the unece as EA (each).
For example you order 80 units (= 1 pallet) and the supplier despatch you 1 pallet. In this case qty=1 should translate to 80.

Ricardoalso:
> Not linked to the migration, but I see no mention/WARNING in the readme regarding the fact that the parsed uom is not evaluated. So it assumes that the despatch advice will have lines in the same uom as the PO. Be carefull with that!
> 
> From experience, the unece code put as uom in the parsed result could relate to a packaging. As well as the gtin (product identifier) that could relate to a packaging with the unece as EA (each). For example you order 80 units (= 1 pallet) and the supplier despatch you 1 pallet. In this case qty=1 should translate to 80.

Not linked to the migration indeed, but thanks for raising this out ! 
We should maybe depend on https://github.com/OCA/community-data-files/tree/19.0/uom_unece and use the `uom.uom.unece_code` field declared in there. WDYT? 

```
def process_data(self, parsed_order_document: dict):
  ...
  if line_info["qty"] or line_info["backorder_qty"]:
      self._check_uom_match(line, line_info)
```

But as already mentioned, since it is a separate concern, it should be part of another PR

despatch_advice_import/readme/CONTRIBUTORS.md
```diff
@@ -0,0 +1,9 @@
+- Laurent Mignon \<laurent.mignon@acsone.eu\>
+
+- [Trobz](https://trobz.com):
+
+  > - Thien \<thienvh@trobz.com\>
+
+- Simone Orsi \<simone.orsi@camptocamp.com\>
+
+- Jacques-Etienne Baudoux \<je@bcim.be\>
```
ivantodorovich, line 9: @Ricardoalso and @yankinmax add yourselves there

ivantodorovich (approved):

ivantodorovich:
@simahawk can you trigger the merge pls ? 🙏🏻

PR #1337, opened 2026-05-06
[18.0][ADD] account_edi_ubl_cii_payment_unece

sebalix:
Integrate the UNECE payment means ([`account_payment_unece`](https://github.com/OCA/community-data-files/tree/18.0/account_payment_unece)) with Odoo standard module `account_edi_ubl_cii`.


account_edi_ubl_cii_payment_unece/models/account_edi_ubl.py
```diff
@@ -0,0 +1,39 @@
...
+        payment_mean_code = None
+        for node in tree.findall("./{*}PaymentMeans/{*}PaymentMeansCode"):
+            if note := node.text:
+                payment_mean_code = note
+                break
+        if not payment_mean_code:
+            return
+        # Look for a matching payment method line
+        payment_method_line = self.env["account.payment.method.line"].search(
+            [
```
jbaudoux, line 29: To not match the payment means for customer invoices

```suggestion
            [
            ("payment_type", "=", "outbound"),
```

account_edi_ubl_cii_unece/readme/DESCRIPTION.md
```diff
@@ -0,0 +1,3 @@
+Integrate UNECE Payment Means (module `account_payment_unece` from
+[OCA/community-data-files project](https://github.com/OCA/community-data-files/))
+with Odoo standard UBL/CII electronic invoices (module `account_edi_ubl_cii`).
```
jbaudoux:
```suggestion
with Odoo standard UBL/CII electronic invoices (module `account_edi_ubl_cii`).

When using for example SEPA direct debit, your can configure the corresponding UNECE code on the payment method and this will be declared properly in the electronic invoice.
Also, when receiving an invoice declared with a payment means SEPA direct debit, you can configure an outbound payment means with that UNECE code and the created invoice will have that payment means set so that you know you don't have to pay it. 
```

jbaudoux (changes requested):
Thanks. LG.

Can you rename the module to ` account_edi_ubl_cii_payment_unece´ to remove confusion with the module implementing tax_unece.

Can you also squash fixups?

account_edi_ubl_cii_payment_unece/models/account_edi_ubl.py
sebalix: Added for both UBL and CII imports.

jbaudoux (approved):

jbaudoux:
@sbejaoui Can you review this one and, if approved, backport it to v16.0 ?

sbejaoui (approved):

jbaudoux:
@sebalix I just tested the module. An `account.payment.method` cannot be created manually.

Something needs to provide a supplier SEPA direct debit payment method data entry so that it can be configured.

Like this https://github.com/OCA/bank-payment-alternative/blob/18.0/account_payment_sepa_direct_debit/data/account_payment_method.xml but in outbound (name=SEPA Direct Debit for suppliers). I think it make sense to add it in this module. WDYT?

jbaudoux:
> Like this https://github.com/OCA/bank-payment-alternative/blob/18.0/account_payment_sepa_direct_debit/data/account_payment_method.xml but in outbound (name=SEPA Direct Debit for suppliers). I think it make sense to add it in this module. WDYT?

@sebalix We even need 2 data entries as there are 2 UNECE code for direct debit: 59 or 49 https://docs.peppol.eu/poacc/billing/3.0/codelist/UNCL4461/

jbaudoux:
Tested. Works fine once the missing data entry is added
image

Would also be worth adding a note that the payment methods can be managed by installing the module https://github.com/OCA/bank-payment-alternative/tree/18.0/account_payment_base_oca
Otherwise, you can't configure it properly.

sebalix:
@jbaudoux yes, that would mean adding UNECE module(s) dependency to such modules (those in OCA/bank-payment-alternative, and also OCA/bank-payment). Or adding new glue module(s) completing the missing unece config on payment methods.

jbaudoux:
> @jbaudoux yes, that would mean adding UNECE module(s) dependency to such modules (those in OCA/bank-payment-alternative, and also OCA/bank-payment). Or adding new glue module(s) completing the missing unece config on payment methods.

My idea was to add the data in this module as this doesn't require a new dependency. And for having the menu to configure payment method lines, just add a note in the README (maybe in a separate commit as this is relevant to v18 and not before). I think only OCA/bank-payment-alternative uses the native odoo account.payment.method.line. It's then up to the user to install it or not.

account_edi_ubl_cii_payment_unece/data/account_payment_method.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!-- Copyright 2026 Akretion (https://www.akretion.com).
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
+<odoo noupdate="1">
+    <record id="direct_debit" model="account.payment.method">
+        <field name="name">Direct Debit to suppliers</field>
+        <field name="code">direct_debit</field>
```
jbaudoux, line 7: For the code you use here, you need to extend the method in AccountPaymentMethod to have the AccountPaymentMethodLine created

```
class AccountPaymentMethod(models.Model):
    _inherit = "account.payment.method"

    @api.model
    def _get_payment_method_information(self):
        res = super()._get_payment_method_information()
        res["direct_debit"] = {"mode": "multi", "type": ("bank",)}
        res["direct_debit_sepa"] = {"mode": "multi", "type": ("bank",)}
        return res
```

like here:
https://github.com/OCA/bank-payment-alternative/blob/18.0/account_payment_sepa_direct_debit/models/account_payment_method.py#L41

account_edi_ubl_cii_payment_unece/data/account_payment_method.xml
```diff
@@ -0,0 +1,18 @@
...
+        <field name="payment_type">outbound</field>
+        <field name="unece_id" ref="account_payment_unece.payment_means_49" />
+    </record>
+
+    <record id="direct_debit_sepa_" model="account.payment.method">
+        <field name="name">SEPA Direct Debit to suppliers</field>
+        <field name="code">direct_debit_sepa</field>
```
jbaudoux, line 14: @alexis-via As we need different code for sepa direct debit inbound and outbound (because inbound requires a pain version and outbound doesn't - also it is in different modules) and you already used
sepa_direct_debit for inbound and sepa_credit_transfer for outbound, we've put direct_debit_sepa for outbound.

alexis-via: For me, for SEPA direct debit outbound, the code is "manual", using the native account.payment.method declared in the account module here: https://github.com/odoo/odoo/blob/18.0/addons/account/data/account_data.xml#L135
When there is no bank file to generated, using the account.payment.method with code "manual" is the good implementation for me. You can have several account.payment.method.line that use the same account.payment.method with code "manual".

jbaudoux: @alexis-via The issue is that the unece code is on account.payment.method and there is a unicity constrain per code, payment_type (in/out). So we have to put a different code for each account.payment.method having a different unece code.

jbaudoux (approved):

PR #1339, opened 2026-05-12
[18.0][IMP] account_invoice_import: Make import compatible with odoo default behavior

mt-software-de:
Right now it is not possible to use odoo default's bill factur-x import and oca simple-pdf import.
Since `create_document_from_attachment` replaced within addon `account_invoice_import`.
With this change, `account_invoice_import` checks if there is a decoder for the attachment,
if so it means odoo process can be used.


mt-software-de:
@arnaudlayec can you maybe give this one a review?

PR #1341, opened 2026-05-13
[18.0][IMP] sale_order_import: allow creating invoice/delivery partners on the fly

SilvioC2C:


sale_order_import/wizard/sale_order_import.py
```diff
@@ -712,3 +750,101 @@ def _post_error_lines_message(self, parsed_order, order):
...
+
+    def _create_missing_invoice_partner_values(
+        self, invoice_to_dict, partner, chatter_msg
+    ):
+        """Prepares a new invoice partner's values for the current import"""
+        vals = invoice_to_dict.copy()
+        vals["type"] = "invoice"
+        vals["parent_id"] = partner.id
+        if country_code := vals.pop("country_code", None):
+            country = self.env["res.country"].search(
```
simahawk, line 793: why not using `bdio._get_country_filter|_get_country_state_filter`?

sale_order_import/wizard/sale_order_import.py
```diff
@@ -712,3 +750,101 @@ def _post_error_lines_message(self, parsed_order, order):
...
+
+        When called upon a record, checks field "create_missing_invoice_partner".
+        When called upon an empty recordset (eg: as a model method), checks context's
+        key "create_missing_invoice_partner".
+
+        Hook method, can be overridden by inheriting modules.
+        """
+        if self:
+            return self.create_missing_invoice_partner
+        return bool(self.env.context.get("create_missing_invoice_partner"))
```
simahawk, line 766: There's no need for a specific flag:
* this piece of code runs only w/ an initialized wizard.
* the flag on the wizard controls the behavior.
* the behavior of the flag can be controller via context by `default_create_missing_invoice_partner`

```suggestion
        return self and self.create_missing_invoice_partner
```

sale_order_import/wizard/sale_order_import.py
```diff
@@ -712,3 +750,101 @@ def _post_error_lines_message(self, parsed_order, order):
...
+
+        When called upon a record, checks field "create_missing_shipping_partner".
+        When called upon an empty recordset (eg: as a model method), checks context's
+        key "create_missing_shipping_partner".
+
+        Hook method, can be overridden by inheriting modules.
+        """
+        if self:
+            return self.create_missing_shipping_partner
+        return bool(self.env.context.get("create_missing_shipping_partner"))
```
simahawk, line 818:
```suggestion
        return self and self.create_missing_shipping_partner
```

sale_order_import/tests/test_order_import.py
```diff
@@ -258,3 +258,114 @@ def test_order_import_log_errored_line(self):
...
+        """
+        parsed_order = dict(
+            self.parsed_order,
+            invoice_to={
+                "country_code": "FR",
+                "email": "test@invoice.partner",
+                "name": "Test Invoice Address",
+            },
+        )
+        wiz = self.wiz_model
```
simahawk, line 302: wiz must be initialized

simahawk (changes requested):
LG overall, minor remarks

sale_order_import/wizard/sale_order_import.py
SilvioC2C: I added the ctx key flag to allow this feature to be used even when calling these methods from the model itself, as injecting a context key is way easier than rewriting all the code that uses uninitialized wizards (most tests use uninitialized wizards too, so I assumed it's a very common case).

sale_order_import/wizard/sale_order_import.py
SilvioC2C: `business.document.import._get_country_filter()` and `business.document.import._get_country_state_filter()` both return a domain for `res.partner.search()`, they don't retrieve the `res.country[.state]` records.

sale_order_import/wizard/sale_order_import.py
simahawk: you are right, I forgot that this wizard has some weird stuff. Fine like that then :+1:

sale_order_import/wizard/sale_order_import.py
simahawk: but you could reuse the domain, right?

simahawk (approved):

sale_order_import/wizard/sale_order_import.py
SilvioC2C: Yes, but [the code that prepares each domain is almost exactly the same as the one already in use here](https://github.com/OCA/edi/blob/18.0/base_business_document_import/models/business_document_import.py#L117-L154), so there's no real benefit in using those methods (except slightly lesser code duplication); also, once we retrieve the domains with those methods, we'd have to parse them to retrieve the IDs to insert in the `create()` values dictionaries, which is an unnecessary extra-step in comparison to the current workflow.

IMHO using those methods would just add more complexity to something that seems quite straightforward: we receive the country/state code => we search for them directly.

SilvioC2C:
@simahawk I added a small fix (method `_create_missing_partner_vals_cleanup()`), did some lesser refactoring (mostly type annotations) and updated the tests (checking a few more fields in the newly created partners), so I directly squashed the changes into the existing commit and force-pushed instead of using a fixup commit.

henrybackman (approved):

vvrossem (approved):

simahawk (approved):

PR #1344, opened 2026-05-19
18.0 fix account_invoice_import_simple_pdf: optional invoice date

jans23:
In 15.0 this module didn't require an invoice date to be set and to be extracted. This fix makes the invoice date optional on 18.0. This is useful for invoices from which their date can't be extracted (e.g. because their date format doesn't match any of the supported patterns).

@hbrunn


hbrunn (approved):
looks fine, please add a test though proving that invoice creation works with and without date

hbrunn:
thanks!

jans23:
@alexis-via please have a look.

PR #1347, opened 2026-05-22
[18.0][FIX] base_business_document_import: avoid random match

SilvioC2C:


SilvioC2C:
@simahawk @jbaudoux

base_business_document_import/models/business_document_import.py
jbaudoux, line 513: You should then likely do the same here then

SilvioC2C: A few lines above we have:
```
domain_delivery = expression.AND([domain, [("type", "=", "delivery")]])
```
which ensures that `domain_delivery` is not an empty domain.

I would've added a more strict requirement by adding this at the top of the method:
```
if partner:
    domain = expression.AND([domain, [("commercial_partner_id", "=", partner.commercial_partner_id.id)]])
```
which will force all search to always filter by contacts in the same commercial entity of the customer/supplier, but then:
- if `partner` is not set, then we have no customer/supplier
- it would contradict the method's docs ("_The shipping partner can be any partner, not especially related to the customer/supplier (partner argument)_"), though imho this part makes no sense (why would we return a random partner?)

What do you think?

base_business_document_import/models/business_document_import.py
```diff
@@ -523,7 +523,11 @@ def _match_shipping_partner(
         )
         if partner:
             return partner
-        if not partner_dict.get("vat"):
+
+        # Check if we have a proper domain here: if we get ``domain = []`` at this point
+        # then we'd be searching for the whole DB for a partner and retrieve a random
+        # one, which makes no sense
+        if not partner_dict.get("vat") and domain and domain != expression.TRUE_DOMAIN:
             partner = self.env["res.partner"].search(domain, limit=1)
```
jbaudoux: This should be part of `_match_partner` as last search. Here it is lacking the extension of the domain with `_match_company_domain`.
That would also solve the same issue above with domain_delivery

jbaudoux (commented):
We should only rely on `_match_partner` and not make an additional search here

SilvioC2C:
@jbaudoux I've moved the fallback to `_match_partner()`

EDIT: and now tests are failing :sob:

SilvioC2C:
@jbaudoux I've reverted the change to fix the fallback in `_match_shipping_partner()`. The update in `_match_partner()` leads to too many test failures and I currently don't have time to check and fix them all.

simahawk:
Reminder: the scope of this work is to fix only the bad fallback on any random partner search.

simahawk (approved):

vvrossem (approved):

PR #1350, opened 2026-05-25
[18.0][IMP] base_import_pdf_by_template: Remove header_items + column to simplify templates

victoralmau:
Remove header_items + column to simplify templates

Please @pedrobaeza and @cristina-hidalgo-tecnativa can you review it?

@Tecnativa


eduezerouali-tecnativa (approved):

base_import_pdf_by_template/migrations/18.0.1.1.0/pre-migration.py
```diff
@@ -0,0 +1,23 @@
...
+         ALTER TABLE base_import_pdf_template_line
+         ADD COLUMN IF NOT EXISTS sequence INTEGER
+         """,
+    )
+    openupgrade.logged_query(
+        env.cr,
+        """
+         UPDATE base_import_pdf_template_line
+         SET sequence = column::integer
+         WHERE column IS NOT NULL
```
pedrobaeza, line 21: column?

victoralmau: Yes, the field is called `column` https://github.com/OCA/edi/blob/84a850abba72b052d1ef7a2261f7dfa71cd811b2/base_import_pdf_by_template/models/base_import_pdf_template.py#L214

pedrobaeza: And they are convertible to integer always? You can use window function for numerating them if not.

victoralmau: The “column” field refers to the corresponding column number; therefore, yes, the value in that field must be a number.

pedrobaeza (approved):
/ocabot merge nobump

base_import_pdf_by_template/migrations/18.0.1.1.0/pre-migration.py
FernandoRomera: @victoralmau 
@pedrobaeza 
column is a reserved PostgreSQL keyword — it must be quoted as "column"

pedrobaeza: Thanks for the pointer. Being a migration script, fixed directly in 6aaa7daa7994e13403b25db7f3ebd3e27aaa2a92

PR #1352, opened 2026-05-26
Forward port 15 edi exchange template oca json

leemannd:
Taking back -> https://github.com/OCA/edi/pull/1217

@simahawk @santostelmo a review is welcome


santostelmo:
@leemannd pre-commit to be checked

PR #1358, opened 2026-06-12
[19.0][MIG] sale_order_import: Migration to 19.0

Ricardoalso:


santostelmo (approved):

sale_order_import/tests/common.py
```diff
@@ -12,6 +12,9 @@ class TestCommon(TransactionCase):
     def setUpClass(cls):
         super().setUpClass()
         cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
+        cls.env["res.config.settings"].create(
+            {"group_product_pricelist": True}
+        ).execute()
```
simahawk: To not hinder perf, never use execute in cases where you don't need to install modules. Always use `set_values`.

sale_order_import/tests/test_sale_order.py
```diff
@@ -11,14 +12,17 @@ class TestOrderImport(TransactionCase):
     def setUpClass(cls):
         super().setUpClass()
         cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
+        cls.partner = cls.env["res.partner"].create({"name": "Display Name Customer"})
+        with Form(cls.env["sale.order"]) as sale_form:
+            sale_form.partner_id = cls.partner
+        cls.sale_order = sale_form.save()
```
simahawk, line 18: if you use save you don't need the ctx manager (that saves already).

simahawk (commented):
minor remarks

yankinmax (approved):

ivantodorovich (approved):
LG

Please squash the `fixup!` commit so it isn't merged like this

BhaveshHeliconia (approved):
LGTM!

PR #1362, opened 2026-06-19
[19.0][ADD] base_business_document_import_iban: to remove direct base_iban dependency out of base_business_document_import

yankinmax:
**This PR is opened in the scope of the `account_invoice_import` migration to v19.0**

Address the `account_invoice_import` [roadmap](https://github.com/OCA/edi/blob/18.0/account_invoice_import/readme/ROADMAP.md) task to remove the direct `base_iban` dependency by moving IBAN specific validation out of the generic business document import helper.

The actual `validate_iban` import lives in `base_business_document_import`, which is the `account_invoice_import` direct dependency. Therefore we add `base_business_document_import_iban` as an auto-install glue addon depending on `base_business_document_import` and `base_iban`. It owns the IBAN validation override, while `base_business_document_import` keeps the generic partner bank matching/creation logic.


ivantodorovich (approved):

Ricardoalso (approved):

ivantodorovich:
@simahawk ok to merge this?

ivantodorovich:
Hello @OCA/edi-maintainers , any chance we can get this merged? 🙏🏻

base_business_document_import/upgrades/19.0.1.0.1/post-update.py
```diff
@@ -0,0 +1,16 @@
...
+@openupgrade.migrate()
+def migrate(env, version):
+    module = env["ir.module.module"].search(
+        [
+            ("name", "=", "base_business_document_import_iban"),
+            ("state", "=", "uninstalled"),
+        ]
+    )
+    if module:
+        module.button_install()
```
jbaudoux, line 16: Changes to `base_business_document_import` should not be in the `base_business_document_import_iban` commit.

jbaudoux (changes requested):

base_business_document_import/upgrades/19.0.1.0.1/post-update.py
ivantodorovich: It kind of makes sense, thuogh.
its purpose is to preserve the current behavior on instances with the module already installed 

Also the module it installs doesn't exist before this commit

jbaudoux: ideally, should inverse the commits, so that the new module exists. Then this can be put in the commit for that module.
One module per commit.

ivantodorovich: done

jbaudoux (commented):
In the first commit, the **repo** README is modified. Can you drop that change?

ivantodorovich:
done

jbaudoux:
@sbidoul name to reserve on pypi base_business_document_import_iban

jbaudoux (approved):

sbidoul:
name reserved

PR #1363, opened 2026-06-19
[19.0][MIG] account_invoice_import: Migration to 19.0

yankinmax:
As a part of migration to v19.0 [the ROADMAP note](https://github.com/OCA/edi/blob/18.0/account_invoice_import/readme/ROADMAP.md) is addressed and `base_iban` direct dependency is removed (it was coming from the `base_business_document_import`):
- [ ] https://github.com/OCA/edi/pull/1362


account_invoice_import/models/res_partner.py
```diff
@@ -20,8 +20,7 @@ class ResPartner(models.Model):
         "account.account",
         company_dependent=True,
         string="Default Expense Account",
-        domain="[('deprecated', '=', False), "
-        "('company_ids', 'in', current_company_id)]",
+        domain="[('active', '=', True), ('company_ids', 'in', current_company_id)]",
```
ivantodorovich: IMO you can remove the domain and replace with with `check_company=True`.

- `('active', '=', True)` is redundant with the ORM, no need to add it IMO
- `('company_ids', 'in', current_company_id)` will be added by check_company

See for example:

- https://github.com/OCA/odoo/blob/806fffb3f461abfd36b5f3fc1a983fd506f1fc81/addons/account/models/partner.py


Perhaps an interesting domain here would be one that filters on `account_type` -- because I think we expect here an expense account

 e.g: `('internal_group', '=', 'expense')`

account_invoice_import/wizard/res_config_settings.py
```diff
@@ -11,13 +11,13 @@ class AccountConfigSettings(models.TransientModel):
     adjustment_credit_account_id = fields.Many2one(
         related="company_id.adjustment_credit_account_id",
         readonly=False,
-        domain="[('deprecated', '=', False), ('company_ids', 'in', company_id), "
+        domain="[('active', '=', True), ('company_ids', 'in', company_id), "
```
ivantodorovich: Same comment as before. Use check_company and drop the explicit active check.

Keep in domain only the leaf for the account type

See for example:
- https://github.com/OCA/odoo/blob/806fffb3f461abfd36b5f3fc1a983fd506f1fc81/addons/account/models/res_config_settings.py

account_invoice_import/wizard/res_config_settings.py
```diff
@@ -11,13 +11,13 @@ class AccountConfigSettings(models.TransientModel):
...
         "('account_type', '=like', 'income%')]",
     )
     adjustment_debit_account_id = fields.Many2one(
         related="company_id.adjustment_debit_account_id",
         readonly=False,
-        domain="[('deprecated', '=', False), ('company_ids', 'in', company_id), "
+        domain="[('active', '=', True), ('company_ids', 'in', company_id), "
```
ivantodorovich: Same

account_invoice_import/wizard/res_config_settings.py
```diff
@@ -11,13 +11,13 @@ class AccountConfigSettings(models.TransientModel):
...
         "('account_type', '=like', 'expense%')]",
```
ivantodorovich: Better than an `=like`, I recommend `('internal_group', '=', 'expense')`

account_invoice_import/wizard/res_config_settings.py
```diff
@@ -11,13 +11,13 @@ class AccountConfigSettings(models.TransientModel):
     adjustment_credit_account_id = fields.Many2one(
         related="company_id.adjustment_credit_account_id",
         readonly=False,
-        domain="[('deprecated', '=', False), ('company_ids', 'in', company_id), "
+        domain="[('active', '=', True), ('company_ids', 'in', company_id), "
         "('account_type', '=like', 'income%')]",
```
ivantodorovich: Better than an `=like`, I recommend `('internal_group', '=', 'income')`

ivantodorovich (commented):
Thanks!

yankinmax:
Thank you @ivantodorovich for the review, good points for improvement!

account_invoice_import/tests/test_invoice_import.py
```diff
@@ -0,0 +1,422 @@
...
+from odoo import Command, fields
+from odoo.tests.common import TransactionCase
+from odoo.tools import file_open, float_is_zero
+
+
+class TestInvoiceImport(TransactionCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.company = cls.env.ref("base.main_company")
```
Ricardoalso, line 20:
```suggestion
        super().setUpClass()
        cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
        cls.company = cls.env.ref("base.main_company")
```

Ricardoalso (commented):
Technically LGTM

CI seems not happy tough

yankinmax:
> Technically LGTM
> 
> CI seems not happy tough

This is because we have a dependency PR here. After test requirements will be merged it will be ok. Tests are green

Ricardoalso (approved):

ivantodorovich (approved):

njeudy (approved):
Ok for me thanks for your work

ivantodorovich:
rebased and dropped the pending test-requirements
should be ready to merge now

jbaudoux (approved):
You poor guys not using UBL...

PR #1335, opened 2026-04-27
[19.0][MIG] purchase_order_import: Migration to 19.0

yankinmax:
Depends on:
- [x] https://github.com/OCA/reporting-engine/pull/1157

PR #1367, opened 2026-07-06
[18.0][IMP] purchase_order_import: replace assert statements in favor of UserError

yankinmax:
We should not use `assert` statements in business code. Python ignores assertions when running in optimized mode.


yankinmax:
Hello @simahawk 
Addressed your request here: https://github.com/OCA/edi/pull/1335#discussion_r3518573238

simahawk:
Commit msg amended.
/ocabot merge minor

PR #1156, opened 2025-05-13
[15.0][IMP] base_import_pdf_by_template: Add dd/MM/Y format date + prevent time_format if not set

victoralmau:
Add dd/MM/Y format date + prevent time_format if not set

Please @sergio-teruel can you review it?

@Tecnativa TT56270

PR #1374, opened 2026-07-28
[17.0][IMP] base_import_pdf_by_template: Prevent time_format if not set

victoralmau:
FWP from 15.0: https://github.com/OCA/edi/pull/1156

Prevent time_format if not set

```
time_format = self.time_format.replace("*", "%")
AttributeError: 'bool' object has no attribute 'replace'
```

Please @pedrobaeza can ou review it?

@Tecnativa TT56270


pedrobaeza (approved):
Fast-tracking as both bugfixes are clear:

/ocabot merge patch

PR #1376, opened 2026-07-31
[15.0][FIX] edi_voxel_stock_picking_oca: cron calls a non existing method

CarlosRoca13:
The scheduled action 'Edi Voxel: Update stock picking export status' calls 'update_voxel_export_status_cron()', but that method doesn't exist in 'voxel.mixin' (it is named '_cron_update_voxel_export_status'), so the cron fails on every run with:

        AttributeError: 'stock.picking' object has no attribute 'update_voxel_export_status_cron'

As a consequence the export status of the sent pickings is never updated. The equivalent cron in edi_voxel_account_invoice_oca already uses the right method name.

cc @Tecnativa TT63932

ping @sergio-teruel @carlosdauden


sergio-teruel (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #1220, opened 2025-10-02
[MIG] account_invoice_import_invoice2data: Migration to 16.0

MarwanBHL:
Migration based on the PR #753

depends  account_invoice_import

Issue #618, opened 2022-06-24
factur-x: new standart 


i-vyshnevska:
the new standard for schema was published and factur-x data is not valid anymore

` The factur-x.xml file is invalid against the official XML Schema definition. Here is the error which may give you an idea on the cause of the problem:
The Factur-x XML file is not valid against the official XML Schema Definition. Here is the error, which may give you an idea on the cause of the problem: Element '{urn:un:unece:uncefact:data:standard:ReusableAggregateBusinessInformationEntity:100}SpecifiedTradePaymentTerms': This element is not expected. Expected is one of ( {urn:un:unece:uncefact:data:standard:ReusableAggregateBusinessInformationEntity:100}SpecifiedTradeSettlementPaymentMeans, {urn:un:unece:uncefact:data:standard:ReusableAggregateBusinessInformationEntity:100}ApplicableTradeTax )., line 98. `

also, work on odoo for factur-x
https://github.com/odoo/odoo/pull/88923

Issue #190, opened 2020-04-17
[12.0] sale_order_ubl TypeError: a bytes-like object is required, not 'tuple'


jchanme:
When I tried to print a quotation, either the button or action menu, it will prompt an error:
`Odoo Server Error
Traceback (most recent call last):
  File "/opt/odoo/odoo-12/addons/web/controllers/main.py", line 1671, in report_download
    response = self.report_routes(reportname, docids=docids, converter=converter)
  File "/opt/odoo/odoo-12/odoo/http.py", line 519, in response_wrap
    response = f(*args, **kw)
  File "/opt/odoo/addons-test/report_xlsx/controllers/main.py", line 59, in report_routes
    reportname, docids, converter, **data
  File "/opt/odoo/odoo-12/odoo/http.py", line 519, in response_wrap
    response = f(*args, **kw)
  File "/opt/odoo/odoo-12/addons/web/controllers/main.py", line 1612, in report_routes
    pdf = report.with_context(context).render_qweb_pdf(docids, data=data)[0]
  File "/opt/odoo/addons-test/sale_order_ubl/models/report.py", line 22, in render_qweb_pdf
    pdf_content = order.embed_ubl_xml_in_pdf(pdf_content=pdf_content)
  File "/opt/odoo/addons-test/sale_order_ubl/models/sale.py", line 193, in embed_ubl_xml_in_pdf
    pdf_content=pdf_content, pdf_file=pdf_file)
  File "/opt/odoo/addons-test/base_ubl/models/ubl.py", line 529, in embed_xml_in_pdf
    original_pdf_file = BytesIO(pdf_content)
TypeError: a bytes-like object is required, not 'tuple'`

I went through the server debug log, there is nothing like this, the last lines of logs is like this:
`2020-04-17 04:38:29,773 21872 DEBUG test odoo.addons.base_ubl.models.ubl: Starting to embed UBL-Quotation-2.1.xml in PDF file
2020-04-17 04:38:29,779 21872 INFO test werkzeug: 10.10.10.50 - - [17/Apr/2020 04:38:29] "POST /report/download HTTP/1.0" 200 - 453 0.815 8.514`

I can't figure out why this is happening, can anyone help?

cvinh:
I have the same error, I think it's related to wkhtmltopdf ? What is your version ?

cvinh:
[...]

PR #1380, opened 2026-08-01
[14.0][IMP] account_invoice_import_invoice2data: adopt invoice2data 1.0 APIs

bosd:
## Summary

[invoice2data 1.0](https://pypi.org/project/invoice2data/1.0.1/) (June 2026, currently 1.0.1) dropped the \`invoice2data.main\` module this wizard imports from, breaking the module for anyone who upgrades the lib. This PR restores compatibility and adopts the new typed-error surface along the way.

Reproducer for the incompatibility:
\`\`\`console
$ pip install "invoice2data>=1.0"
$ python -c "from invoice2data.main import extract_data"
ModuleNotFoundError: No module named 'invoice2data.main'
\`\`\`

## Changes

- **Import fix**: \`extract_data\` from \`invoice2data\` directly. The \`.main\` path was removed in 1.0.
- **Typed errors** (\`raise_on_error=True\`, invoice2data #190):
  - \`RequiredFieldsMissingError\` -> user message names the specific fields + the matched template ("template acme.yml matched but could not parse: date, amount") -- direct handle for a template author to fix.
  - \`TemplateSyntaxError\` -> distinguishes an admin-fixable template bug from a per-invoice parsing failure.
  - \`NoTemplateFoundError\` -> falls through to the tesseract branch cleanly (was previously masked by \`except Exception\`).
- **Show \`template_name\`** (invoice2data 1.0, #618) alongside the extracted result in the wizard log.
- **Log-level bridge** points at \`logging.getLogger("invoice2data")\` (parent namespace of every 1.0 sub-logger) instead of the removed \`invoice2data.main.logger\`.
- **Pin \`invoice2data>=1.0\`** in the manifest's external_dependencies + requirements.txt.

## Compatibility

Requires invoice2data >= 1.0. Downstream users on the previous pre-1.0 line should stay on the previous module version (14.0.2.4.0); this PR bumps to 14.0.2.5.0.

## Test plan

- [x] Manifest lints (\`external_dependencies\` shape valid)
- [x] Local wizard file imports cleanly with invoice2data 1.0.1 installed
- [ ] OCA CI (14.0 matrix)

## Follow-up

Same change is worth porting to the 16.0 migration branch (see the pending [PR #1220](https://github.com/OCA/edi/pull/1220)) once it lands. cc @MarwanBHL @alexis-via

### DIFF
--- a/account_invoice_import_invoice2data/README.rst
+++ b/account_invoice_import_invoice2data/README.rst
@@ -233,6 +233,20 @@ Changelog
 =========
 
 
+14.0.2.5.0 (2026-08-02)
+~~~~~~~~~~~~~~~~~~~~~~~
+
+* [FIX] Restore compatibility with invoice2data >= 1.0. The pre-1.0 API
+  path ``invoice2data.main`` was removed upstream; import from
+  ``invoice2data`` directly.
+* [IMP] Surface actionable errors: on a matched-but-incomplete template,
+  the wizard now shows the template name + the specific fields that
+  could not be parsed (via ``raise_on_error=True`` and the new typed
+  exceptions ``NoTemplateFoundError`` / ``RequiredFieldsMissingError`` /
+  ``TemplateSyntaxError``).
+* [IMP] Log which template matched (``template_name``, added in
+  invoice2data 1.0) alongside the extracted result.
+
 14.0.2.2.0 (2023-03-03)
 ~~~~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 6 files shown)


bosd:
Closing -- I misread the compatibility surface. Odoo 14.0 runs on Python 3.8, but invoice2data 1.0.0 requires Python >=3.10, so `pip install invoice2data>=1.0` on the 14.0 CI fails outright (last 3.8-compatible upstream version is 0.4.5). The current 14.0 module code (`from invoice2data.main import extract_data`) is correct against the pre-1.0 line that 14.0 users are pinned to by Python-version constraints -- nothing actually broken here.

The 1.0 API adoption (typed errors, `template_name` in the output, `raise_on_error=True`) belongs on the 16.0 line, where Python 3.10+ is standard. Re-doing there. Sorry for the noise.

PR #1386, opened 2026-08-08
[18.0][IMP] account_invoice_import: add option force_no_vat_deduction to _pre_process_parsed_inv_taxes()

alexis-via:


alexis-via:
This new option is used by the module l10n_fr_einvoicing_import to allow per-directory line VAT deduction configuration, cf https://github.com/akretion/fr-einvoicing/blob/18.0/l10n_fr_einvoicing_import/wizards/account_invoice_import.py#L29

njeudy (approved):
great need this too .. tested with my own case. We need this un 16.0 and 19.0 to !

njeudy:
Reviewed and approved — and ported to the other two branches, so the option lands everywhere at once:

- 16.0 → #1387
- 19.0 → #1388

Both are straight cherry-picks of 534e1026 with your authorship preserved: `_pre_process_parsed_inv_taxes()` is identical on all three branches, so no adaptation was needed.

Worth noting for whoever merges: `l10n_fr_einvoicing_import` on the `fr-einvoicing` 18.0 branch already calls `_pre_process_parsed_inv_taxes(..., force_no_vat_deduction=True)` since aba1448, so any instance running that module without this PR raises a `TypeError` on vendor bill import. We hit it while carrying aba1448 onto our 16.0 and 19.0 migration branches — the three PRs should land together.

ivantodorovich (approved):

PR #1349, opened 2026-05-25
[18.0][IMP] base_import_pdf_by_template: Add date_time_format field to improve all formats

victoralmau:
Add `date_time_format` field to improve all formats

Please @pedrobaeza and @cristina-hidalgo-tecnativa can you review it?

@Tecnativa

PR #1395, opened 2026-08-31
[18.0][FIX] base_import_pdf_by_template_account: Fix tests based on the new version of pypdf

victoralmau:
Fix tests based on the new version of pypdf

Related to https://github.com/OCA/edi/pull/1349#issuecomment-5476755993

Please @CarlosRoca13 and @pedrobaeza can you review it?

@Tecnativa

### DIFF
--- a/base_import_pdf_by_template_account/demo/base_import_pdf_template.xml
+++ b/base_import_pdf_by_template_account/demo/base_import_pdf_template.xml
@@ -130,7 +130,7 @@
         <field name="related_model">lines</field>
         <field name="field_id" ref="account.field_account_move_line__product_id" />
         <field name="sequence">0</field>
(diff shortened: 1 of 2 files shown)


CarlosRoca13 (approved):

pedrobaeza (commented):
It's annoying that the library changes these things. Do we have the version frozen in our deployments?

victoralmau:
> It's annoying that the library changes these things. Do we have the version frozen in our deployments?

In my opinion, no—this is a specific case for a “less precise” regex (but the best one I've found for the sample PDF) in the tests.

pedrobaeza (approved):
Well, I was asking if there's going to be a problem in the deployments if the version is not the latest one (because the pip way to work), but let's merge it and see through the integration tests.

/ocabot merge patch

PR #1397, opened 2026-09-01
[18.0][ADD] sale_order_import_ubl_address_extended

SilvioC2C:
Glue module for ``sale_order_import_ubl`` and ``base_address_extended``

Module ``sale_order_import_ubl`` depends on ``base_ubl_parse``, which parses a partner's UBL values to a create/write-like dictionary.
When import a SO that also creates a new invoicing/shipping partner, it tries to add field ``street_number``, which is declared by module ``base_address_extended``.
If it is not installed, ``sale.order.import._create_missing_partner_vals_cleanup()`` removes ``street_number`` from the partner creation values.
If the module is installed, then Odoo creates a new partner with a dictionary containing ``street`` and ``street_number``, and the latter triggers an inverse method that recomputes ``street`` from fields ``street_name``, ``street_number`` and ``street_number2``; however, ``street_name`` is missing from the partner's creation values, so ``street`` becomes just ``street_number`` (street name is lost).

This glue module fixes this inconsistent behavior by parsing the ``street`` value and correctly adding ``street_name``, ``street_number`` and ``street_number2`` to the partner's creation values, so that ``street`` is correctly updated even when the inverse method is triggered.


divad1196 (approved):

simahawk (commented):
I'm not sure this is a feature that belongs only to ubl parsing.
sale_order_import has its own abstraction layer that relies on dicts containing certain info. I tend to say that we should not depend on the _ubl module.
In the future, if you want to import orders w/ full addresses from non xml files this module will make it OOTB.

SilvioC2C:
Superseded by #1404

PR #1403, opened 2026-09-04
[18.0][MIG] account_edi_ubl_cii_retrieve_tax: Migration to 18.0

jbaudoux:
cc @sbejaoui 

Added a repo `.prettierignore` to not reformat test files


jbaudoux:
This module needs a complete refactor for odoo v18.
The processing is now done in 2 steps:
1. create a dict with the values - here we lack the VATEX code
2. retrieve the tax based on the dict

However the retrieval is lacking a proper hook to process additional data from the dict.

Nevertheless, there is a workaround by setting a default tax on the account. It's not processing the received data but still allows to have correct result (if you set yourself the value).

Closing the PR as it requires changes in odoo

PR #1404, opened 2026-09-04
[18.0][ADD] sale_order_import_address_extended

SilvioC2C:
Glue module for ``sale_order_import`` and ``base_address_extended``

When importing a SO that also creates a new partner, ``sale_order_import`` may try to set field ``street_number``, which is declared by module ``base_address_extended``.
If the latter is not installed, ``sale.order.import._create_missing_partner_vals_cleanup()`` removes ``street_number`` from the partner creation values.
If the module is installed, however, then Odoo creates a new partner with a dictionary containing ``street`` and ``street_number``, and the latter triggers an inverse method that recomputes ``street`` from fields ``street_name``, ``street_number`` and ``street_number2``; however, ``street_name`` is missing from the partner's creation values, so ``street`` becomes just ``street_number`` (street name is lost).

This glue module fixes this inconsistent behavior by parsing the ``street`` value and correctly adding ``street_name``, ``street_number`` and ``street_number2`` to the partner's creation values, so that ``street`` is correctly updated even when the inverse method is triggered.


SilvioC2C:
Supersedes #1397

sale_order_import_address_extended/wizard/sale_order_import.py
```diff
@@ -0,0 +1,33 @@
...
+from odoo.tools import street_split
+
+
+class SaleOrderImport(models.TransientModel):
+    _inherit = "sale.order.import"
+
+    def _create_missing_partner_vals_cleanup(self, vals: dict) -> dict:
+        # OVERRIDE: handle ``street``, ``street_name``, ``street_number`` and
+        # ``street_number2`` fields.
+        # If the module ``base_address_extended`` is installed, Odoo will call
```
divad1196, line 14: It's very good to explain the "why", we often forget to do it

It could be a bit lighter IMO
> Because of `base_address_extended`,
> Odoo calls `res.partner._inverse_street_data()` after a partner is created
> If one of `street_[name|number|number2]` fields is missing, then `street` is cleared
> Fix: we derive `street_[name|number|number2]`  from `street` before creating the partner

- no conditional because `base_address_extended` is guaranteed to be installed
- Remove `_inverse_street_data` details. A dev can just go look at the code if they doubt your claims.

divad1196 (approved):
LGTM

simahawk (approved):

simahawk:
@SilvioC2C pls check tests

SilvioC2C:
@simahawk the second commit should fix the issue, though I'm not sure it's the best way to do it. Mind reviewing?

simahawk:
> @simahawk the second commit should fix the issue, though I'm not sure it's the best way to do it. Mind reviewing?

Indeed, not ideal. 
It's only the presence of the base module or your override in the glue module that causes the issue?
If the latter, I'd rather add a flag on the wizard to turn on/off the behavior.

SilvioC2C:
@simahawk installing the `base_address_extended` module is enough to make the tests fail.

simahawk:
In the base tests, can't we simply check if the the street starts with the street name? This is enough IMO

SilvioC2C:
@simahawk done

SilvioC2C:
@simahawk can you retry?

PR #1410, opened 2026-09-10
[FIX] account_invoice_import_ubl: read the item name (BT-153) for the line label

am-technix:
`cac:Item/cbc:Name` is the item name and is mandatory in EN 16931, whereas `cac:Item/cbc:Description` is optional and free-form. The parser only ever read the latter and fell back on `-`.

Two real-world consequences, both seen on invoices received through a French accredited platform (PDP):

* an issuer that does not fill in the description produces an invoice whose lines are **all** labelled `-`, while `cbc:Name` holds `Gateway - general`, `Credit Card`, `Chargeback`...
* an issuer that puts an internal identifier in the description produces lines labelled with a UUID, repeated identically on every line of the invoice, while `cbc:Name` holds `Forfait mensuel pour une instance b2-15 (...)`.

This PR reads BT-153 first and keeps BT-154 as the fallback.

Checked against 6 real UBL invoices from 2 issuers: every line now carries a readable label, and no other value changes.

### DIFF
--- a/account_invoice_import_ubl/readme/CONTRIBUTORS.md
+++ b/account_invoice_import_ubl/readme/CONTRIBUTORS.md
@@ -1,2 +1,4 @@
 - Alexis de Lattre \<<alexis.delattre@akretion.com>\>
 - Andrea Stirpe \<<a.stirpe@onestein.nl>\>
+- [teamDSI](https://www.team-dsi.fr):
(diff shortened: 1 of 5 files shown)


am-technix:
Added a unit test covering the three cases: BT-153 and BT-154 both present and differing, BT-154 alone, and neither of them. The sample file is synthetic, so no customer data is involved.

PR #1411, opened 2026-09-10
[FIX] account_invoice_import_ubl: unit price that does not match the line amount

am-technix:
`cac:Price/cbc:PriceAmount` was used as-is, and only recomputed from `cbc:LineExtensionAmount` when the tag was missing altogether. Two real-world cases, both on invoices received through a French accredited platform (PDP), end up with lines at zero, each doubled by an "Adjustment on ..." line created by `_post_process_invoice()`:

* an issuer that sends `<cbc:PriceAmount>0.00</cbc:PriceAmount>` on every line, while `cbc:LineExtensionAmount` holds the real amounts. A 9-line invoice was imported as 18 lines: 9 at zero and 9 adjustments;
* an issuer that quotes a price for a base quantity other than one, since BT-149 (`cbc:BaseQuantity`) was not read at all.

BT-131 is the authoritative net amount of the line, so this PR honours `cbc:BaseQuantity` and falls back on `LineExtensionAmount / quantity` whenever the resulting unit price does not reproduce it. When the file is internally consistent, the behaviour is unchanged.

Checked against 6 real UBL invoices from 2 issuers: the sum of the imported lines now matches BT-106 to the cent on all of them, with no adjustment line left.

**One related case left out on purpose, on which I would welcome an opinion.** A unit price that is arithmetically correct but smaller than the `Product Price` decimal precision (an issuer billing at `0.000015` per unit for `74912` units, for a line amount of `1.12`) is still rounded to zero when stored on the invoice line, and still produces an adjustment line. Raising the decimal precision is one answer, but it is a database-wide setting for what is really a per-invoice edge case. I did not want to smuggle a quantity rewrite into this PR, so I left it out; happy to follow up if you have a preference.

### DIFF
--- a/account_invoice_import_ubl/tests/__init__.py
+++ b/account_invoice_import_ubl/tests/__init__.py
@@ -1,1 +1,2 @@
 from . import test_ubl
+from . import test_ubl_price_unit
--- a/account_invoice_import_ubl/tests/files/UBLInvoice-price-unit.xml
+++ b/account_invoice_import_ubl/tests/files/UBLInvoice-price-unit.xml
@@ -1,0 +1,105 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<Invoice
+    xmlns="urn:oasis:names:specification:ubl:schema:xsd:Invoice-2"
+    xmlns:cac="urn:oasis:names:specification:ubl:schema:xsd:CommonAggregateComponents-2"
+    xmlns:cbc="urn:oasis:names:specification:ubl:schema:xsd:CommonBasicComponents-2"
+>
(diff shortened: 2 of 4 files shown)


am-technix:
Fixed the `pre-commit` failure (a `black` line-length nit on the `float_is_zero()` call), and added a unit test covering the three cases: `cbc:PriceAmount` at 0.00, a price quoted for a base quantity other than one, and a consistent file whose price is kept unchanged. The sample file is synthetic.

Note this will trivially conflict with #1410 on `tests/__init__.py`. Happy to rebase whichever lands second.

PR #1412, opened 2026-09-10
[FIX] account_invoice_import_facturx: line level allowances counted twice

am-technix:
BT-131 (`ram:LineTotalAmount`) is already net of the line level allowances and charges, but the parser used it as the amount of the product line and *then* appended each `ram:SpecifiedTradeAllowanceCharge` as a separate line. Every allowance was therefore counted twice, and the difference ended up on the global adjustment line built by `_post_process_invoice()`.

A real invoice received through a French accredited platform (PDP) made the problem obvious: 370 lines, 115 allowances, and a single "Adjustment" line of **-1471.95 EUR**, exactly twice the sum of the allowances, posted to the adjustment income account. The total was right, the breakdown was not: the expense account was overstated and revenue appeared out of nowhere.

The same invoice also showed that this issuer sends a **negative** `ram:ActualAmount` for its allowances, while BT-92 and BT-136 are defined as positive amounts, the direction being carried by `ram:ChargeIndicator` alone. Taking the amount as-is negated it twice and turned each allowance into a charge, which is how a "Remise" line ended up increasing the invoice.

This PR rebuilds the gross amount of the product line from BT-131 and the allowances and charges taken in absolute value, so that the lines returned always add up to BT-131 whatever the sign convention of the issuer.

Checked against 8 real Factur-X invoices from 5 issuers, including the 370-line one: the sum of the imported lines now matches BT-106 to the cent on all of them, with no adjustment line left, and the allowances read as negative amounts as they should.

### DIFF
--- a/account_invoice_import_facturx/readme/CONTRIBUTORS.md
+++ b/account_invoice_import_facturx/readme/CONTRIBUTORS.md
@@ -1,1 +1,3 @@
 - Alexis de Lattre \<<alexis.delattre@akretion.com>\>
+- [teamDSI](https://www.team-dsi.fr):
+  - Antoine Morit \<<amorit@team-dsi.fr>\>
--- a/account_invoice_import_facturx/tests/__init__.py
+++ b/account_invoice_import_facturx/tests/__init__.py
@@ -1,1 +1,2 @@
 from . import test_facturx
+from . import test_facturx_line_allowance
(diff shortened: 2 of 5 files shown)


am-technix:
Added a unit test. The sample is `Facture_FR_EN16931` from this very module, converted to XML, with a 2.00 allowance added on the first line and all the totals recomputed accordingly; it validates against the EN16931 XSD. The issuer writes the allowance as a negative `ram:ActualAmount`, as seen in the wild, so the test covers both the double counting and the sign convention. Without the fix the invoice comes out with an extra adjustment line and the allowance added instead of deducted.

PR #1418, opened 2026-09-17
[14.0][FIX] edi_oca: fix search logic and pagination

eantones:
Backport of https://github.com/OCA/edi-framework/pull/216, plus two more fixes on the same access filter of `_search`: keep the requested order when nothing is filtered out, [as 17.0 does](https://github.com/OCA/edi-framework/blob/427c99c7a2e28637bd8eb9b33d59f9fd9bc2dafb/edi_oca/models/edi_exchange_record.py#L513-L516), and apply `offset`/`limit` after the access filter instead of before it, so a page is never short and rows are not repeated or skipped across pages. This also removes the refill that re-read the table to the end on every page, which made the list very slow for a user with many hidden exchanges.

The grouped counts are fixed apart in https://github.com/OCA/edi/pull/1417.

18.0 and 19.0 already have the first two fixes (the first one comes from there, and they return the original query when nothing is filtered out), so only the pagination fix is forward-ported: 18.0 in https://github.com/OCA/edi-framework/pull/347, 19.0 in https://github.com/OCA/edi-framework/pull/348.


SilvioC2C:
Overall LGTM, but pre-commit and both tests are failing. Mind taking a look? :smile:

eantones:
> Overall LGTM, but pre-commit and both tests are failing. Mind taking a look? 😄

Thanks for the review. Yes, I saw them, the three checks were already red when I opened the PR and I left them alone because they are the 14.0 branch's own, they fail the same way on the branch's last push and on every open 14.0 PR. pre-commit fails in the two setuptools-odoo hooks because setuptools-odoo 3.1.8 imports pkg_resources, which setuptools 82 removed. Both test jobs end with 1 error of 289, `test_invoice_number_parsing` in account_invoice_import_simple_pdf, whose first sample `INV/2020/05/0042` uses a year4 pattern that only accepts the last five years, so it stopped matching on 2026-01-01. All edi_oca tests pass, including the new one.

